Issue #35, opened 2014-07-30
Improvements for quadrature routines


albop:
Since the quad branch is closed, I'm opening a new issue with the comments I made about it, so that we can start a fresh discussion.
There are currently a few issues with the quadrature routines. The two items  can have side-effects on future code and should be fixed quickly.
- the `gridmake` function does not enumerate points in a way that is consistent with default Python conventions. If you compute `gridmake(array([0,1]), ([-1,2]))` it produces:

```
array([[ 0, -1],
       [ 1, -1],
       [ 0, -2],
       [ 1, -2],
```

meaning that the first order varies faster. Now, if you want to represent values on this grid by a 2d array `vals` such that `vals[i,j]` contains values, then when you do vals.ravel(), you don't enumerate points in the same order because last index is supposed to vary faster. This one is quite annoying.
- Multidimensional functions don't always return an array with the same number of dimensions. That is not a problem in Matlab, but in Python, if I do `quad.qnwnorm([2,2])` I get 2-dimensional arrays and with qnwnorm([2]) I get 1-dimensional vectors. This is problematic, since in generic code, it will force one to always distinguish dimension 1 from higher dimensions. My opinion here is that multidimensional routines, should always return multidimensional objects in a predictible fashion. (maybe the 1-d routines could be exported too)
- (minor) It looks like qnwnorm will fail if one column of the covariance matrix (and the corresponding line) is full of zeros. That was already the case in the compecon toolbox, but it is still a common usecase.
- (minor) in the future, we may want to have more engaging names than `qnwnorm` and co, don't we ?

A possible way to deal with these issues would be to rename quad.py into ce_quad.py so that the latter can be left untouched and remain as close as possible from the original version (including Fortran order). A `quadrature.py` could then contain Python compliant versions, possibly with more explicit names.

As for the `gridmake` replacement, the function `cartesian` does the required thing. It is also faster. (cf http://stackoverflow.com/questions/1208118/using-numpy-to-build-an-array-of-all-combinations-of-two-arrays)

Actually all these issues concern only the multidimensional functions.

albop:
I've been looking at this issue again since the `cartesian` routine has been renamed. Replacing Fortran order by C order in `quad.py` seems very easy (as in replacing `ckron(*weights[::-1])` by `ckron(*weights)` and `gridmake` by `cartesian`. 
[...]

PR #46, opened 2014-08-10
Added cartesian product.

albop:
This is one step to adress #35 . 
After some experiments, I wrote a new function `cartesian` to compute the cartesian product of a series of vectors. The function accepts an `order` keyword to switch between `C` and Fortran order. It is meant to replace the function in `ce_tool`.
The behaviour of the current `gridmake` function can be replicated by `def gridmake(*a): return cartesian(*a, order='C')` (or `F` during the transition).
There is also a `mlinspace(a,b,ndims,order='C')` function which naturally extends `linspace` for multidimensional vectors spaces.
 The intensive work is done in a numba helper function that essentially merges `repeat` and `tile` for 1d vectors (inspired by Julia's `repeat` function). It is very simple and njitted. Because there is no creation of intermediary arrays, the new function is approximately 30% faster on my laptop than the stackoverflow solution.
I wrote simple performance and compliance tests, but I didn't know how to distinguish them so I kept them on the same level.


mmcky:
@albop Just confirming I can now delete this branch?

albop:
Yes, no problem.

Issue #36, opened 2014-07-30
Performance


spencerlyon2:
As part of the python 3 compatibility push I have been running through all the solutions notebooks. I noticed that the following notebooks took an extra long time to run:

* ifp_solutions
* odu_solutions

This issue is merely a placeholder for us to document where we have noticed opportunities for performance improvements.

mmcky:
@spencerlyon2 Can this issue now be closed? I have on my to do list to get vbench or some performance tracker up and running.

spencerlyon2:
Yep, if it is on your radar we don't need the issue

PR #103, opened 2014-11-28
GTH_SOLVE: Numba version

oyamad:
This is to open a discussion on the trade-off between speed and simplicity/clarity of the code, with a concrete example with `gth_solve` where I replaced the current, Numpy-based vectorized version with a Numba version.

* The routine is divided into the wrapper part  and the jit-compiled part `_gth_solve_jit`.
* Matrix operations are replaced with many for loops in `_gth_solve_jit`, while the code there is now almost the same as the Julia version.
* The Numba version is 3x-30x faster than the Numpy version: see the performance comparison [here](https://github.com/oyamad/gth_solve_numba).  
  (The option `overwrite` has been dropped as it did not contribute much to speed-up.)

You guys have already had a discussion on the trade-off before #36. What degree of speed-up justifies the use of Numba? Any thoughts on this particular case?


oyamad:
I updated `gth_solve`.

* I believe the routine should check whether the denominator is nonzero. Then by adding one line, it can allow a reducible matrix input.
* Then I need a `break` statement in a for loop (actually it can be replaced by some dirty code, which I want to avoid), but with it automatic loop-jitting is not activated. So I need to separate the for loops part to be run in a nopython mode.
* I added the NumPy version back.
* There remains the testing issue (the NumPy version is not tested if Numba is installed).

(This PR is to be merged into the `numba_improvements` branch.)

oyamad:
Something strange is going on.
The tests having failed for the previous commit, I tried just changing the order of the element that caused the error (`Q`) and some other (`np.eye(2)`) in `testset` in [`test_mc_tools.py`](https://github.com/QuantEcon/QuantEcon.py/commit/8ea1206a8edf67d6c2a4602cd4b95c326bb36604), and nothing other than that. Then now the tests passed.

jstac:
That is strange.  Are they failing locally or when you update the PR?

oyamad:
The previous commit [c51ce3f](https://github.com/oyamad/QuantEcon.py/commit/c51ce3fb1db66265dfecc4a4d3e9b7cb1212e6ef) failed both on Travis CI and on my machine (with the same errors); just changing the order made the tests pass both on Travis CI and on my machine..

jstac:
@oyamad

jstac:
@oyamad I accidentally clicked on 'close and comment'... sorry about that.  I've reopened.  I'll look at this tomorrow.

oyamad:
I guess I have identified the place that caused the strange behavior above, which is [this part](https://github.com/QuantEcon/QuantEcon.py/blob/8ea1206a8edf67d6c2a4602cd4b95c326bb36604/quantecon/mc_tools.py#L255):

* It turned out that `self.P[rec_class, :][:, rec_class]` is allocated in Fortran order.
* It seems that the behavior of Numba is unstable for an array in Fortran oder; see numba/numba#1103.
* It should work properly if I replace `self.P[rec_class, :][:, rec_class]` by `self.P[:, rec_class][rec_class, :]`,
where the latter will be in C order. (Actually I don't see the logic behind here.)

spencerlyon2:
Good find. I'm not totally sure, but my initial impression is that the second case (with trailing :) is C order because you are taking an entire row of a matrix. C order stores matrices one row at a time, so this should be a contiguous slice for a C ordered array.

oyamad:
The issue has been handle amazingly quickly in numba/numba#1104.

The fix in `gth_solve` is to add `order='C'` when making a copy `A1` of the input `A`. I added a test to make sure that the problem is avoided with this.

oyamad:
As another issue, which is better, `self.P[rec_class, :][:, rec_class]` or `self.P[:, rec_class][rec_class, :]`, or is there another better way?

```python
import numpy as np

P = np.zeros((3, 3))
P[0, [0, 2]] = 0.4, 0.6
P[1, 1] = 1
P[2, [0, 2]] = 0.2, 0.8
print P
[[ 0.4  0.   0.6]
 [ 0.   1.   0. ]
 [ 0.2  0.   0.8]]

rec_class0 = [0, 2]
P_0 = P[rec_class0, :][:, rec_class0]
P_1 = P[:, rec_class0][rec_class0, :]

print(P_0)
print(P_0).flags
[[ 0.4  0.6]
 [ 0.2  0.8]]
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False

print(P_1)
print(P_1).flags
[[ 0.4  0.6]
 [ 0.2  0.8]]
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
```

Does `OWNDATA : False` mean that the array is actually a view? If that's the case, the former seems more efficient in terms of memory consumption. Is there any difference in access speed?

jstac:
Yes, it's a view, as far as I know, while `P_1` is a copy.  The latter is also seen to be a copy because it has no base attribute.

``` python
print P_0.base
[[ 0.4  0.2]
 [ 0.6  0.8]]
print P_1.base
None
```

So `P_0` seems more efficient.  I don't really know which is better though.

jstac:
@oyamad @mmcky Hi guys, I've kind of lost track of what's happening with this PR and the different issues.  Would you mind to update me?  From memory, we are looking to merge in

* a utilities file
* a numbafied routine for sampling from MCs
* some standard code for numba imports with a common warning message

Just a quick update of the situation would be helpful.  Also, if possible I would like the utilities file added soon because one of Tom's coauthors has some utilities that look useful and are needed for a lecture Tom wants to add to quant-econ.net.

oyamad:
@jstac Regarding this PR (on `gth_solve`):

* the main part is Numba jitted with `nopython=True`;
* the NumPy version is kept alive, which is executed if not numba_installed;
* the current structure of the code does not allow the unittest to cover the NumPy version.

If it's ok, I would like to merge this PR into `numba_improvements` to be reviewed with other changes in that branch.

As for the discussions regarding `numba_improvements`, we should open a new PR and make a list of the issues to discuss there.

jstac:
@oyamad That sounds like a good next step.  Please go ahead.  

(Thanks for the summary.)

oyamad:
Merged into 'numba_improvements`. Closing.

PR #108, opened 2015-01-23
FIX: Discrete Lyapunov with complex input

ChadFulton:
Minor issue with the discrete Lyapunov solver via doubling: a missing complex conjugation operation causes incorrect results if the input is complex. The unit test shows the failure in the current setup and passes with the fix.


cc7768:
Excellent.  Thanks for noticing this, we will have a look at the PR asap.

jstac:
@ChadFulton Many thanks, much appreciated.    Merging.

PR #110, opened 2015-01-26
STY: made all .py files pep8 compliant

spencerlyon2:
In the examples directory I left any case of "aligned equal signs". For example, I left this

```python
T    = 20
A0   = 100.0
A1   = 0.05
d    = 0.20
beta = 0.95
```

Instead of changing to

```python
T = 20
A0 = 100.0
A1 = 0.05
d = 0.20
beta = 0.95
```

as pep8 would suggest. I felt that was a fairly innocent violation.

Thanks to @cc7768  for helping on this. We split up the work, but squashed into a single commit before creating the PR

### DIFF
--- a/examples/3dplot.py
+++ b/examples/3dplot.py
@@ -3,14 +3,15 @@ from mpl_toolkits.mplot3d.axes3d import Axes3D
 import numpy as np
 from matplotlib import cm
 
+
 def f(x, y):
     return np.cos(x**2 + y**2) / (1 + x**2 + y**2)
 
 xgrid = np.linspace(-3, 3, 50)
 ygrid = xgrid
-x, y = np.meshgrid(xgrid, ygrid)  
+x, y = np.meshgrid(xgrid, ygrid)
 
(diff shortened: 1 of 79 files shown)


jstac:
Wow, thanks guys, nice work.

You didn't change any variable names, did you?  For example, matrix `A` changed to `a`?  I couldn't see any such changes.  I ask because the lectures might need to be changed correspondingly.  

I am very pro consistency and like to be close to PEP8.  The only reason I consistently violate it is to bring variable names closer to mathematical symbols --- for example, upper case for some variable names. 

PS Does anyone know when the @ operator is coming to numpy?  I'm really looking forward to changing all those `np.dot(A, B)` to `A @ B`

spencerlyon2:
Nope, didn't change any names.

It was mostly spaces after commas, newlines before functions/methods/classes and at EOF, and making sure line length doesn't go beyond 80.

It will come with python 3.5. From the [release schedule](https://www.python.org/dev/peps/pep-0478/) it look like it won't be official until September

Issue #85, opened 2014-10-27
Improve PyPi quantecon page


jstac:
Some basic information needs to be added to this page:

https://pypi.python.org/pypi/quantecon

As a model, see, for example, this page:

https://pypi.python.org/pypi/gunicorn/

spencerlyon2:
ping

Do we have a better pypi page/readme yet?

jstac:
It could definitely be improved.  @mmcky Did we have a plan for this?  I don't want to replicate info from quantecon.org but it might be good to make the page look a bit nicer and have the link to quantecon.org more prominent.

mmcky:
@jstac Put together some suggestions which can be found in the PR just submitted. 
I have followed the lead of ``pandas`` and integrated the PYPI information into ``setup.py`` - which will flow through to PYPI without a manual upload.

mmcky:
This should propagate through to the PYPI page when we next update PyPI.

PR #112, opened 2015-01-29
Adding more information for the PYPI README

mmcky:
Adding a new DESCRIPTION, LONG_DESCRIPTION, LICENSE and CLASSIFIERS for the PYPI page. 
Using setup tools this should automatically build a new PYPI README.

To view the markdown we may want to submit to pypi test server first.  

This closes issue #85

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,8 @@
 from distutils.core import setup
 import os
 
-# Write a versions.py file for class attribute
+#-Write Versions File-#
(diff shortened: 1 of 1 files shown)


jstac:
Many thanks. I don't have time to look at this on the test server so I'm going to accept as is --- given that it's only documentation and that we can easily fix it ex post if there's problems with the html.

PR #113, opened 2015-01-29
BUG: Found bug in solutions notebook 

spencerlyon2:
The solutions notebook as still using old naming and calling the LinearProcess class, not ARMA...

also added str, repr, _repr_latex method to ARMA (``_repr_latex_`` gives an intelligent latex representation of the process if viewed from the notebook ... pretty cool actually)


jstac:
Thanks!

PR #117, opened 2015-02-08
Add symbolics base class

davidrpugh:
@albop @spencerlyon2 @jstac @cc7768 

I have been thinking about how to specify a generic base class for continuous-time models.  Here is my first draft of said base class (only spent about and hour or so on this so I am sure that there are many improvements that could be made).

The basic idea is that the user should specify the following primitives:

1. `rhs`: This would be a `sympy.Matrix` representing the right-hand side of a system of linear/non-linear differential/difference equations.
2. `independent_var`: This would be a `sympy.Symbol` representing the independent variable in the model (i.e., time in a dynamic setting).
3. `dependent_vars`: List of `sympy.Symbols` representing the dependent variables for the model (obviously number of equations should match the number of dependent variables).
4. `params`: dictionary of parameter values.

From these primitives, the base class would provide the functionality to build the Jacobian for the right-hand side and convert both the symbolic Jacobian and the symbolic right-hand side into vectorized, NumPy-aware functions.  

I could imagine two classes that might inherit directly from `SymbolicBase`:

```python
class DifferentialEquation(SymbolicBase):
    pass


class DifferenceEquation(SymbolicBase):
    pass
```

All instances of `DifferentialEquation` would then be passed to a `Solver`, which depending on the type of differential equation and boundary conditions, might be the solvers that currently exist in the `IVP` module (or perhaps some other solvers such as orthogonal collocation solvers, finite element solvers, etc.  Meanwhile, all instance of `DifferenceEquation` would be passed to a `Simulator` that would efficiently simulate the system of difference equations.

Thoughts and feedback would be much appreciated.

### DIFF
--- a/quantecon/models/solow/bases.py
+++ b/quantecon/models/solow/bases.py
@@ -62,7 +62,7 @@ class SymbolicBase(object):
 
     @property
     def _variable_subs(self):
-        """Variable substitutions."""
+        """Generic variable substitutions."""
         return dict(zip(self.dependent_vars, [X[i] for i in range(self.N)]))
 
     @property
@@ -184,7 +184,7 @@ class SymbolicBase(object):
     def _validate_rhs(self):
         raise NotImplementedError
 
-    def evaluate_jacobian(self, t, X, *args):
+    def evaluate_jacobian(self, t, X):
         """
         Return the Jacobian matrix of partial derivatives evaluated at specific
         values of the independent variable, `t`, and dependent variables, `X`.
@@ -195,8 +195,6 @@ class SymbolicBase(object):
             Independent variable.
         X : numpy.ndarray (shape=(N,))
             Array of values for the `N` dependent variables.
-        args : tuple
-            Additional parameter arguments necessary to evaluate the Jacobian.
 
         Returns
         -------
@@ -204,10 +202,10 @@ class SymbolicBase(object):
             Evaluated Jacobian matrix.
 
         """
-        jac = self._jacobian(t, X, *args)
+        jac = self._jacobian(t, X, *self.params.values())
         return jac
 
-    def evaluate_rhs(self, t, X, *args):
+    def evaluate_rhs(self, t, X):
         """
         Return the right-hand side of a system of differential/difference
         equations evaluated at specific values of the independent variable,
@@ -219,8 +217,6 @@ class SymbolicBase(object):
             Independent variable.
         X : numpy.ndarray (shape=(N,))
             Array of values for the `N` dependent variables.
-        args : tuple
-            Additional parameter arguments necessary to evaluate the Jacobian.
 
         Returns
         -------
@@ -228,5 +224,5 @@ class SymbolicBase(object):
             Evaluated Jacobian matrix.
 
         """
-        rhs = self._rhs(t, X, *args)
+        rhs = self._rhs(t, X, *self.params.values())
         return rhs
--- a/quantecon/models/solow/bases.py
+++ b/quantecon/models/solow/bases.py
@@ -187,8 +187,13 @@ class SymbolicBase(object):
         else:
             return value
 
-    def _validate_rhs(self):
-        raise NotImplementedError
+    def _validate_rhs(self, rhs):
+        """Validate the rhs attribute."""
+        if not (rhs.shape[0] == len(self.dependent_vars)):
+            mesg = "Number of equations must equal number of dependent variables."
+            raise ValueError(mesg)
+        else:
+            return rhs
 
     def compute_eigenvalues(self, t, X):
         """
--- a/quantecon/models/solow/bases.py
+++ b/quantecon/models/solow/bases.py
@@ -9,6 +9,7 @@ from __future__ import division
 import collections
 
 import numpy as np
+from scipy import linalg
 import sympy as sym
 
 # declare generic symbolic variables
@@ -188,6 +189,28 @@ class SymbolicBase(object):
 
     def _validate_rhs(self):
         raise NotImplementedError
+
+    def compute_eigenvalues(self, t, X):
+        """
+        Compute the eigenvalues of the Jacobian matrix of partial derivatives
+        evaluated at specific values of the independent variable, `t`, and
+        dependent variables, `X`.
+
+        Parameters
+        ----------
+        t : float
+            Independent variable.
+        X : numpy.ndarray (shape=(N,))
+            Array of values for the `N` dependent variables.
+
+        Returns
+        -------
+        eigvals : numpy.ndarray (shape=(N,))
+            Array of eigenvalues of the Jacobian matrix.
+
+        """
+        eigvals = linalg.eigvals(self.evaluate_jacobian(t, X))
+        return eigvals
 
     def evaluate_jacobian(self, t, X):
         """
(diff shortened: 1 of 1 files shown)


jstac:
This is nicely put together, obviously.  

Sorry to be negative but to be honest I don't know whether I would use this kind of code rather than just access the relevant solvers directly.   I guess I get more excited about code that implements algorithms.

Just my immediate reaction.  Always happy to be shown the error of my ways.

davidrpugh:
@jstac 

If you solve a lot of systems of differential/difference equations and just access the solvers directly you will end of typing the same kind of code (i.e., code to represent the model, represent its Jacobian, access the solvers, compute eigen-values, etc) over and over again.  Given a suitable representation of the problem to be solved most all of this repetitive code can be automated. This is the point of the base classes.  All of the above has already been implemented in the `solow` module, but I want to extract the functionality into a lower level base class that any continuous time model can use.

To reiterate, the big value add to this approach is getting SymPy to compute the Jacobian matrix. Having access to the exact Jacobian substantially improves accuracy and efficiency of ODE solvers and for non-trivial problems computing the Jacobian by hand is tedious and error-prone.

Also if it wasn't clear, users of the code would only need to provide a SymPy representation of their model and then directly apply a solver of their choice to that problem (which is pretty much what I think you want to do). 

The eventual user syntax that I am aiming for would be something like...
```python
rhs = # list of symbolic expressions defining the rhs of some system of ODEs
param = # dictionary of model parameters
model = DifferentialEquation(rhs, params)
solver = FiniteDifferenceSolver(model)
solution = solver.solve(*args, **kwargs)
```

Currently the `ivp` module only implements finite difference solvers, but I have written code for orthogonal collocation and finite element solvers for ODEs that I plan on contributing in near future.

jstac:
Fair enough.  Anyone else have thoughts on this?

albop:
Yes, I have some. I was responding but lost my browser in the process. Will rewrite it.

albop:
@davidrpugh and @jstac : in these discussions, I recognize a recurrent pattern. There are two approaches that are both worth investigating (or a matter of taste)

1/ provide modeling bricks that can be used to write a solution code by hand, in a transparent and concise way, using Python's expressiveness. The code is distributed as simple examples, that can be easily adapted to other problems.

2/ design once for all a generic model object that can be solved without additional coding, trading some transparency for increased user-friendliness, and efficiency. this require careful thought and frequent interactions with end-users, until a satisfying amount needs are covered.  

The work I have been doing on dolo is essentially focused on 2/. @davidrpugh, when I see your work on continuous time models, I have the feeling that you are leaning in that direction too. I would encourage you to have a look at [dolo](https://github.com/EconForge/dolo) and see whether what you are doing is not already there in which case it may be great to build something on top of it. The work-in progress documentation is hosted on [readthedocs](http://dolo.readthedocs.org/en/doc/)

Let me elaborate a bit:

- one question is "how to define a multivariate function and evaluate it efficiently with its derivatives ?" A function doing that with sympy could do that in a few lines (<10?) and would be at the same time, transparent and useful in many contexts. To me, it would be a good candidate for inclusion in quantecon. Now, doing that efficiently is a different matter and a few different options should be evaluated before building a big machinery on it.

- The sympy approach has a few shortcomings though, notably numerical stability issues due to the fact that it doesn't guarantee execution order. There are also other options out there (theano, funcdesigner, direct ast) that don't share the same defects but of course no one is perfect. Lately, I've been using the AST more and more as it is very previsible. dolo currently has an API to construct functions from expressions strings. It understands time lags and compile expressions to numexpr functions using either sympy or ast. I am relatively happy with the API, but the implementation could be improved quite a bit. For instance, one latent goal would be to be able to define a pure python function, differentiate and compile it directly using just a decorator. The way numba.guvectorize would do, but with derivatives. If there is strong demand for that kind of functionality (ala patsy, but for economic equations) it could be made into an independent package.

- As for the model object, I don't know exactly which class of model is targeted by the continuous time module (perfect foresight, stochastic optimization problems, differential games, option pricing, etc..). Given the diversity of model types, there is probably not a one size-fits-all solution, but if the focus is on macro models with representative agents, then extending dolo to continuous time would be a natural thing to do. Concretely, we would define the type of equations that a continuous time macro model is allowed to have, possibly add support for dot notation, and that would be it for the model definition part (notwithstanding any improvement in the process). Then the solution routines would need to be written as well as other components that would clearly fit better in quantecon (like continuous time distributions).  @davidrpugh, please tell me if you have any interest in that, in which case I will make the required adjustments. Extending dolo to continuous time has been in the pipeline for a while but without a precise timeline assigned to it.

davidrpugh:
@albop 

Thanks for the comments. I just realized that I unintentionally focused this thread on continuous-time models.  I am actually interested in building a general purpose tool for solving systems of non-linear differential equations with boundary conditions of the type commonly encountered by economists. Obviously continuous-time models of economic dynamics fall into this category, but there are many other commonly encountered classes of models whose equilibria reduce to a system of differential equations. For example (and off the top of my head)...

* Signaling models
* Auction models
* Evolutionary microeconomic dynamics
* Search and match models
* Models of assortative matching

I have written orthogonal collocation and finite elements solvers for specific research projects in the above categories.  In my mind it seems ideal to refactor the solvers so that they work on reasonably generic boundary value problems, contribute the solvers to QuantEcon, and then use them to solve specific research problems.

I don't think I would be interested in extending dolo to handle continuous-time representative agent models partly due to time constraints (I am already finding it difficult to balance ongoing projects) and partly due to personal research interests (I am moving away from DSGE-style models in favor of agent-based models).

albop:
@davidrpugh: indeed I could have looked more carefully and see `DifferenceEquation` in your explanations. 

This doesn't change my main point though: there should be a function responsible for compiling these difference equations to a numeric function, that would also work for more general functions. Otherwise, you will want to refactor it again sooner or later, for instance when you go from difference equation to more general edps. This is something I did in dolo, and I really think that we should use something like that in quantecon. I'll open an issue to discuss the matter, and when we find a solution that everybody like we can revisit the next issue of building models easily. At least that is the way I would proceed forward. @davidrpugh and @jstac : What do you think ?

In the meantime, I am not in favour of merging this PR, as I don't believe that organic growth produces good apis. To me this code is either too specific or too generic.

davidrpugh:
@jstac and @albop 

This PR wasn't meant to be merged anytime soon! It was meant as a placeholder for me to get some feedback on my api design.  In the meantime I am starting a fresh repo for generic boundary value problem solvers (i.e., the collocation and finite element solvers) that I have already written. I will likely ask for more specific feedback on those solvers as soon as they are ready to merge.

I agree with the need for a more complete solution to the problem of representing functions symbolically, computing derivatives, and then wrapping the functions for numerical evaluation. I use this functionality a lot and would be interested in a more efficient solution then my current SymPy approach.

albop:
Perfect !
 My general feeling is that new features when not completely trivial, should live outside for a while (i.e. in another repo), be tested and stabilized before discussing inclusion. It should be possible to attract scrutiny from other developers without doing a pull request, which are not really meant for that, and feel awkward, when one decides to reject one after a long time.
Maybe we could use some specific tag for issues that are meant to attract attention to what happens in outside repositories (clones or independent ones).

davidrpugh:
@albop 

That seems like a good idea to me.  I am happy to close this PR.

jstac:
Thanks guys, interesting discussion.  I'm closing the PR.

PR #120, opened 2015-02-10
Fixed import statements in __init__.py to include solow module.

davidrpugh:
@jstac @spencerlyon2 

I noticed when I upgraded to the new release of `quantecon` that
```
from quantecon.models import solow
```
command that had been working on my machine was no longer working.  Not sure exactly what the problem is. After checking the `__init__.py` file in the `models` module I noticed that `solow` was not in the module's name space.  This PR should fix that problem. Again, everything is working on my machine but it would be good to get independent verification.


jstac:
Works on mine.  Merging...

PR #124, opened 2015-02-13
ENH: added formatting to iteration printing

spencerlyon2:
This adds a bit prettier printing scheme to compute fixed point.

It might belong somewhere else in the project as it can be used in any iterative algorithm.

Perhaps we should lump this and @albop's timer code into a `util.py` file or something similar?

### DIFF
--- a/quantecon/compute_fp.py
+++ b/quantecon/compute_fp.py
@@ -6,12 +6,35 @@ Compute the fixed point of a given operator T, starting from
 specified initial condition v.
 
 """
-
+import time
 import numpy as np
 
 
-def compute_fixed_point(T, v, error_tol=1e-3, max_iter=50, verbose=1, *args,
-                        **kwargs):
+def _print_after_skip(skip, iter=None, dist=None, etime=None):
+    if iter is None:
+        # print initial header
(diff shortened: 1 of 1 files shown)


albop:
@spencerlyon2: which timer code are you refering to ?  Is it related to the discussion we had a while ago https://github.com/EconForge/dolo/issues/34 about adding a time decorator for time consuming functions ? If so, I don't like that idea anymore: it makes more sense to use a context manager for that. I guess it would fit perfectly in quantecon. I'm opening an issue for that.

spencerlyon2:
@albop I was talking about this code: https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/timing.py

Although the `watcher` seems quite cool

jstac:
@spencerlyon2 This is nice. The function `compute_fixed_point` is already quite generic and can be used by lots of routines.  My two cents is that it's good where it is.

Shall we leave it where it is for now with the option of pulling it out and putting it in a util.py file in the future should demand arise?

albop:
Ahem, that was out of topic: I had forgotten about the tic-tac-toc routines. 
As for relocating code, I don't have any position, except a small dislike for long files. The status quo seems fine.

PR #129, opened 2015-04-05
Update lss

jstac:
Hi All,

I've

1. Added a noise term to the observation equation in the linear state space class in `lss.py`
2. Changed the name of the class from `LSS` to `LinearStateSpace` 

(I think it's better to have more descriptive names as the size of the library grows.)

The changes associated with point 1 should not break backwards compatibility but the change in point 2 of course does.  I've updated all related code, examples and solutions in QuantEcon.py accordingly, and I've made the relevant changes in the lecture file (on a branch, waiting to be merged).

Please let me know ASAP if you have issues with these changes.  Otherwise I'll merge.

### DIFF
--- a/examples/paths_and_hist.py
+++ b/examples/paths_and_hist.py
@@ -1,7 +1,7 @@
 
 import numpy as np
 import matplotlib.pyplot as plt
-from quantecon import LSS
+from quantecon import LinearStateSpace
 import random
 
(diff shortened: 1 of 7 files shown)


jstac:
Oops, forgot to commit an update for the test.  Fixing now...

PR #131, opened 2015-04-07
numba-fied simulation

jstac:
@albop 
@cc7768 
@spencerlyon2 
@mmcky 
@oyamad 

Hi guys, I want to draw on your numba experience if you have time --- if you're busy please ignore.

This summer I want to spend a bit more time on optimization.  Actually @oyamad has an open PR on this topic waiting for comments.  I think the speed gains he got in that instance were less than an order of magnitude.  For such small numbers I'd be inclined to avoid the extra complication.  However, I'm sure there's a lot of low hanging fruit in QuantEcon.py where we can get multiple orders of magnitude.  As a start I've numba-fied simulation of a VAR in lss.py.  See the first function and the autojit in line 67 of lss.py in this commit.  This should be the kind of thing numba is good at --- loop, simple function, all memory preallocated --- but it actually slows the code down on my machine.  The code I used for the test is below.  I guess the issue is how I access and write to the arrays.  Any thoughts would be appreciated.  

```python
A = [[0.4, 0.5], 
    [0.7, 0.2]]
A = np.asarray(A)
C = np.ones((2,2))

ts_length = int(1e5)
n = m = 2
x = np.empty((n, ts_length))
x0 = (0, 0)
w = np.random.randn(m, ts_length-1)
v = C.dot(w) 

%timeit simulate_linear_model(A, x0, x, v, ts_length)
```


jstac:
Edits:

1. I'm using numba 0.14.0 on Python 2.7.  I haven't tried newer versions.
2. looked again at PR #103 by @oyamad and the speed up is more than I remembered.  I'm inclined to accept this PR unless there are further comments.

albop:
@jstac: I guess numbafied function is running in object mode (that is compiled code that makes the same types checks as the interpeted one: little performance gains). The probable cause is that the call to the `dot` function make the compiler leave nopython mode. The other issue is that nopython does not currently support array expression, so that `a[:,i] + b[:,i]` is not understood (but that is supposed to come soon.
An obvious way to fix that would be to rewrite the matrix multiplication explicitly, for each line of the matrix `x`. Since the whole point of refactoring the function was to optimize it, it is probably not to bad if the implementation is numba specific.
Another thing: recent numba versions introduced support for np.random.random, so that in principle, the computation of the random innovations could be done inside the loop saving the need to preallocate the random matrix. Depending on the implementation it may also save some computational time, but I'm not sure about that. I will try to dig a bit into it if I can find time this week.

oyamad:
@jstac Please give me some time before you merge #103,
I need to replace the code in `gth_solve.py` with the simplest version `gth_solve_jit3` in [here](http://nbviewer.ipython.org/github/oyamad/gth_solve_numba/blob/master/loop_jitting02.ipynb), as well as rewrite the docstring and test accordingly since now the input matrix must be an irreducible matrix.

cc7768:
For what its worth, the list of functions that are available in `nopython` mode are listed [here](http://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#standard-ufuncs)

oyamad:
A [quick trial](http://nbviewer.ipython.org/gist/oyamad/90f283ccd39c648b0897):
The version with explicit matrix multiplication `simulate_linear_model_2` is about 400x faster.

```
1 loops, best of 3: 644 ms per loop
1000 loops, best of 3: 1.59 ms per loop
```

(As just an experiment, the transposed version `simulate_linear_model_3` did not speed up almost at all.)

jstac:
@oyamad Thanks!!  Wow, what a speed up.  I guess that proves the conjecture of @albop and @cc7768 regarding the dot function.

This is really cool.  Unless people have further comments I'll implement @oyamad 's solution in lss.py and merge into master.

mmcky:
@albop Good points. 
@cc7768 Great Link. 

Also, splitting the lines between ``np.dot()`` and the array expressions the majority of the time is spent on the ``np.dot`` method call (~33.6%)  and the ``array`` expressions (~57.7%). 

```
Timer unit: 1e-06 s

Total time: 0.485746 s
File: <ipython-input-13-6057ce74e9b1>
Function: simulate_linear_model at line 15

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    15                                           def simulate_linear_model(A, x0, x, v, ts_length):
    ...
    54         1           56     56.0      0.0      x[:, 0] = x0
    55    100000        42507      0.4      8.8      for t in xrange(ts_length-1):
    56     99999       162969      1.6     33.6          dot = np.dot(A, x[:,t])
    57     99999       280214      2.8     57.7          x[:,t+1] = dot + v[:, t]
    58         1            0      0.0      0.0      return x
```

As @albop suggests we could rewrite the array expression lines, for example, explicitly and then run through numba. 

When I get some time I will dig up a tutorial I came across a short while ago about digging into what actually get's compiled through autojit to view the output. That should help resolve some of the mystery behind what autojit successfully compiles etc.

spencerlyon2:
Great work guys.

That speedup is awesome.

Whenever I struggle with figuring out what gets jitted, I just use Julia :smirk:

mmcky:
Oops I see this thread has been updated. Please ignore :) ...

oyamad:
Some remarks:

* `autojit` has been deprecated at least in the [current version](https://github.com/numba/numba/blob/master/numba/decorators.py#L13).
* Array creation is allowed outside a for loop, so you don't have to preallocate memory for the output array `x`; see `simulate_linear_model_4` in [here](http://nbviewer.ipython.org/gist/oyamad/90f283ccd39c648b0897/lss_numba02.ipynb).
* As @albop says, functions from np.random are supported in [version 0.18.1](http://numba.pydata.org/numba-doc/0.18.1/release-notes.html#version-0-18-1) (though I haven't tried yet).

jstac:
@oyamad Thanks, that's good to know too.  The code without preallocation of memory looks much more natural (unless you're a hard core C programmer...)

I think I'll still pass the shocks in since it might take a while for people to update to the latest version of numba.  It doesn't make much difference in terms of readability.

jstac:
Rewrote simulate_linear_model according to @oyamad 's simulate_linear_model_4.

jstac:
Thanks for all your help guys.  Merging.

Issue #115, opened 2015-01-29
Question: do we need Kalman and LSS classes


spencerlyon2:
I know that in the lectures we use these two classes separately, but to me the kalman filtering routines should be methods of the LSS class that represents linear gaussian state space models.

jstac:
Hmmm.  I guess we could do that.  It's nice to unify things. If we put Kalman routines in LSS we would need to generalize the latter to accommodate noisy updating of y_t from x_t, rather than the current deterministic rule y_t = G x_t.  But that's no big deal.

Aside: LSS isn't a great name in retrospect.  Perhaps we should rename it LinearStateSpace at the same time.

jstac:
As per our email discussion, Kalman will subclass LSS, and the update rule for y_t will be generalized to include noise.  I still propose that we change the name LSS to LinearStateSpace.

jstac:
New discussion of this issue can be found in #134.  This issue should be closed when #134 is merged.

spencerlyon2:
Closing as #134 has been merged

PR #134, opened 2015-04-07
Update kalman

jstac:
@spencerlyon2 @cc7768 @mmcky 

Hi all, there was a discussion of whether or not Kalman should subclass the linear state space class (formerly LSS, now LinearStateSpace) from lss.py, or vice versa.  See #115.

In retrospect I don't think subclassing is the right way to go.  For example, a Kalman filter isn't a special kind of linear state space model.  A better way to think of it is that a linear state space system is part of the data needed to set up a Kalman filter.  The other part of the data is the prior distribution.  Hence I've rewritten Kalman to take an instance of LinearStateSpace as one of its arguments.  This replaces the individual matrices from the transition and observation equations.

Basically we are replacing inheritance with composition, which seems a better fit in this case.

Among the advantages of the new version of Kalman compared to the previous one are that we don't need the logic for checking the matrices, coercing them into arrays of the right shape, etc., in two places.  Now it's just in lss.py.  Second, once we've created an instance of LinearStateSpace we have access to its functionality if required.

All thoughts are appreciated.


davidrpugh:
Seems good refactor to me. Good case to favor composition over
inheritance...

cc7768:
This looks clean. Good changes.

jstac:
Tom has signed off, solutions updated, just need to update lecture correspondingly.

albop:
The refactored version looks good to me.
I just bumped into this piece of news: https://www.google-melange.com/gsoc/project/details/google/gsoc2014/reo/5808496591241216 . One should keep an eye on this project. It could be very useful.

Issue #137, opened 2015-04-13
Numba version of mc_sample_path


oyamad:
Hi all,
I played further with Numba, by numba-fying? [`mc_sample_path`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/mc_tools.py#L279).
The code and performance tests are found [here](http://nbviewer.ipython.org/github/oyamad/mc_sample_path_numba/blob/master/mc_sample_path_numba02.ipynb).

The Numba version can be more than 100x faster than the original `qe.mc_sample_path` depending on the problem size (and the environment).

0. Cell [3] contains the Numba version `mc_sample_path` and the NumPy version `mc_sample_path_numpy`, which follows implementation approach 1 in [@albop 's comment](https://github.com/QuantEcon/QuantEcon.py/issues/132#issuecomment-90955976).

0. `mc_sample_path_2` in cell [15] follows the approach of Code 1 in [my comment](https://github.com/QuantEcon/QuantEcon.py/issues/132#issuecomment-90792658).

As always, there is a tradeoff in introducing Numba: speed v.s. clarity/transparency. Would the use of Numba be justified in this case?

Some remarks:

* The main bottleneck in the original `qe.mc_sample_path`, compared to the NumPy version, is in random sampling, which is done for `sample_size` times. The NumPy version is faster as it calls a random sampling routine only once, with size= sample_size.

* It is interesting that the speed of the original and the NumPy versions, both of which rely on [`np.searchsorted`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.searchsorted.html), is almost constant in the size of the matrix.

jstac:
@oyamad This is great.  You write beautiful code.  Is that a bisection search in the inner loop of the `mc_sample_path` routine?

I don't have a very strong opinion over options 1 and 2 above (cell [3] vs cell [15]).  I guess I like [15] a little better because everything to do with numba is wrapped up in a single block of code where the first line makes it immediately clear this only concerns numba optimization.

I think the use of numba would most certainly be justified in this case.  Anything over an order of magnitude is worth looking at, and anything that gets two orders of magnitude should certainly be implemented.

My only concern is that people will have older versions of numba, where generation of random variables inside the function is slow...

mmcky:
@oyamad This is indeed a great example to consider regarding the style discussions we have been having in the #132 discussions. Thanks for adding this to the discussion. Here are my thoughts ...

[...]

Issue #133, opened 2015-04-07
Numba warning --- implement a common warning


jstac:
If we're going to use numba systematically throughout QuantEcon.py then each time we try to import, we should reference a common warning should that import fail.  For example see the warning in lss.py.  The warning should be put in a separate file so other modules can use it.

jstac:
Should we make this a warning that numba is going to become a dependency?

spencerlyon2:
I think the warning should mention that it is currently strongly recommended and will one day be an explicit dependency.

albop:
To me, we should use the fact that the library is in its early days to increase requirements when necessary, i.e. without delays. Most of the people will get the warning when they update quantecon, and at the moment they update, they will be gently prompted to install numba.

I am a little bit more worried about people who have an incorrect version of numba (too high or too low). In corporate environments, it is often "possible" to install things. But keeping them up-to-date is another matter and requires constant fights.

Maybe it would also be good to give some idea of how stable the library is and whether api breaks are to be expected or not. If we say that version 1.0 is to be expected X months from now, then we have some room to make disruptive changes now.

jstac:
Thanks for these thoughts.

In terms of versions of Numba, I guess the best we can do for people is stay behind the curve a little, and not immediately adopt the latest features that come out.

Dependencies: I've thought this over a bit more and my current view is that although Numba is going to play a very important role for us, I'm not sure we need to move it to an explicit dependency in the short or medium term.  My reasons are

1. Having a NumPy based version of routines is a great way to test Numba-fied functions.  

2. The amount of code that actually gets Numba-fied out of a given module will be small.  Hence it's not a big deal to pair it with equivalent vectorized code.

3. In many cases we will already have the NumPy routines already written.  

If people agree with this then let's just adopt a warning telling people their code will run significantly slower without Numba, and revisit the dependency issue down the road.

Good idea about a plan for version 1.0.  I think we should feel free to break APIs in the short term.

mmcky:
Just running through open issues. I have pushed a suggestion in branch ``numba_warning`` and will submit a PR for comment to get this issue closed.

PR #140, opened 2015-04-21
Numba version of mc_sample_path/MarkovChain.simulate

oyamad:
This is to follow #137:

I am very sorry, but I have been busier than I had expected. I wanted to get everything prepared as I was going to try a little different implementation against our discussion in some aspects, but I could not have found enough time. Anyway here's a preliminary first draft for discussion, where I am absolutely open to suggestions.

0. The loop version of `np.searchsorted` for jitting (tentatively called `search_cdf`) is pulled out from the main routine, where auto loop-jitting still works (I confirmed this by the html annotation feature; thanks @albop for the information!).

0. `search_cdf` might better be called `searchsorted`, in which case, it should work like `np.searchsorted` (the current code assumes that `v` is always smaller than `cdf[-1]`). (Docsting is yet to be written.)

0. I haven't written code for switching to a NumPy version. The current `mc_sample_path` works with NumPy by only redefining `search_cdf` to be `lambda cdf, v: np.searchsorted(cdf, v, side='right')`, so I was hesitating duplicating the whole code.

0. `cdfs` once computed may be stored in the `MarkovChain` instance. In that case, the code should be placed in the `simulate` method of `MarkovChain`, and `mc_sample_path` will just return `MakovChain(P).simulate(init, sample_size)`.

0. Motivated by a comment by @albop I was thinking of changing the API, for example by giving `MarkovChain` a method `replicate` in addition to `simulate`, as in [LinearStateSpace](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/lss.py#L68).

0. This is a different issue, but I wonder if there would be demand for a version of this function that is specialized for a sparse matrix input. That would use `search_cdf`, too.

### DIFF
--- a/quantecon/mc_tools.py
+++ b/quantecon/mc_tools.py
@@ -92,6 +92,7 @@ from .discrete_rv import DiscreteRV
 from .graph_tools import DiGraph
 from .gth_solve import gth_solve
 from warnings import warn
+from numba import jit
 
 
 class MarkovChain(object):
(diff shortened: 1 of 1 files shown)


mmcky:
Hi @oyamad thanks - no problem - my original thought on workflow was that we could work together on a common branch in the main QuantEcon.py repo called something like ``numba`` or  ``numba-improvements`` in which we could jointly edit and progress towards a final implementation based on your two examples. This is the workflow that I am familiar with in which we can pull the branch down make some edits and commit them to the branch for review etc. Once we are happy with the final implementation we can then issue a PR to the master branch for inclusion of the feature.

Is there a better way to do this?  Can I submit edits directly to ``oyamad:mc_sample_path_numba`` or is this PR effectively a branch? I am happy to learn if that is a better model.

jstac:
@oyamad It's very nice code.  

I think it would be good to have a `replicate` function, if it's not too much work.  I'm not sure about the demand for working with sparse matrices but it seems likely these matrices are common in applied problems.

I'm looking forward to seeing the PR.

jstac:
@mmcky If you have time it would be great if you could implement #133 as part of that `numba_improvements` branch you mention (and close the issue once it's done).

oyamad:
@mmcky Sure let's follow the workflow as you describe.
I think I have succeeded in creating a branch `numba_improvements` in the main QuantEcon.py repo.

Closing.

Issue #143, opened 2015-05-04
[Update] ``Cartesian`` Module 


mmcky:
Review ``cartesian`` module and update as required:

  - [ ] line_profile to see if further ``numbafication`` is useful or not

Current discussion from @albop:
@oyamad , @mmcky : On the cartesian function, the reasoning was that the loops and list comprehensions are made over the number of dimensions, so that they are probably only a few iterations. Appart from the loop over _repeat_1d, it seems likely that everything else is going to stay in python mode, with little performance gain if any. This was just my feeling though, I didn't test it.

  - [ ] update so that numba isn't an explicit requirement (numpy version?)
  - [ ] Incorporate Issue #98

mmcky:
It looks like this has been moved to ``gridtools.py``. 
This should then be applied to ``gridtools.py``

Can ``cartesian.py`` now be safely deleted?

rht:
>  line_profile to see if further numbafication is useful or not

With
```python
from numpy import linspace
x = linspace(0,20,100)

tic = time.time()
prod = cartesian([x,x,x])

correct = True
for i in range(999):
    n = prod[i,0]*100+prod[i,1]*10+prod[i,2]
    correct *= (i == n)
print(time.time() - tic)
```

99.9% happens in a numba-ified function  `_repeat_1d`
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    14                                           @profile
    15                                           def cartesian(nodes, order='C'):
    16                                               '''
    17                                               Cartesian product of a list of arrays
    18                                           
    19                                               Parameters
    20                                               ----------
    21                                               nodes : list(array_like(ndim=1))
    22                                           
    23                                               order : str, optional(default='C')
    24                                                   ('C' or 'F') order in which the product is enumerated
    25                                           
    26                                               Returns
    27                                               -------
    28                                               out : ndarray(ndim=2)
    29                                                   each line corresponds to one point of the product space
    30                                               '''
    31                                           
    32         1         18.0     18.0      0.0      nodes = [np.array(e) for e in nodes]
    33         1          8.0      8.0      0.0      shapes = [e.shape[0] for e in nodes]
    34                                           
    35         1          3.0      3.0      0.0      dtype = nodes[0].dtype
    36                                           
    37         1          2.0      2.0      0.0      n = len(nodes)
    38         1        103.0    103.0      0.0      l = np.prod(shapes)
[...]

Issue #132, opened 2015-04-07
Numpy version to fall back on in lss.py


jstac:
Put in an ordinary numpy version of `simulate_linear_model` to fall back on when numba is not installed.  Better than falling back on the un-jitted current version, which doesn't use any vectorization.

oyamad:
I would like to discuss this issue before updating PR #103 on jitting `gth_solve`.

1.
One approach is that we make it our premise that the user has Numba installed (if he/she doesn't we assume that he/she knows how to install Numba), and we avoid repeating one algorithm with two versions, one of Numba and the other of Numpy.

2.
If we decide to always prepare two versions, one way to code is as follows:
```python
def simulate_linear_model(A, x0, v, ts_length):
    A = np.asarray(A)
    n = A.shape[0]
    x = np.empty((n, ts_length))
    x[:, 0] = x0
    
    if numba_installed:
        _simulate_linear_model_jit(A, v, ts_length, n, x)
        return x
    
    # if not numba_installed
    for t in range(ts_length-1):
        x[:, t+1] = A.dot(x[:, t]) + v[:, t]
    return x

@jit(nopython=True)
def _simulate_linear_model_jit(A, v, ts_length, n, out):
    for t in range(ts_length-1):
        for i in range(n):
            out[i, t+1] = v[i, t]
            for j in range(n):
                out[i, t+1] += A[i, j] * out[j, t]
```

Another would be:
```python
def simulate_linear_model(A, x0, v, ts_length):
    # Numpy version

def _simulate_linear_model_loop(A, x0, v, ts_length):
    # Current version with loops

if numba_installed:
    simulate_linear_model = jit(_simulate_linear_model_loop)
```

The former appears to be better, since it minimizes repetition and we can make explicit the code segment in `nopython` mode (at least for this case). (There may be other possible ways.)

What do you guys think?

jstac:
This is a very interesting discussion.  Here are my thoughts:

First, we've seen what numba can do, it's a great technology, it's under active development and it's accessible to most of our user base (since they'll be installing Anaconda).  Hence we should exploit it extensively.

[...]

PR #144, opened 2015-05-04
Numba improvements

spencerlyon2:
I'm just opening this as a PR so that it is easy to see what has been changed. We can still push to the branch to make more changes to the PR. 

I have _no_ intention of merging this myself.


oyamad:
@spencerlyon2 Thanks for issuing the PR!

Summary of changes and issues to discuss:

0. `gth_solve.py`: The main part is jitted in a new function called `_gth_solve_jit` with nopython mode; if not numba_installed, the previous NumPy version is executed. The test does not cover the NumPy version. See #103 for details ([this](https://github.com/QuantEcon/QuantEcon.py/pull/103#issuecomment-95621781) in particular).

0. `mc_tools.py`:  `mc_sample_path` is autojitted if numba_installed; a NumPy version `mc_sample_path_numpy` is also implemented, which overwrites `mc_sample_path` if not numba_installed. The only difference between the two version is that the former uses the custom `searchsorted`, which is contained in `utilities.py`, while the latter uses `np.searchsorted`. See #137 for details.

0. `utilities.py`: Contains `searchsorted`, which is a jitted and simplified version of `np.searchsorted`; it is an alias of `np.searchsorted` if not numba_installed. This function will no longer be need and should be discarded once `np.searchsorted` is supported by Numba.

0. `numba_import_fail_message` is in `common_messages.py`; `numba_installed ` is defined with `try from Numba import jit except ...` in `external.py`.

0. Issues on `cartesian` are to be discussed in #143.

0. The issue of keeping/preparing a NumPy counterpart for each routine implemented with Numba, one of the main issues in #132. Will we drop support for NumPy version sometime in the future, and if so, at what occasion? (I feel somehow uneasy when I see the duplications...)

0. These may belong to other PRs:
  * Possible changes in the API of `MarkovChain`/`mc_sample_path`: see [this](https://github.com/QuantEcon/QuantEcon.py/issues/137#issuecomment-93534791) and [this](https://github.com/QuantEcon/QuantEcon.py/issues/137#issuecomment-95180890).
  * Support for sparse matrices in `MarkovChain` (I have code for `mc_sample_path` but not yet for `gth_solve`).

0. This is not really important, but the "coverage" as displayed on the top page gets decreased as we write more Numba implemented code, since the `coverage` module inspects the jitted functions to not covered by the tests. See [Code coverage of jitted functions - Google Groups](https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/CqN2zkGb9nA).

mmcky:
@oyamad Excellent summary. I think most of these discussions points are captured in other open issues now - so I don't see any reason not to merge this in. Do you? 

Point 7 (i). I have opened an issue to track this (#146) based on content in #137 
Point 7 (ii). I have opened an issue to track this (#145). 

Point 8. Interesting issue - I agree not entirely important at this stage - but good to be aware of. Thanks for pointing this out.

oyamad:
@mmcky thanks.

I still have a concern about the duplication in `mc_tools.py`. I think `mc_sample_path_numpy` is not necessary and the following is sufficient:

```python
def mc_sample_path(P, init=0, sample_size=1000):
    """
    See Section: DocStrings below
    """
    n = len(P)

    # CDFs, one for each row of P
    cdfs = np.empty((n, n), order='C')  # see issue #137#issuecomment-96128186
    np.cumsum(P, axis=-1, out=cdfs)

    # Random values, uniformly sampled from [0, 1)
    u = np.random.random(size=sample_size)

    # === set up array to store output === #
    X = np.empty(sample_size, dtype=int)
    if isinstance(init, int):
        X[0] = init
    else:
        cdf0 = np.cumsum(init)
        X[0] = searchsorted(cdf0, u[0])

    # === generate the sample path === #
    for t in range(sample_size-1):
        X[t+1] = searchsorted(cdfs[X[t]], u[t+1])

    return X

if numba_installed:
    mc_sample_path = jit(mc_sample_path)
```

If not numba_installed, `searchsorted` works as `np.searchsorted` (with `side='right'`), as defined in `utilities.py`.

mmcky:
@oyamad I see what you are saying now - thanks. I think this is a good idea.  Effectively it boils down to moving the ``numpy`` vs ``numba`` comparison down a level to the now internal ``searchsorted`` utility functions. So we should update the test to compare jitted ``searchsorted`` with the numpy ``searchsorted`` (with ``side="right"``) to ensure both give the same result over time. In your implementation above the rest of ``mc_sample_path`` is exactly the same logically jitted or not.

The only downside I see is in this implementation we won't be able to check the function ``mc_sample_path`` final output between the ``jit`` version and the ``non jit`` version easily - but that would only effectively be checking for jit errors which are probably less of an issue versus logic errors etc.  

@jstac I will leave the final decision with you. But @oyamad suggestion does reduce duplication in this special case. 

**Performance:**
I had a quick look at performance (http://nbviewer.ipython.org/github/mmcky/work-notebooks/blob/master/quantecon/PerformanceCheckMCTools.ipynb). Sorry didn't suppress %timeit text but the graphs in ``Out[7]`` and ``Out[15]`` suggests that ``numba`` function is best for anything greater than a sample size of ~70. ``localB`` is what would happen in the case of no ``numba`` installed and is the slowest. Interestingly the non-numpy ``looped`` version of searchsorted runs faster even in pure python which is ``localA``. I am not sure why this is.

oyamad:
@mmcky Thank you for the performance comparison. It is very interesting.
I suspect that `qe.numpy` and `localB` would be faster than `localA` for matrices of larger size.

mmcky:
@oyamad Yeah I think you are right. Last night I had a quick look in our library for a simple utility for producing a ``Transition Matrix`` of size n. I didn't see one - I could produce some performance checks through the matrix dimensions this afternoon.

mmcky:
@oyamad Indeed. I added a really simple test which is just using a random matrix ``P`` with each row normalised  to 1. ``localB`` is much better than ``localA``. (http://nbviewer.ipython.org/github/mmcky/work-notebooks/blob/master/quantecon/PerformanceCheckMCTools.ipynb)

jstac:
@oyamad @mmcky You've done a lot of work.  It's looking really good. I like the look of the code and I'm keen to merge this.  

I agree with @oyamad 's suggestion about cutting `mc_sample_path_numpy`.  There's no need for it.  As long as the home grown version of `searchsorted` is properly tested --- which it is --- there's no need to keep `mc_sample_path_numpy` just for the sake of testing.

So I propose that we adopt @oyamad 's suggestion, cut this function, and then merge.

oyamad:
I deleted `mc_sample_path_numpy` and `test_mc_sample_path_functions`. Now no test is there for `mc_sample_path`.

jstac:
Thanks for all this work.  I'll merge now and at the same time open an issue to add a test for `mc_sample_path`.  Great to have this merged!

spencerlyon2:
Alt Text

Issue #98, opened 2014-11-18
Accessing the cartesian routine


oyamad:
Hi,
I wanted to use [`cartesian`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/cartesian.py), but found it a bit tricky to access it.

Does not work:
```python
import quantecon as qe
qe.cartesian([[0, 1], [0, 1], [0, 1]])
```
It raises `AttributeError: 'module' object has no attribute 'cartesian'`.
(The same goes for `qe.cartesian.cartesian([[0, 1], [0, 1], [0, 1]])`.)

Works:
```python
import quantecon.cartesian
quantecon.cartesian.cartesian([[0, 1], [0, 1], [0, 1]])
```

Is this as intended?
It is not mentioned in the [documentation](https://quanteconpy.readthedocs.org/en/latest/tools/cartesian.html).

spencerlyon2:
Seems like this is resolved. Can anyone confirm and close?

albop:
The cartesian package has to be renamed to `gridtools` or `grid_tools` as @jstac has suggested. Both seem perfect to me.

spencerlyon2:
I'd vote for `gridtools`.

jstac:
Fine by me.

PR #149, opened 2015-05-07
REF: renamed cartesian.py as gridtools.py

albop:
Small pull request to close issue #98  .
I haven't changed the files MANIFEST and SOURCES.txt, as I don't know very well how/when they are generated. Do they really need to be versioned ?


jstac:
@albop Thanks!  Merging.

PR #151, opened 2015-05-08
Update warning message if numba import fails

mmcky:
See issue #133


jstac:
@mmcky Thanks for getting this done.  I'm going to merge and we can tweak later if we want to.

Issue #153, opened 2015-05-20
Option to supply a random seed


oyamad:
Given the discussion in #147, I would like to discuss the way how to give the user the option to supply a random seed; we better fix a common interface and build a module that provides that interface.

A simple form would be something like this:

``` python
def some_function(main_arguments, random_state=None):
    """
    Parameters
    ----------
    ...

    random_state : scalar(int) or np.random.RandomState

    """
    if random_state is None or isinstance(random_state, int):
        random_state_ = np.random.RandomState(random_state)
    elif isinstance(random_state, np.random.RandomState):
        random_state_ = random_state
    else:
        raise ValueError('some message')

    # Generate a random number
    u = random_state_.random_sample()
```

For example, `scikit-learn` has a [guideline on random numbers](http://scikit-learn.org/stable/developers/contributing.html#random-numbers) and a common routine [check_random_state](https://github.com/scikit-learn/scikit-learn/blob/bb39b49/sklearn/utils/validation.py#L503) that converts the input into a `np.random.RandomState` instance.

mmcky:
I think this is a good approach. I like the idea of using random state object as in a recent test I had to ensure numpy's internal state was consistent before running two different functions for comparison which required setting the seed twice. This would avoid this problem. We should wrap the instance checks into a utility or support function as ``skikit-learn`` has done.

mmcky:
@oyamad I have pushed a branch called ``reorg-utilities`` within which I have included a copy of the ``scikit-learn`` function for checking random seeds as part of a broader reorg of the quantecon utilities. I think for future growth it is probably better they are contained in a subpackage which allows for categorization of tools while retaining a common api. 

```python
from quantecon.util import check_random_seed
```

oyamad:
@mmcky Thanks! I have incorporated it into the `mc_random` branch (#154).

oyamad:
We need to rewrite `from quantecon.utilities import searchsorted` in `test_utilities.py` to import `util.array.searchsorted`. We may rename the file to `test_array.py`. In which directory should that file go?

mmcky:
@oyamad Thanks for pointing that out. 

  1. Migrated ``test_utilities.py`` to ``util\tests\test_array.py``
  2. Also re: Reference you can use ``from quantecon.util import searchsorted`` as the ``__init__.py`` controls the api and I have imported it to the ``util`` namespace. 

[...]

Issue #145, opened 2015-05-05
Sparse Matrix Implementations


mmcky:
Support for sparse matrices in MarkovChain 

Code available for for mc_sample_path (@oyamad) but not yet for gth_solve.

Issue #139, opened 2015-04-17
Improve online documentation


jstac:
The documentation would be better if functions and classes all had at least one example and possibly more.  E.g., see the examples here:

http://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.expon.html

There is a section devoted to examples.  In the NumPy / SciPy docs, I find this section is often more useful than any other kind of documentation.  I propose we push towards improving our documentation in this way.

oyamad:
It would be helpful if we have a style guide for the Examples section (just for a reference point).

We have a 2x2 table of choices regarding the following:
* whether include the import statement;
* if random, whether specify `random_state`.

Example:

1.
With `import`

```python
>>> import quantecon as qe
>>> qe.random.sample_without_replacement(5, 3)
array([2, 3, 4])
```

2.
Without `import`

3.
With `random_state`

```python
>>> qe.random.sample_without_replacement(5, 3, random_state=1234)
array([0, 2, 1])
```

4.
Without `random_state`

PR #154, opened 2015-05-20
Random MarkovChain

oyamad:
~~(This is not yet ready to merge.)~~

I wrote some functions that generate random `MarkovChain` or stochastic matrix. I hope this module will be useful at least for developers and probably for users as well.

0. `random_probvec` generates random probability vectors (whose entries are nonnegative and sum to one). (Note that just generating random numbers from [0, 1] and divide them by the sum would generate probability values that are skewed toward 0.5.)

0. If one wants an n x n stochastic matrix with a small number of states with positive probabilities for each row, where the number k of such states must be common across states here, then k integers have to be chosen randomly from 0, 1, ..., n-1, which is done by `random_choice_without_replacement`.

0. `random_markov_chain` will support sparse matrix format once `MarkovChain` does.

0. `random_choice_without_replacement(n, k, num_trials=None)` is equivalent to `np.random.choice(n, k, replace=False)` and is much (100x?) faster with Numba.

0. `random_choice_without_replacement` may be included in `utilities.py`. Its name seems too long. Anybody have any idea?

0. API regarding `random_state` will be modified once we fix a common interface (I opened a new issue #153 to discuss this).


mmcky:
@oyamad Neat - It will be good to have this for easily creating some matrix for testing or demonstration etc. 

Regarding the name. What do you think about just mimicking the np structure by adding a random subpackage or module? We could then use the hierarchy of the package to capture the essence contained in the name ``random_choice_without_replacement`` as is done in numpy. 

```python
import quantecon as qe
qe.random.choice(n, k, replace=False)
```

oyamad:
@mmcky Thank you for your comments. It is a good idea to make it parallel to NumPy, which would make the names easier to remember.

Just for information:
* The same functionality in the standard library is called [`random.sample`](https://docs.python.org/2.7/library/random.html#random.sample). (In fact I took its code and jitted it with an additional outer loop which is for the case where `num_trials` is not `None`.)
* I also found scikit-learn has the same functionality which is named [`random.sample_without_replacement`](http://scikit-learn.org/stable/developers/utilities.html#efficient-random-sampling) [[source](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/_random.pyx#L227)].

mmcky:
@oyamad Thanks for the additional info - good to know. Given how much ``numpy`` is used in this package - I think ``numpy`` probably seems to make the most sense to mimic. Having said that I like the ``scikit-learn`` use of ``sample`` rather than ``choice`` but not really a big deal. 

Happy to help re-package this as a subpackage if you like.

oyamad:
@mmcky Please feel free to edit the branch, thanks!

Maybe we would better fix the interface for `random_state` first.

oyamad:
* [ ] Sparse matrix support: see #145
* [ ] `random_choice_without_replacement`: rename and move to `random` subpackage or module
* [ ] `random_state`: common interface: see #153

oyamad:
I think this PR is ready for merging. Importantly, this contains the feature of `random_state` option with `util.check_random_state`, so that after this we can start working on adding the option to other functions/classes (unless anyone has different thoughts).

(Sparse matrix support and renaming and moving `random_choice_without_replacement` can be postponed.)

mmcky:
Hi @oyamad - Sorry for my delay. Just reacquainting myself with this issue. This is looking good. Are you still happy for this to be merged? I see that nosetests is reporting all OK for ``test_random_mc.py``.

mmcky:
Does this need to be updated prior to merge?
* [ ] `random_choice_without_replacement`: rename and move to `random` subpackage or module

mmcky:
@oyamad I am happy to move code around to fulfil the checkbox above. 
Let's hold off on sparse matrix support and open a separate issue so this can be merged (as you suggest). 

I wanted to run a few things by you before I implement this:
 
1. I have previously indicated that we should probably emulate numpy given its strong use throughout the package. However I think ``sample`` is more aligned with common terminology rather than the numpy use of ``choice``. Would you mind if we changed to ``sample``?
2. I see a number of useful utilities in random_mc. I propose to move ``random_choice_without_replacement`` to ``sample(replace=False)`` where replacement is an option to a sample function that resides in ``util/random.py``. We can then fix up the API to call qe.random.sample(replace=False) for ``random_choice_without_replacement``
3. Should we also move ``random_probvec`` to ``util/random.py`` with usage ``qe.random.probvec()``
4. Should we also move ``random_stochastic_matrix`` to ``util/random.py`` with usage ``qe.random.stochastic_matrix()``? 

Thanks

oyamad:
@mmcky

1, 2: Please do go ahead as you suggest.

3, 4: @jstac once mentioned having a subpackage (say `markov`) that collects routines regarding markov chains and markov decision processes. `random_probvec` and `random_stochastic_matrix` might be put there as `qe.markov.random.probvec` (or `qe.markov.random.probability_vector`) and `qe.markov.random.stochastic_matrix`. (I have no strong opinion though.)

jstac:
I support having a `markov` subpackage.  

I feel that names like `qe.markov.random.probvec` etc are too verbose.  I propose that we put this function in somewhere like `util` and perhaps make it visible in the top level `qe` namespace.  I'm not sure about `random_stochastic_matrix`.  Either subpackage would be fine.

mmcky:
@oyamad @jstac I have moved ``random_probvec`` and ``random_sample_without_replacement`` to the ``util.random`` file. These are now named ``probvec`` and ``sample_without_replacement`` within this file. These functions have also been imported to top level name space as the name you have used previously. Therefore these functions can be called by using ``qe.random_probvec(...)`` or using the hierarchy of the package structure ``qe.random.probvec()``. 

I have migrated ``random_mc.py`` to a new ``markov`` sub package and update internal calls to the utility functions ``random_probvec`` etc. The associated tests ``test_random_mc.py`` have been migrated to ``markov/tests/test_random_mc.py`` and internal references updated. 

Given the migration to the ``markov`` sub package do you think we still need to postpend ``_mc`` on the files?

oyamad:
@mmcky thanks!
(I am not sure, but maybe we don't need both `random_sample_without_replacement` and `random. sample_without_replacement`, and just `random.sample_without_replacement` is enough? The same for `probvec`.)

Right, we don't need `_mc` anymore. And are we going to migrate `mc_tools.py` to `markov`, too? Then we might consider changing `mc_tools.py` to, say, `markov_chain.py`, and functions like `mc_sample_path` might be renamed as well (perhaps with exporting them under the previous names for backward compatibility)?

oyamad:
@mmcky Could you also update the Example sections? Thanks!

mmcky:
@oyamad Sure - good reminder. Thanks.

mmcky:
@oyamad I only came across one instance of ``mc_tools`` in ``lucas_stokey.py``. This has been updated and will now run.

mmcky:
@oyamad I think the reorganisation is complete. When you have some time - it would be great if you could have a quick look at the branch. Once that is done we should go ahead and get approval to merge.

oyamad:
@mmcky Many thanks!

I will have a closer look tomorrow, but one quick comment: maybe `gth_solve.py` also should be put in `markov`.

mmcky:
@oyamad Great thanks. I have moved ``gth_solve`` to the markov subpackage and updated relevant ``__init__.py`` files and the associated test suite.

oyamad:
@mmcky Thanks!

It seems that `MarkovChain` etc can be called by both `qe.MarkovChain` and `qe.markov.MarkovChain`. Which do we recommend the user to use? (Functions/classes in `models` can only be called with `qe.models.*`.)

oyamad:
I made a few changes:

0. Examples sections updated (subject to change after we fix a style guide; see #139).

0. I replaced `random_probvec` and `random_sample_without_replacement` with `probvec` and `sample_without_replacement`; I think at least in this case there is no need to give two names to one function. (I didn't touch the `__init__.py` files.)

0. I moved `test_sample_without_replacement_*` from `markov` to `util/tests/test_random.py`.
   (There, `from quantecon.random import sample_without_replacement` did not work, so I imported with `from quantecon.util.random import sample_without_replacement`. Is this correct?)

mmcky:
@oyamad These are excellent points. I have discussed with @jstac today and we have decided to do a bit more reorganisation. The ``util`` subpackage is probably best geared towards internal ``quantecon`` utilities that support the package infrastructure. This includes functions like the ``check_random_seed`` and special testing functions etc. Therefore I have moved ``probvec`` and ``sample_without_replacement`` to a ``random`` subpackage. These are topical functions that we can use via the package structure and should be accessed by

```python
qe.random.probvec()
qe.random.sample_without_replacement()
```

Thanks for updating the examples section, I had only focused on the ``examples/`` folder and neglected the docstrings. I will check those docstrings now based on this new round of revisions. 

I have moved associated tests

oyamad:
@mmcky Very good, thanks. I am now very happy with this PR.

This is an independent issue probably for another PR: some other files might also be moved to `util`, such as
* common_messages.py
* external.py
* timing.py

mmcky:
Great thanks for the feedback @oyamad. Good idea re: migration to ``util`` will add to a separate issue. 
I have also migrated ``tauchen.py`` to the ``markov`` package

cc7768:
> I have also migrated `tauchen.py` to the `markov` package

What are the chances we can rename the `tauchen.py` file to something along the lines `markov_approximation.py` and then name the `approx_markov` function `tauchen`? This has always made more sense to me, and I think we've talked about it, but I don't remember what discussion has been had.

mmcky:
@cc7768 Good idea. I will do this revision as part of this PR.

mmcky:
Merging this issue now. Thanks @oyamad for this addition. Thanks everyone for comments and suggestions.

PR #161, opened 2015-07-29
Add the same test case as in the Julia version

oyamad:
Add the same test case as in QuantEcon/QuantEcon.jl#62.


spencerlyon2:
Looks great, thanks!

Issue #146, opened 2015-05-05
Update MarkovChain with ``replicate `` method and Future Numba Improvements


mmcky:
From @albop

>> Now, I believe that it would be useful to have the function doing the actual simulaion run in nopython entirely, so that one can loop around it. Is it really useful to keep the ability to supply init, either as a point or as a distribution ? Eventually it gets converted into a scalar index, which seems easy to do outside of the function.

>> From an api perspective, what about a function like:

>> simulate_markov_chain(P, init, sample_size, n_exp=None) which would simulate a markov chain 

>> characterized by matrix P over sample_size periods and:
        perform one single simulation starting at init if init is an integer and n_exp is None
        simulate many draws if init is a vector (len(init) draws) or n_exp is an integer (n_exp draws)

From @jstac 

>> @albop "Is the reason why you want to speed up the code, because you are simulating over a very long period ? Or is it because, you want to run many simulations ? (like 100000)"

>> This routine is orientated to the former. We could also add one for the latter. In that case vectorization would probably play a bigger role.

From @oyamad 

>> Motivated by a comment by @albop I was thinking of changing the API, for example by giving MarkovChain a method replicate in addition to simulate, as in LinearStateSpace.

oyamad:
I have pushed 95627f2d73e3f50e6f2fce5d465a01f8b8d8461f for discussion on this matter.

0. First of all, I think `cdfs` should be stored once computed, avoiding repeating computation each time. It is now stored as `MarkovChain.cdfs`.

0. I made some changes to have the methods of `MarkovChain` look similar or consistent with the corresponding methods of [`LinearStateSpace`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/lss.py#L63).

  0. I added the `replicate` method. It returns `num_reps` observations of the state at time `T`.

  0. I changed the order and the names of the arguments of `simulate` to be consistent with `LinearStateSpace.simulate`. I don't know how much benefit there is by this (the cost is obvious).

0. `init` now does not accept a distribution.

0. Having written up  the code, I started thinking the proposal by @albop would be much better. Namely, design simulate as `simulate(ts_length, init=None, num_reps=None)`, which would return one sample path of length `ts_length` by default, and if `init` and/or `num_reps` is not None, return `len(init)` or `num_reps` sample paths (and discard `replicate`). A concern in this case is that the returned array, which is of shape `(num_reps, ts_length)`, could become huge.

0. I haven't written tests.

[...]

PR #166, opened 2015-08-08
MarkovChain.simulate API change; random_state option added

oyamad:
This follows the discussions in #146.
(I hope I did "rebase" in the right way.)

0. I propose to change the `simulate` method as `simulate(ts_length, init=None, num_reps=None, random_state=None)`; see https://github.com/QuantEcon/QuantEcon.py/issues/146#issuecomment-109540676.
   * `init` now does not accept a distribution; see https://github.com/QuantEcon/QuantEcon.py/issues/146#issuecomment-117006253 and https://github.com/QuantEcon/QuantEcon.py/issues/146#issuecomment-117425074.

0. `random_state` option is added to `MarkovChain.simulate` (not to `MarkovChain.__init__`).

0. [`test_mc_sample_path`](https://github.com/QuantEcon/QuantEcon.py/blob/3ba9f53e3064385f2f852b820fa3ff2cef5a7083/quantecon/markov/tests/test_core.py#L259) is added with `random_state` specified.

### DIFF
--- a/quantecon/markov/core.py
+++ b/quantecon/markov/core.py
@@ -91,19 +91,7 @@ from .gth_solve import gth_solve
 from ..graph_tools import DiGraph
 
 # -Check if Numba is Available- #
-<<<<<<< HEAD
-<<<<<<< HEAD
-from ..util import searchsorted, numba_installed, jit
-=======
-from .external import numba_installed, jit
-
-from .util import searchsorted, check_random_state
->>>>>>> `random_state` option, as well as tests, added
-=======
-from ..external import numba_installed, jit
-
-from ..util import searchsorted, check_random_state
->>>>>>> Import statements corrected
+from ..util import searchsorted, check_random_state, numba_installed, jit
 
 
 class MarkovChain(object):
--- a/quantecon/markov/core.py
+++ b/quantecon/markov/core.py
@@ -313,12 +313,13 @@ class MarkovChain(object):
         random_state = check_random_state(self.random_state)
 
         try:
-            num_reps = len(init)  # init is an array; make num_reps not None
+            k = len(init)  # init is an array
+            num_reps = k  # make num_reps not None
             init_states = np.asarray(init, dtype=int)
-        except:  # init is a scalar(int) or None
+        except TypeError:  # init is a scalar(int) or None
             k = 1 if num_reps is None else num_reps
             if init is None:
-                init_states = np.random.randint(self.n, size=k)
+                init_states = random_state.randint(self.n, size=k)
             elif isinstance(init, int):
                 init_states = np.ones(k, dtype=int) * init
             else:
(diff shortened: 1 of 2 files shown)


oyamad:
Regarding testing methods/functions with randomness (i.e., `MarkovChain.simulate` and `mc_sample_path`), I wonder what we want to test. The current [`test_mc_sample_path`](https://github.com/QuantEcon/QuantEcon.py/blob/3ba9f53e3064385f2f852b820fa3ff2cef5a7083/quantecon/markov/tests/test_core.py#L259) verifies that given a certain seed value, `mc_sample_path` returns a certain sequence of integers (states). Is it what should be tested?

If `MarkovChain.simulate` and `mc_sample_path` are properly implemented,

* if the markov chain is irreducible, for a well-behaved seed value (i.e., a seed value that generates uniformly distributed numbers), along a sample path generated by these functions (with long enough length) the frequency should be close to the stationary distribution; and

* if in addition the markov chain is aperiodic, for a large enough period `T`, having these functions generate many enough sample paths the frequency distribution at `T` should be close to the stationary distribution.

[`test_mc_sample_path_lln`](https://github.com/QuantEcon/QuantEcon.py/blob/241193b288faa69f8c07a62e9fb3201e65dc5c7d/quantecon/markov/tests/test_core.py#L296) and [`test_simulate_ergodicity`](https://github.com/QuantEcon/QuantEcon.py/blob/241193b288faa69f8c07a62e9fb3201e65dc5c7d/quantecon/markov/tests/test_core.py#L236) verify these properties to hold. Any thoughts on this issue?

oyamad:
In case `init` is an array and `num_reps` is supplied, the current implementation ignores `num_reps` and generates `len(init)` sample paths:

```python
>>> # If init is an array, num_reps is ignored
... mc.simulate(10, init=[0, 1], num_reps=3)
array([[0, 0, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 0]])
```

But would it be better to generate `len(init) * num_reps` sample paths like the following, not ignoring `num_reps`?

```python
[[0, ...],
 [1, ...],
 [0, ...],
 [1, ...],
 [0, ...],
 [1, ...]]
```

mmcky:
@oyamad I see this PR has some merge conflicts - shall I rebase? 

    But would it be better to generate len(init) * num_reps sample paths like the following, not ignoring num_reps?

This seems like sensible behaviour - rather than ignoring ``num_reps``.

oyamad:
I did a rebase and made a forced push. (I wonder if this is a good conduct...)
I am going to change the behavior for `init` + `num_reps`.

mmcky:
Great - thanks @oyamad. I will wait for the update in behaviour and then we can merge this PR. 
I am always a bit hesitant to force anything in git - but I don't see this being a big problem when working in a branch.

oyamad:
Now `simulate` works for array `init` and int `num_reps` as:

```python
>>> mc.simulate(10, init=[0, 1], num_reps=3, random_state=1234)
array([[0, 0, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 0, 0, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 0, 1, 1],
       [0, 0, 1, 1, 1, 0, 0, 0, 1, 1],
       [1, 1, 0, 1, 1, 1, 0, 1, 1, 0]])
```

mmcky:
Thanks @oyamad are you happy for this to be merged?

    test_mc_sample_path_lln and test_simulate_ergodicity verify these properties to hold. Any thoughts on this issue?

I think these are good additions to the tests suite.

oyamad:
@mmcky Yes, thanks!

Issue #165, opened 2015-08-06
Migrate some modules to util subpackage


mmcky:
Some files that should be moved to `util`
* common_messages.py
* external.py
* timing.py

oyamad:
`import warnings` in [lss.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/lss.py#L12) was forgotten to move to [external.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/external.py#L22)?

mmcky:
@oyamad Thanks. Working on a ``update-util`` branch now. I have added to ``external.py``

mmcky:
Closing as any comments can be added to the PR.

Issue #40, opened 2014-08-04
ENH: make calls to `compute_fixed_point` allocate less memory


sglyon:
The idea is quite simple: whenever we call `compute_fixed_point(T, ...)` we should update the definition of `T` so that any temporary arrays can be optionally supplied by the caller. See the new `compute_lt_price` function in the `LucasTree` class of quantecon/models/lucastree.py for an example (there we avoid allocating `Tf` on each iteration).

PR #171, opened 2015-08-24
MDP

oyamad:
I finally finished coding the Markov Decision Process (MDP) class. There are a couple of issues to discuss.

0. An `MDP` instance can be created in two ways, in which the transition matrix is

   * a 3-dimensional array defined on S x A x S ("product formulation")
   * a 2-dimensional array defined on SA x S ("state-action pairs formulation")

   where S is the state space, A the action space, and SA the set of feasible state-action pairs. The former should be straightforward; the latter requires additional information about the correspondence between the indices of SA and the actual state-action pairs. See the following example:

   ```python
   >>> s_indices = [0, 0, 1]  # State indices
   >>> a_indices = [0, 1, 0]  # Action indices
   >>> R = [5, 10, -1]
   >>> Q = [(0.5, 0.5), (0, 1), (0, 1)]
   >>> beta = 0.95
   >>> mdp = MDP(R, Q, beta, s_indices, a_indices)
   ```

   Here, S = {0, 1}, A = {0, 1}, and SA = {(0, 0), (0, 1), (1, 0)} (action 1 is infeasible at state 1). The correspondence is represented by `s_indices` and `a_indices`. I wonder whether this is an "optimal" implementation strategy.

0. Complications are basically kept in `__init__` (so it is very long), to keep other methods simple and more or less intuitive.

0. Any suggestions for improvement in particular on variable/method/function names (and of course for other things too) will be appreciated.

0. I tried to follow the ideas in #40 and #41. I wonder if I did it right. The method [`operator_iteration`](https://github.com/QuantEcon/QuantEcon.py/blob/ccd9e24bd829bc57d517056664b486790aff2c44/quantecon/markov/mdp.py#L542) may be replaced with `compute_fixed_point`, whereas the latter does not exactly match here.

0. Results of `solve` are stored in [`MDPSolveResult`](https://github.com/QuantEcon/QuantEcon.py/blob/ccd9e24bd829bc57d517056664b486790aff2c44/quantecon/markov/mdp.py#L790). Are there other important informations to store there?

0. The policy iteration algorithm is usually initialized with a *policy function*, while the `policy_iteration` method here requires an initial *value function*, to be consistent with the other solution methods.

0. Deals only with infinite horizon. `beta=1` is not accepted. (Do we want a finite horizon algorithm?)

0. I also added [`random_mdp`](https://github.com/QuantEcon/QuantEcon.py/blob/ccd9e24bd829bc57d517056664b486790aff2c44/quantecon/markov/random.py#L155). I hope this will be useful for performance tests.

I also added some notebooks on `MDP` to [QuantEcon.site](https://github.com/QuantEcon/QuantEcon.site/tree/mdp/_static/notebooks/inwork/mdp):

* [Getting started with simple examples](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_intro.ipynb)
* [Theoretical background and implementation details](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_theory.ipynb)
* Examples
  * [Automobile replacement](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_ex_rust96.ipynb)
  * [Discrete optimal growth model](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_ex_optgrowth.ipynb)

I have some more examples [here](https://github.com/oyamad/mdp).

### DIFF
--- a/docs/source/markov.rst
+++ b/docs/source/markov.rst
@@ -8,3 +8,4 @@ Markov
    markov/core
    markov/gth_solve
    markov/random
+   markov/mdp
--- a/docs/source/markov.rst
+++ b/docs/source/markov.rst
@@ -8,4 +8,4 @@ Markov
    markov/core
    markov/gth_solve
    markov/random
-   markov/mdp
+   markov/ddp
--- a/docs/source/markov/ddp.rst
+++ b/docs/source/markov/ddp.rst
@@ -1,7 +1,7 @@
-mdp
+ddp
 ===
 
-.. automodule:: quantecon.markov.mdp
+.. automodule:: quantecon.markov.ddp
     :members:
     :undoc-members:
     :show-inheritance:
--- a/docs/source/markov/mdp.rst
+++ b/docs/source/markov/mdp.rst
@@ -1,0 +1,7 @@
+mdp
+===
+
+.. automodule:: quantecon.markov.mdp
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/quantecon/markov/__init__.py
+++ b/quantecon/markov/__init__.py
@@ -5,5 +5,6 @@ Markov Chain SubPackge API
 from .core import MarkovChain
 from .core import mc_compute_stationary, mc_sample_path 			#-Future Deprecation-#
 from .gth_solve import gth_solve
-from .random import random_markov_chain, random_stochastic_matrix
+from .random import random_markov_chain, random_stochastic_matrix, random_mdp
 from .approximation import tauchen
+from .mdp import MDP
--- a/quantecon/markov/__init__.py
+++ b/quantecon/markov/__init__.py
@@ -5,6 +5,7 @@ Markov Chain SubPackge API
 from .core import MarkovChain
 from .core import mc_compute_stationary, mc_sample_path 			#-Future Deprecation-#
 from .gth_solve import gth_solve
-from .random import random_markov_chain, random_stochastic_matrix, random_mdp
+from .random import random_markov_chain, random_stochastic_matrix, \
+    random_discrete_dp
 from .approximation import tauchen
-from .mdp import MDP
+from .ddp import DiscreteDP
--- a/quantecon/markov/ddp.py
+++ b/quantecon/markov/ddp.py
@@ -1,15 +1,15 @@
 """
-Filename: mdp.py
+Filename: ddp.py
 
 Author: Daisuke Oyama
 
-Module for solving Markov decision processes (MDP) with finite states
-and actions.
+Module for solving dynamic programs (also known as Markov decision
+processes) with finite states and actions.
 
-Markov Decision Processes
--------------------------
+Discrete Dynamic Programming
+----------------------------
 
-A finite Markov decision process consists of the following components:
+A discrete dynamic program consists of the following components:
 
 * finite set of states :math:`S = \{0, \ldots, n-1\}`;
 * finite set of available actions :math:`A(s)` for each state :math:`s
@@ -78,7 +78,8 @@ equation, or the unique fixed point of the Bellman operator, and that
 Solution Algorithms
 -------------------
 
-The `MDP` class currently implements the following solution algorithms:
+The `DiscreteDP` class currently implements the following solution
+algorithms:
 
 * value iteration;
 * policy iteration;
@@ -115,16 +116,15 @@ from .core import MarkovChain
 from ..util import numba_installed, jit
 
 
-class MDP(object):
+class DiscreteDP(object):
     """
-    Class for dealing with a Markov decision process (MDP) with finite
-    states and actions.
+    Class for dealing with a discrete dynamic program.
 
-    There are two ways to represent the data for instantiating an `MDP`
-    object. Let n, m, and L denote the numbers of states, actions, and
-    feasbile state-action pairs, respectively.
+    There are two ways to represent the data for instantiating a
+    `DiscreteDP` object. Let n, m, and L denote the numbers of states,
+    actions, and feasbile state-action pairs, respectively.
 
-    1. `MDP(R, Q, beta)`
+    1. `DiscreteDP(R, Q, beta)`
 
        with parameters:
 
@@ -137,7 +137,7 @@ class MDP(object):
        next period is `s'` when the current state is `s` and the action
        chosen is `a`.
 
-    2. `MDP(R, Q, beta, s_indices, a_indices)`
+    2. `DiscreteDP(R, Q, beta, s_indices, a_indices)`
 
        with parameters:
 
@@ -216,7 +216,7 @@ class MDP(object):
 
     * Discount factor 0.95
 
-    **Creating an `MDP` instance**
+    **Creating a `DiscreteDP` instance**
 
     *Product formulation*
 
@@ -226,7 +226,7 @@ class MDP(object):
     >>> R = [[5, 10], [-1, -float('inf')]]
     >>> Q = [[(0.5, 0.5), (0, 1)], [(0, 1), (0.5, 0.5)]]
     >>> beta = 0.95
-    >>> mdp = MDP(R, Q, beta)
+    >>> ddp = DiscreteDP(R, Q, beta)
 
     (`Q[1, 1]` is an arbitrary probability vector.)
 
@@ -240,13 +240,13 @@ class MDP(object):
     >>> R = [5, 10, -1]
     >>> Q = [(0.5, 0.5), (0, 1), (0, 1)]
     >>> beta = 0.95
-    >>> mdp = MDP(R, Q, beta, s_indices, a_indices)
+    >>> ddp = DiscreteDP(R, Q, beta, s_indices, a_indices)
 
     **Solving the model**
 
     *Policy iteration*
 
-    >>> res = mdp.solve(method='policy_iteration', v_init=[0, 0])
+    >>> res = ddp.solve(method='policy_iteration', v_init=[0, 0])
     >>> res.sigma  # Optimal policy function
     array([0, 0])
     >>> res.v  # Optimal value function
@@ -256,7 +256,7 @@ class MDP(object):
 
     *Value iteration*
 
-    >>> res = mdp.solve(method='value_iteration', v_init=[0, 0],
+    >>> res = ddp.solve(method='value_iteration', v_init=[0, 0],
     ...                 epsilon=0.01)
     >>> res.sigma  # (Approximate) optimal policy function
     array([0, 0])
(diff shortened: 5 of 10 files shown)


jstac:
Wow, this is great!!  I have to spend so time with the code but the implementation looks outstanding.

The notebooks are very helpful too.  

This is a really big contribution to open source economics.

robertdkirkby:
Few points on the 'modified policy iteration' that may or may not be helpful (hopefully they are). I caution that they are all based on my experience with the stochastic version of this algorithm and I am not sure which will carry over exactly to the deterministic.

From speed perspective --- and you may already do this, I was not sure from the pseudocode and don't write python myself so didn't want to go digging --- once you calculate the policy indexes sigma, rather than calculate r(sigma) at each of your k iterations in step 3, you can calculate it once at the beginning of step 3 as a matrix and then just reuse for each of the k iterations. In my experience this provides a nono-trivial speed boost.

I suspect the following may not be true of the deterministic case but you might want to run a test to be sure. In the stochastic MDP extension of your current algorithm it is possible to contaminate everything with -Infs (if your final true v has some -Infs and your initial v0 doesn't). This has happened to me in practice. This is easily avoided by also checking that span(u-v) is finite before you do step 4. Again, I don't think this will happen in the deterministic case, but just to mention it.

Another thing I would mention based on my experience with the stochastic version of your modified policy iteration algorithm. While the value function iteration is guaranteed to converge if you set iter_max high enough (thanks to contraction mapping property) this is not true of the modified policy iteration algorithm if your k is big relative to grid size (of S and A). This is something I have found by experience. It might therefore be worth adding an internal option that after a certain number of tries, if convergence has not yet been reached then the modified policy iteration algorithm will just switch over automatically over to doing value function iteration (ie. you might perhaps have iter_max/2 with modified policy iteration, and then from there to iter_max with value fn iteration). That way the user can be certain that if they just keep increasing iter_max they will eventually get convergence.

(ps. I haven't contributed anything to quantecon, but I 'watch' and this is my first comment here)

oyamad:
Dear @robertdkirkby: Thank you for your comments, and welcome to quantecon!

1.
> once you calculate the policy indexes sigma, rather than calculate r(sigma) at each of your k iterations in step 3, you can calculate it once at the beginning of step 3 as a matrix and then just reuse for each of the k iterations.

Let me explain what I do in the code. `sigma` in step 2 and `u` in step 3 (as described [here](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_theory.ipynb#Modified-policy-iteration)) are actually computed simultaneously. This `sigma` is used to generate the operator `T_sigma` using `R_sigma` and `Q_sigma` computed once, and the same `T_sigma` is used for k times without recomputing [here](https://github.com/QuantEcon/QuantEcon.py/blob/ccd9e24bd829bc57d517056664b486790aff2c44/quantecon/markov/mdp.py#L755), if I understand correctly. But maybe I am missing something important.

2.
> it is possible to contaminate everything with -Infs (if your final true v has some -Infs and your initial v0 doesn't).

This can happen when the input reward is such that for some `s`, `R[s, a] = -inf` for all `a`. I implicitly assumed the use will not give such data. Maybe we should check that this is not the case.

3.
> While the value function iteration is guaranteed to converge if you set iter_max high enough (thanks to contraction mapping property) this is not true of the modified policy iteration algorithm if your k is big relative to grid size (of S and A).

Thank you for reminding me of this, I have been forgetting that in the proof of convergence of modified policy iteration, I needed the assumption that `v^0` be such that `T v^0 >= v^0` (otherwise monotonicity of `T` would have no bite). This condition is satisfied in particular when `v^0 = v_sigma` for some feasible policy `sigma` (for example the policy that maximizes the one-shot reward function) or when `v^0` is very small that `v^0 <= min_(s, a) r(s, a) / (1 - beta)`. I think I will change the default `v_init` (when it is not supplied) to either of these, and correct the notebook. Thanks!

oyamad:
I chose to set `v_init` for modified PI to `min_(s, a) r(s, a) / (1 - beta)`.

Any suggestions are welcome for this and for the other methods as well.

mmcky:
Due to ``util`` changes to close #165 there are some small merge conflicts that will need to be fixed in this branch prior to merge.

robertdkirkby:
Thanks @oyamad 

1. Sounds like you already have this covered.

2. I would strongly advocate that you allow for r() to take -Inf values. Below are three examples of why it would be useful (the second is really just an explicit example of the third). Before the examples though I would highlight that it is quite easy to do: the value function iteration requires no modification whatsoever, the policy iteration method could check for r()=-Inf and throw an error saying this is not allow. The modified policy iteration method requires only the minor modification mentioned previously, checking that span(v-u) is finite before doing Step 4; given that you already calculate span(v-u) this will cost you essentially nothing in terms of run time. In summary, I would argue that is both very useful in many applications to let r() take -Inf values, and is easy enough to implement in the codes.
a. solving a stochastic growth model where util(consumption=0)=-Inf, and allowing for the state-action of capital =0. (if capital equals zero, then output is zero, and so consumption must be zero, giving a return on -Inf). In some sense this is just a 'mistake' by the user who should rule out capital =zero. But I suspect it is a 'mistake' users would often make.
b. Implementing a stochastic growth model with putty-clay investment, so cannot choose (action of) capital levels (k') less than current capital level (k) state minus depreciation (delta). If you allow r() to take -Inf values this is easy, just make r()=-Inf for all actions which would imply k'<(1-delta)k.
c. r()=-Inf are very useful for implementing constraints. In many economic applications you will have r() being a combination of u() and some (eg. budget) constraints. If r()=-Inf is allowed, then you can just make any breaching of the constraints a -Inf value and the rest of the MDP will be just as normal.

3. Do you have a copy of the proof handy? I would be interested in seeing it.

robertdkirkby:
On 2 again.

Realized you might have the contraints issue dealt with by more complex Q's, rather than r()=-Inf values. I need to think about how this setup with Q works a bit more closely. 

I suspect that the answer is that it is possible, but more complex in practice, to implement the budget constraints by a combination of using r() without -Inf values, and instead having zeros in r() and using Q to give zero probabilities to any such choices. This does seem like it would be harder for the user to set up though.

You probably already know the answer to this though.

oyamad:
Thanks @robertdkirkby

2.
The "product formulation" assumes that constraints are expressed by setting `r(s, a) = -inf` for infeasible `(s, a)`'s. This works with no problem as long as for all states `s`, `r(s, a) > -inf` for some action `a`. I think this accommodates your point c.

I am afraid it would be problematic if we have some `s` such that `r(s, a) = -inf` for all `a`. Mathematically, even if the optimal value function is well defined, the optimal policy is not well defined. If the values are all `-inf`, NumPy returns `-inf` as `max` and the first index that attains `max`, namely index `0` in this case `0`, as `argmax()`. In the case `u(0) = -inf` and the current capital is `0`, zero consumption yields `-inf` by assumption, whereas positive consumption also yields `-inf` for another reason, namely infeasibility. Note that assigning indices to actions can be arbitrary. If index `0` is assigned to some positive consumption and `1` is assigned to zero consumption, then `argmax()` would still be `0`. So I think the user should be advised to assign reward `-inf` only to infeasible actions (and not to include states under which all the actions are infeasible).

3.
This is not very difficult. It is similar to the proof of convergence of policy iteration.
Let `v^0`, `v^1`, `v^2`, ... and `sigma1`, `sigma2`, ... be the sequences of values and policies obtained by modified policy iteration.
If `v^0 <= T v^0`, then we have `T v^0 <= v^1 <= T v^1` (and so on).
Let me type the proof here:

```
v^0 <= T v^0 (by assumption)
    = T_sigma1 v^0 (by the definition of sigma1)
    <= T_sigma1^2 v^0 (by the monotonicity of T_sigma1)
    ...
    <= T_sigma1^k v^0
    <= T_sigma1^k+1 ^v0
    = v^1 (by the definition of v^1);
```

from the inequality `T_sigma1^k v^0 <= v^1`, we have

```
T_sigma1^k+1 v^0 <= T_sigma1 v^1 (by the monotonicity of T_sigma1)
                 <= T v^1 (since T_sigma v <= T v for any sigma and v).
```

By this, we have `T^i v^0 <= v^i` for all `i` (by the monotonicity of `T`). It is clear that `v^i <= v^*`. So, if `v^0 <= T v^0`, then the sequence converges monotonically to `v^*` from below.

robertdkirkby:
2. It does cover my c. I missed the 'all' in 'for some s, all a' in your earlier reply. I now agree with your earlier statement that it might be worth checking for as I suspect many beginner users will make the kind of mistake described by the u(c=0)=-inf and including capital=0 example.

3. Thanks.

oyamad:
The description of the algorithms in the [notebook](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/mdp/mdp_theory.ipynb#Solution-methods) was misleading because of `iter_max`; with it, finite termination is trivial (the problem was that I tried to described the algorithms and implementation at the same time). I removed the termination condition with `iter_max` from the description.

oyamad:
I did rebase and forced push.

oyamad:
As another demo, I used `MDP` to solve the [career-job model](http://quant-econ.net/py/career.html):

* [Modeling career choice](http://nbviewer.ipython.org/github/oyamad/mdp/blob/master/mdp_ex_career.ipynb)

A remark:
In using [`compute_fixed_point`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/compute_fp.py#L36) in the [Solutions](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.py/blob/master/solutions/career_solutions.ipynb), the default `max_iter=50` is used. This is binding, and makes a significant difference in particular in the case of `beta = 0.99`. In fact, the median first passage time in this case seems to be about `14` with a large enough number of iterations (or with policy iteration), whereas the Solutions notebook claims that it is about `11`.

oyamad:
Following the suggestion from @robertdkirkby , I added a method that checks that every state has at lease one action with finite reward. It also checks, in the case of state-action pairs, that for every state, there is at least one available action.

oyamad:
Following a discussion with @jstac, I renamed `MDP` to `DiscreteDP`. I hope I have made all the relevant changes properly.

jstac:
Everything looks great.  Thanks again!  I'll merge now and start writing the lecture.

PR #177, opened 2015-09-07
DiscreteDP refactoring

oyamad:
The current code sorts `s_indices` and `a_indices`, whether or not they are already sorted. This PR checks whether they are already sorted (and if they are, does not sort them again).

It also contains two other small corrections (too small for independent PRs).


jstac:
@oyamad Thanks, merging.

PR #184, opened 2015-09-12
probvec: Use less memory

oyamad:
Modify `probvec` to make memory consumption smaller.

* Remove a temporary array.
* Replace `np.diff`, which was a big bottleneck, with a custom routine `_diff`, which uses Numba.


spencerlyon2:
Looks great

jstac:
@oyamad Thanks, this is neat.  We are seeing some limitations of vectorized code.

PR #188, opened 2015-09-16
Fix optgrowth solution notebook

oyamad:
Due to the change #41 in the behavior of [`compute_fixed_point`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/compute_fp.py#L36), the solution notebook for Exercise 1 in the lecture on [Infinite Horizon Dynamic Programming](http://quant-econ.net/py/dp_intro.html) has to be revised.

* In `compute_fixed_point(gm.bellman_operator, w, max_iter=n)`, `w` is now modified *in-place*, so the initial condition `w = 5 * gm.u(gm.grid) - 25` has to be set inside the for loop.

* I suspect the [figures](http://quant-econ.net/py/dp_intro.html#og-opf) in the lecture have to be fixed as well.

* In the notebook I executed, the Figure for Exercise 2 looks different from the one in the previous notebook or in the [lecture](http://quant-econ.net/py/dp_intro.html#exercise-2).

* [optgrowth_solutions.ipynb](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.py/blob/3231cf38881e7b7a0aae196bc5622408d5cec839/solutions/optgrowth_solutions.ipynb)


mmcky:
@oyamad Thanks for this. I see - because ``w`` is updated in place it now needs to be reset in each iteration. Will merge and update the figures in the lectures.

Issue #190, opened 2015-09-17
Arellano_solutions migration to matplotlib 1.4.3 in python 3.4


finmod:
The following error appears in arellano_solutions in the plotting the last three time series.

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
C:\Users\Denis\Anaconda3\lib\site-packages\quantecon\markov\core.py in simulate(self, ts_length, init, num_reps, random_state)
    351         try:
--> 352             k = len(init)  # init is an array
    353             dim = 2

TypeError: object of type 'numpy.int64' has no len()

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-7-45523b0b0cfd> in <module>()
      1 T = 250
----> 2 y_vec, B_vec, q_vec, default_vec = ae.simulate(T)
      3 
      4 # Pick up default start and end dates
      5 start_end_pairs = []

C:\Users\Denis\Anaconda3\lib\site-packages\quantecon\models\arellano_vfi.py in simulate(self, T, y_init, B_init)
    164         mc = qe.markov.MarkovChain(self.Py)
    165 
--> 166         y_sim_indices = mc.simulate(T, init=y_init)
    167         B_sim_indices = np.empty(T, dtype=np.int64)
    168         B_sim_indices[0] = B_init

C:\Users\Denis\Anaconda3\lib\site-packages\quantecon\markov\core.py in simulate(self, ts_length, init, num_reps, random_state)
    367             else:
    368                 raise ValueError(
--> 369                     'init must be int, array_like of ints, or None'
    370                 )
    371 

ValueError: init must be int, array_like of ints, or None

oyamad:
@finmod Thank you for the report. This seems to be a bug in `MarkovChain`.

[This line](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/core.py#L365) should have been `elif isinstance(init, (numbers.Integral, np.integer)):` as in [here](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/util/random.py#L27), since "numpy.int64 is not instance of int" (numpy/numpy#2951) (I didn't know this).

To double check, I have to set up a Python 3 environment, and I will then fix this tomorrow.

oyamad:
Maybe `elif isinstance(init, numbers.Integral):` is enough (numpy/numpy#4547).

finmod:
None of the two patches.

 

[...]

PR #193, opened 2015-09-18
Fix: MarkovChain.simulate

oyamad:
* Address #190: Check that `init` is an instance of `numbers.Integral`
* Check that `init` is in `{0, ..., n-1}` or `{-n, ..., -1}`


mmcky:
@oyamad Thanks for this. Notebooks are now running in my ``py34`` env

Additional notebooks have started to be written based on this excellent addition to the library. FYI - If you ever want to test additions to ``markov`` over **all** notebooks in the repository - this can now be done using ``python scripts/test-solutions.py`` and that will provide you with a small report.

Issue #170, opened 2015-08-17
readthedocs - missing a few autojit functions


mmcky:
@cc7768 @spencerlyon2 Do either of you have an idea of why this utilities page only has one of two functions contained in the module. When I compile docs locally it has both functions - but they do not appear on readthedocs!

Summary of observations
----------------------------------
  1. Seem to be related to ``numba`` functions. 
  1. ``random.utilities`` is missing ``sample_without_replacement``
  1. ``lss.py`` is missing a function

Generated ``rst`` files look to be correct. Might be a ``readthedocs`` environment issue.

spencerlyon2:
I'm sorry I don't know what is wrong here... Strange that you get it to work locally but rtd fails

oyamad:
There were issues on this, such as numba/numba#134 and numba/numba#470, but it seems to have been fixed by numba/numba#508. I don't see why it is not working for us...

mmcky:
**Update:** As far as I can tell - this might be an issue due to the use of a Mock ``numba`` module perhaps. Maybe it's preventing proper parsing of the docstrings attached to jitted functions. Replacing the ``rtd-pip-requirements.txt`` file with a conda enironment hasn't solved the issue. Will look into this further as it will become a big issue for us.

mmcky:
We have now removed mock from ``rtd`` compilation and making use of the ``conda`` environment with a ``rtd`` yaml config file and coda environment definitions. **Note:** the ``rtd`` support for conda is currently in ``beta`` so we should keep an eye on the documentation site for compilation issues.

Issue #155, opened 2015-05-29
NumbaWarning with Python 3


oyamad:
`NumbaWarning` occurs with Python 3; for example see Travis CI result [314.2](https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/61942188) for `master`. No such warning is raised with Python 2; see [314.1](https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/61942187).

I will try to see what is happening once I install Python 3 on my computer.

oyamad:
[Numba Warnings have disappeared](https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/92814486) upon  updating to Python 3.5 by PR #210.
This is no longer an issue, so closing.

PR #195, opened 2015-09-19
Fix NumbaWarning

oyamad:
This tries to fix one of the Numba warnings shown e.g. in Travis CI [421.2](https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/81067437), on `sample_without_replacement` (although I haven't been able to reproduce the warning messages in my environment); see also #155.

The warning says Arg 3 (`random_state`) of `sample_without_replacement` is not legal in nopython mode. So I took the loops out as an independent function in nopython mode.

As a side effect, this should resolve item 2 in #170.

### DIFF
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -122,9 +122,36 @@ def sample_without_replacement(n, k, num_trials=None, random_state=None):
     random_state = check_random_state(random_state)
     r = random_state.random_sample(size=(m, k))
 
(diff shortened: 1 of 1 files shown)


oyamad:
Let me withdraw this PR.

I don't see why these warnings show up, and why only with Python 3, while I cannot reproduce them in my environment so far. I also don't know if we really need to fix them so as to the warnings to disappear.

(I will issue another PR for the second commit.)

PR #196, opened 2015-09-19
sample_without_replacement refactored

oyamad:
Use less memory


jstac:
@oyamad Great.  Thanks.

PR #198, opened 2015-09-25
DiscreteDDP: minor fix in docstring and pep8 compliance

oyamad:
Sorry to make such a small fix.

* Looking at the [documentation](http://quanteconpy.readthedocs.org/en/latest/markov/ddp.html) of `DiscreteDP`, I realized that I had forgotten to add `r` at the beginning of the docstring that contains math expressions... An `r` is added now, and some minor edits are made in docstrings.

* Wrong indentation is fixed.

### DIFF
--- a/quantecon/markov/ddp.py
+++ b/quantecon/markov/ddp.py
@@ -708,12 +708,12 @@ class DiscreteDP(object):
         sigma = self.compute_greedy(v)
 
         res = DPSolveResult(v=v,
-                             sigma=sigma,
-                             num_iter=num_iter,
-                             mc=self.controlled_mc(sigma),
-                             method='value iteration',
-                             epsilon=epsilon,
-                             max_iter=max_iter)
+                            sigma=sigma,
+                            num_iter=num_iter,
+                            mc=self.controlled_mc(sigma),
+                            method='value iteration',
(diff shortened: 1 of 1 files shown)


jstac:
@oyamad Thanks.

oyamad:
Thanks @jstac.

But there are a few mistakes left. (Sorry, I should have built the document locally; now I have found how to do [here](https://github.com/QuantEcon/QuantEcon.py/tree/master/docs).)

* The math expressions intended to be displayed are not displayed: `..math::` must to be `.. math::` (with a space).

I will have to make another PR...

There is one minor issue: if you look at the source code of [ddp.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/ddp.py), every thing on line 28 and below is in light blue. This seems to be because the `[` on line 26 is not closed and the strings are made "raw" (surrounded by `r"""` and `"""`). I feel uncomfortable with this and would like to fix it. There are at least the following two resolutions.

* Avoid `\beta \in [0, 1)`, and write `0 \leq \beta < 1` instead.
* Remove `r` before `"""`, and double all the backslashes.

Anybody have any other idea?

jstac:
@oyamad The first solution seems like the best one.

PR #199, opened 2015-09-25
Fix P update in stationary coefficients.

thomassargent30:
In the stationary coefficients for Kalman filter, update was being done with

```
P = Matrix
...
dot(P, P)
```

but this gives us P, P^2, P^4... instead of P, P^2, P^3...

Opening PR so John can make it prettier if he would like.

### DIFF
--- a/quantecon/kalman.py
+++ b/quantecon/kalman.py
@@ -192,17 +192,21 @@ class Kalman(object):
         if K_infinity is None:
             S, K_infinity = self.stationary_values()
         # == compute and return coefficients == #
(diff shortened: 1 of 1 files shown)


thomassargent30:
Leave open: Will add a corresponding test soon.

jstac:
@cc7768 I'm accepting the PR so that the correction can be merged immediately but I'll add an issue stating that we need a test.  Thanks for finding the error and handling this.

cc7768:
Sounds good.

PR #206, opened 2015-11-05
Relocating models code to examples directory

mmcky:
This PR moves the ``quantecon/models`` code to the ``examples/`` directory and updates the relevant example files that formerly imported from ``quantecon/models``. 

Issues
--------
  1. Any suggestions on how best to update the ``solutions/`` notebooks given we can't use relative imports in a non-package context?

A: One possibility is to re-arrange both ``examples/`` and ``solutions/`` to be in topic folders. 
For example ``examples/arellano/`` may contain
  - ``arellano_vfi.py``
  - ``arellano_solutions.ipynb``

and ``examples/lakemodel/`` would contain
  - ``lake.py``
  - ``lakemodel_example.py``
  - ``lakemodel_solutions.ipynb``

B: An alternative would be to start a  new repository ``quantecon-examples`` that would then contain models that could be easily imported. This would solve the import issue for ``solutions``.


mmcky:
Based on recent discussions - to preserve ``python`` import structures and to improve the mapping between examples and the lecture content - this content will be migrated to a separate repo that will contain a collection of examples compiled by chapter. 

Any examples that pertain specifically to ``quantecon.org`` will remain in ``examples``

mmcky:
**CLOSING**

This PR is not going to be merged as ``models`` is no longer going to be merged to ``examples`` but rather migrated to a separate repository with ``QuantEcon`` organisation that supports quant-econ.net with examples and solutions notebooks, in addition to ``teaching`` and demonstration material. This new repository is called ``quantecon.applications``

PR #213, opened 2015-11-26
fixed .size error

sbarratt:
.size does not exist for tuples. Replaced with len

### DIFF
--- a/quantecon/discrete_rv.py
+++ b/quantecon/discrete_rv.py
@@ -35,7 +35,7 @@ class DiscreteRV(object):
         self.Q = cumsum(q)
 
     def __repr__(self):
-        return "DiscreteRV with {n} elements".format(n=self._q.size)
+        return "DiscreteRV with {n} elements".format(n=len(self._q))
 
     def __str__(self):
         return self.__repr__()


cc7768:
Thanks for submitting this! Great for people to see these types of things and leave some feedback. 

If I'm not mistaken (which is possible), `q` should be a numpy array. Maybe instead of changing from `.size` to `len` we should make sure the input is a numpy array (or change it to one). @jstac @mmcky @spencerlyon2 @oyamad any thoughts?

sbarratt:
You could do:

```python
self._q = np.array(q)
```

oyamad:
Or,

```python
self._q = np.asarray(q)
```

which does not make a copy when `q` is already a numpy array.

Otherwise, if one wants to keep `.size` (and `self._q` as is), could do `n=self.Q.size`, where `self.Q` is a numpy array.

jstac:
@oyamad 's first suggestion sounds good.  Just waiting for some major reorganizations of the repo and lectures to finish and then we can make this change.  @sbarratt Many thanks.

jstac:
@sbarratt Thanks again for flagging this.  I'm closing the PR but we'll make the fix suggested above.

PR #219, opened 2016-01-12
Update docs to remove models subpackage and update the qe_api.py scri…

mmcky:
…pt to python3


mmcky:
@jstac once this CI turns green - we can accept this PR. 

New documentation: http://quanteconpy.readthedocs.org/en/update-docs/index.html

PR #222, opened 2016-01-14
Change converter functions to use default float

cc7768:
I don't know if there was an original reason for using `"float32"` in our converter functions, but it might be nice to use the system default. Opening this to see what people think -- I'd rather use system default.

```
In [7]: x = np.random.randn(25)

In [8]: np.asarray(x, dtype="float32").dtype
Out[8]: dtype('float32')

In [9]: np.asarray(x, dtype="float64").dtype
Out[9]: dtype('float64')

In [10]: np.asarray(x, dtype="float").dtype
Out[10]: dtype('float64')
```


spencerlyon2:
Big :+1:

jstac:
Original reason was that I didn't know any better.  Thanks.

PR #224, opened 2016-01-17
Add `distributions` to quantecon/__init__.py

oyamad:
Fixes QuantEcon/QuantEcon.notebooks#21.

(`models/career.py` used to have `from quantecon.distributions import BetaBinomial` which made `quantecon.distributions` importable without `from . import distributions` in `__init__.py`, but `models` has now gone.)


mmcky:
Thanks @oyamad.

PR #230, opened 2016-02-24
Adjustments to cleanup travis CI

mmcky:
1. Use a full anaconda environment to have the same environment as users. This might slow down the setup phase - and if so we can restore to a minimal environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
+# Travis Configuration for QuantEcon.py
+#
 # Thank you for the writeup given by dan-blanchard at https://gist.github.com/dan-blanchard/7045057
+#
+
+sudo: false        #Use new Container Infrastructure
 language: python
 python:
   - 3.5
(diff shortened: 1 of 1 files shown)


mmcky:
Trying to remove any dependency on ``sudo`` so that we can leverage the faster containerized travis builds. 

@spencerlyon2 or @cc7768: Do you recall what these lines were for in the setup phase of travis?

```bash
  - sudo rm -rf /dev/shm
  - sudo ln -s /run/shm /dev/shm
``

spencerlyon2:
Nope,  those don't look familiar to me.

// Spencer

mmcky:
Looks like a way to speed up things by holding objects in RAM. I am going to remove them from our test configuration as they require ``sudo`` to execute and we should migrate to the new container service. The easy way to do that is specify ``sudo: false``. 

http://askubuntu.com/questions/169495/what-are-run-lock-and-run-shm-used-for

Issue #229, opened 2016-02-18
filterwarnings


oyamad:
It seems that once `quantecon` is imported, the warnings the user issues by `warnings.warn` are ignored, probably because of `warnings.filterwarnings('ignore')` in [arma.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/arma.py#L16) (and there is also `warnings.simplefilter("ignore", RuntimeWarning)` in [matrix_eqn.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/matrix_eqn.py#L22)).

If it is indeed the case, this is problematic. Is there any way to put `warnings.simplefilter` in a "local scope"?

oyamad:
```python
>>> import warnings
>>> warnings.warn('warning')
__main__:1: UserWarning: warning
>>> import quantecon
>>> warnings.warn('warning')
>>>
```

spencerlyon2:
We should definitely not be silencing all warnings. I haven't looked, but do we know why we are even expecting to have warnings in arma.py?

mmcky:
I propose we remove the ``warnings.filterwarnings('ignore')`` and then deal with any warnings that arise.

mmcky:
Comment in ``matrix_eqn.py``

```python
# == Suppress warnings from checking conditioning of matrices == #
```

mmcky:
I propose we remove the ``warnings.filterwarnings('ignore')`` and then deal with any warnings that arise.

mmcky:
Closing this issue as discussion has moved to PR #231

PR #231, opened 2016-02-25
Remove warnings supression as noted by Issue #229

mmcky:
This PR removes the warning supression from `arma.py` and `matrix_eqn.py`.

**TODO:** The next step is to fix the underlying cause of the warning.

### DIFF
--- a/quantecon/arma.py
+++ b/quantecon/arma.py
@@ -4,16 +4,13 @@ Authors: Doc-Jin Jang, Jerry Choi, Thomas Sargent, John Stachurski
 
 Provides functions for working with and visualizing scalar ARMA processes.
 
+TODO: 1. Fix warnings concerning casting complex variables back to floats
+
 """
 import numpy as np
 from numpy import conj, pi
 import matplotlib.pyplot as plt
 from scipy.signal import dimpulse, freqz, dlsim
-
-# == Ignore unnecessary warnings concerning casting complex variables back to
-# floats == #
-import warnings
-warnings.filterwarnings('ignore')
 
 
 class ARMA(object):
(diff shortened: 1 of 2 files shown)


mmcky:
The ``arma`` warnings are a result of:

```python
w, spect = self.spectral_density(two_pi=False)
```

the ``spect`` array is complex. With the example on the ``quant-econ.net`` page the coeeficient on the ``j`` term are all zero. In general is it safe to use the real component of this array? If so we can add a.real on spec to send matplotlib an array of reals.

mmcky:
Moving related comments here from Issue #229 

Comment in ``arma.py``

```python
# == Ignore unnecessary warnings concerning casting complex variables back to
# floats == #
```
@jstac In general is it safe to use the real component of this array?

mmcky:
Comment in ``matrix_eqn.py``

```python
# == Suppress warnings from checking conditioning of matrices == #
```

mmcky:
@jstac would you mind reviewing this spectral density question when you have time? Its in relation to some warnings we are trying to remove in the quantecon.py package.

oyamad:
What is the status of this PR?

oyamad:
Is it possible to merge this PR as is and open another issue for discussion on arma.py and matrix_eqn.py? I find it a bit annoying that warnings are all ignored in the whole session once `quantecon` is imported...

mmcky:
Sure. I will have a quick review of this this afternoon to double check that all of the suppression statements have been removed and then I will setup a separate issue to fix ``arma.py`` and ``matrix_eqn.py``.

oyamad:
Thanks!

mmcky:
This looks ready to merge (passes tests).

Note: this will cause warnings to appear when using the library. 

https://github.com/QuantEcon/QuantEcon.py/issues/271
https://github.com/QuantEcon/QuantEcon.py/issues/272

PR #236, opened 2016-03-14
Drop `num_actions` from DiscreteDP

oyamad:
This is to remove `num_actions` from `DiscreteDP`, following my argument in https://github.com/QuantEcon/QuantEcon.jl/pull/94#issuecomment-175002454.

And also remove a small inefficiency in `modified_policy_iteration`.


mmcky:
@oyamad Thanks for this. I will plan to leave this open for comment another day or two and then I will merge into master.

PR #246, opened 2016-03-21
Remove optional numba installation code

mmcky:
This PR removes code that was written to support environments that did't have ``numba`` installed. Now that ``numba`` is a strict dependency of the package - this code can now be removed. This will simplify how code is written etc. 

**Updates:**
  1. ``markov/gth_solve`` should we remove the following given support for Python 3.5+:

```python
+#-TODO: Remove as we no longer support Python <3.5?
+try:
+    xrange
+except:  # python3
+    xrange = range
```

  2. ``gth_solve`` so as not to change the code to much, I have added a ``use_jit`` header to the function signature to allow one to optionally use numpy. @oyamad Is this ok?

**Deprecations:**
  1. The ``numba_installed`` package wide variable is **no** longer available.


oyamad:
@mmcky Yes, please remove the `try`-`except` block, and `use_jit` is fine with me.

mmcky:
@oyamad Thanks - I removed the ``try``-``except`` block and updated ``xrange`` to ``range`` in the functions.

mmcky:
Looking at the diff this ``re-base`` doesn't look right. I will be closing this PR and implementing the simpler set of changes that remove the code that supported ``optional`` numba.

PR #251, opened 2016-04-14
Remove code that supported optional numba installation

mmcky:
This PR removes code that was written to support environments that did't have ``numba`` installed. Now that ``numba`` is a strict dependency of the package - this code can now be removed. This will simplify how code is written etc. 

**Updates:**
  1. ``markov/gth_solve`` should we remove the following given support for Python 3.5+:

```python
+#-TODO: Remove as we no longer support Python <3.5?
+try:
+    xrange
+except:  # python3
+    xrange = range
```

  2. ``gth_solve`` so as not to change the code to much, I have added a ``use_jit`` header to the function signature to allow one to optionally use numpy. @oyamad Is this ok?

**Deprecations:**
  1. The ``numba_installed`` package wide variable is **no** longer available.


mmcky:
In #246 It was resolved to remove the python2.7 range statements.

mmcky:
I will leave this open for a day - in case there are any comments on the final diff.

quantecon/lss.py
```diff
@@ -9,10 +9,10 @@
 import numpy as np
 from numpy.random import multivariate_normal
 from scipy.linalg import solve
+from numba import jit
 
-#-Check if Numba is Available-#
-from .util import numba_installed, jit
-
+#TODO: Should this be specified with (nopython=True?)
```
oyamad: `@jit(nopython=True)` will work if the line `A = np.asarray(A)` is removed (assuming that the input `A` is an numpy ndarray). (But I suspect we wouldn't get much speedup.)

quantecon/random/utilities.py
```diff
@@ -75,6 +77,8 @@ def _diff(r, out):
         out[i, n] = 1 - r[i, n-1]
 
 
+#TODO: Should this use (nopython=True) flag?
```
oyamad: `nopython=True` will not work, due to `random_state` at least.

quantecon/markov/random.py
```diff
@@ -218,7 +219,8 @@ def random_discrete_dp(num_states, num_actions, beta=None,
     ddp = DiscreteDP(R, Q, beta, s_indices, a_indices)
     return ddp
 
-
+#TODO: Should this use (nopython=True) flag?
```
oyamad: `nopython=True` should work if `dtype=int` is replaced with `dtype=np.int64`. (Again I suspect we won't have much speed up.)

oyamad: In a (near) future PR, I would like to make this function `_sa_indices` public. One possibility is to move it to a new file `markov/utilities.py`. Another is to put it to `gridtools.py`, with an appropriate name?

quantecon/lss.py
mmcky: Thanks @oyamad Agree. Let's leave this as is.

quantecon/markov/random.py
mmcky: Thanks @oyamad.  Let's accept this PR and then we can update this function.

mmcky:
Thanks for your comments @oyamad. I will now remove the ``#TODO`` statements.

PR #252, opened 2016-04-15
MarkovChain: Bug fix in _compute_stationary with state_values

oyamad:
I forgot to change `recurrent_classes` to `recurrent_classes_indices`.


jstac:
@oyamad Many thanks.

PR #253, opened 2016-04-17
probvec: Use guvectorize with target='parallel'

oyamad:
This is to modify `probvec` to use [`guvectorize`](http://numba.pydata.org/numba-doc/dev/user/vectorize.html#the-guvectorize-decorator) with `target='parallel'`.
(`probvec` is mainly used to generate samples of `MarkovChain` or `DiscreteDP`.)

* Use of `guvectorize` simplifies the logic slightly; we don't need to explicitly write the outer loop.
* This version puts `sort()` in a jitted loop with nopython mode, the bottleneck of [the current `probvec`](https://github.com/QuantEcon/QuantEcon.py/blob/05d72ce1cb6a24317838909b019c9f711ccad61f/quantecon/random/utilities.py#L50).
* [Sorting by Numba may be slower than Numpy](http://numba.pydata.org/numba-doc/0.25.0/reference/numpysupported.html#other-methods), but with `target='parallel'` this version is faster than the current version on `master` for cases with large size, whereas it seems slower with small size (probably due to overhead).

Timing on a machine with 4 cores:

`master`:

```python
In [1]: import quantecon as qe

In [2]: seed = 1234

In [3]: m, k = 5000, 5000

In [4]: %timeit qe.random.probvec(m, k, random_state=seed)
1 loop, best of 3: 1.63 s per loop

In [5]: beta = 0.95

In [6]: num_states, num_actions = 500, 100

In [7]: %timeit qe.markov.random_discrete_dp(num_states, num_actions, beta, random_state=seed)
1 loop, best of 3: 1.45 s per loop
```

This version:

```python
In [1]: import quantecon as qe

In [2]: seed = 1234

In [3]: m, k = 5000, 5000

In [4]: %timeit qe.random.probvec(m, k, random_state=seed)
1 loop, best of 3: 681 ms per loop

In [5]: beta = 0.95

In [6]: num_states, num_actions = 500, 100

In [7]: %timeit qe.markov.random_discrete_dp(num_states, num_actions, beta, random_state=seed)
1 loop, best of 3: 747 ms per loop
```

This is the first case in quantecon where `guvectorize` is used; we might discuss its pros and cons here.

### DIFF
--- a/quantecon/random/tests/test_utilities.py
+++ b/quantecon/random/tests/test_utilities.py
@@ -3,13 +3,33 @@ Tests for util/random.py
 
 Functions
 ---------
+probvec
 sample_without_replacement
 
 """
 import numpy as np
 from numpy.testing import assert_array_equal, assert_raises
 from nose.tools import eq_
-from quantecon.random import sample_without_replacement
+from quantecon.random import probvec, sample_without_replacement
+
+
+# probvec #
+
+class TestProbvec:
+    def setUp(self):
+        self.m, self.k = 2, 3  # m vectors of dimension k
+        seed = 1234
+
+        self.out_parallel = probvec(self.m, self.k, random_state=seed)
+        self.out_cpu = \
+            probvec(self.m, self.k, random_state=seed, parallel=False)
+
+    def test_shape(self):
+        for out in [self.out_parallel, self.out_cpu]:
+            eq_(out.shape, (self.m, self.k))
+
+    def test_parallel_cpu(self):
+        assert_array_equal(self.out_parallel, self.out_cpu)
 
 
 # sample_without_replacement #
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -10,7 +10,7 @@ from ..util import check_random_state
 
 # Generating Arrays and Vectors #
 
-def probvec(m, k, random_state=None):
+def probvec(m, k, random_state=None, parallel=True):
     """
     Return m randomly sampled probability vectors of dimension k.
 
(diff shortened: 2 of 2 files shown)


spencerlyon2:
I'm not opposed to using guvectorize at any point in the library, but I don't love this particular change. The reason is that if I was using these tools within a larger algorithm, I would almost certainly want to parallelize a different section of my code. Having this routine always run in parallel might make that difficult.

oyamad:
@spencerlyon2 We shouldn't accept this PR if it can interfere anything more important; `probvec` is just used mainly (or only) for benchmark testing.

But can you elaborate on your point? Do you have in mind any case where you may use this particular routine within a larger algorithm?

mmcky:
Hi @oyamad and @spencerlyon2. Where are we up to with this PR?

spencerlyon2:
I thought about it more. I vote that we accept this. 

It would be great it numba allowed us to use `guvectorize` and then select the execution target at runtime instead of compile time, but I don't think that is possible.

oyamad:
We could possibly do:

```py
def probvec(m, k, random_state=None, parallel=True):
    ...
    if parallel:
        _probvec_parallel(r, x)
    else:
        _probvec_cpu(r, x)


def _probvec(r, out):
    ...

_probvec_parallel =
    guvectorize(['(f8[:], f8[:])'], '(n), (k)', nopython=True, target='parallel')(_probvec)

_probvec_cpu =
    guvectorize(['(f8[:], f8[:])'], '(n), (k)', nopython=True, target='cpu')(_probvec)
```

mmcky:
@spencerlyon2 Are you still concerned with: 

    Having this routine always run in parallel might make that difficult.

We can implement options through a wrapper (as per @oyamad solution) or make ``parallel`` the default.

spencerlyon2:
I think the proposed solution sounds great. Let's go with that!

mmcky:
@oyamad please let me know if you are happy with this update then we can merge this into the master branch. Thanks.

oyamad:
@mmcky Thanks, please do merge. (I added some docstring and tests.)

mmcky:
Thanks @oyamad! Merging now.

PR #265, opened 2016-10-20
Player, NormalFormGame: Make `payoff_array`'s C contiguous

oyamad:
Make `payoff_array`'s C contiguous when when a `payoff_profile_array` is passed to `NormalFormGame`, in order for Numba not to complain.

### DIFF
--- a/quantecon/game_theory/normal_form_game.py
+++ b/quantecon/game_theory/normal_form_game.py
@@ -475,11 +475,16 @@ class NormalFormGame(object):
                         'size of innermost array must be equal to ' +
                         'the number of players'
                     )
-                self.players = tuple(
-                    Player(
+                payoff_arrays = tuple(
+                    np.empty(data.shape[i:-1]+data.shape[:i], dtype=data.dtype)
+                    for i in range(N)
+                )
+                for i, payoff_array in enumerate(payoff_arrays):
+                    payoff_array[:] = \
                         data.take(i, axis=-1).transpose(list(range(i, N)) +
                                                         list(range(i)))
-                    ) for i in range(N)
+                self.players = tuple(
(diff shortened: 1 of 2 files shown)


oyamad:
The errors occur with Numba 0.29.0, but not with 0.28.1 or 0.26.0.

spencerlyon2:
Sorry, I didn't quite follow. 

Does this change-set  fix the errors on all versions of numba, or do we still have issues on 0.29?

oyamad:
Sorry for not being clear.

* The errors from Travis CI are not related to this PR. They are from random/utilities.py and lss.py. Those errors only occur with the latest Numba.
* What this PR does is to make all the `NormalFormGame.players[i].payoff_array`'s C contiguous, where there is no point of having a non-contiguous numpy array, and there is indeed a case (in some code I am currently writing) that passing a tuple of a C-contiguous array and a F-contiguous array does a bad thing making Numba complain (which has nothing to do with the errors here).

oyamad:
For the error from random/utilities.py, I filed an issue numba/numba#2159. I haven't investigated for the one from lss.py.

oyamad:
Opened numba/numba#2169 for the error from [lss.py](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/lss.py).

In fact, this error can be avoided by removing `A = np.asarray(A)` assuming that the input `A` is already an `np.array`, which allows us to have `nopython=True` for `@jit`.

oyamad:
I changed the patch (and made a forced push), where `order='C'` is now applied when instantiating a `Player`.

oyamad:
Let me bring the previous commit back...

oyamad:
Rebased on `master`. This is ready to merge.

mmcky:
Thanks @oyamad merging this into ``master`` now.

PR #268, opened 2016-11-07
game_theory: Add lemke_howson

oyamad:
Implement the Lemke-Howson algorithm.

1. Implemented with floating point arithmetic. I also tried integer pivoting (as described in von Stengel 2007), but overflow easily occurs with `np.int64`. In a Julia version we may want to try with `BigInt`.

2. The `full_output` option mimics root finding routines in scipy such as [`brentq`](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.optimize.brentq.html).

2. `sicpy.optimize.OptimizeResult` is used again as in `DPSolveResult` (but with `sort` in `__repr__`). Maybe we should put it in `util` for example, to be inherited by `DPSolveResult` and `NashResult` here to get rid of duplicates in the code.

Cc: @kevtimova @ArnavSood @shizejin

### DIFF
--- a/quantecon/game_theory/__init__.py
+++ b/quantecon/game_theory/__init__.py
@@ -6,3 +6,4 @@ from .normal_form_game import Player, NormalFormGame
 from .normal_form_game import pure2mixed, best_response_2p
 from .random import random_game, covariance_game
 from .support_enumeration import support_enumeration, support_enumeration_gen
+from .lemke_howson import lemke_howson
--- a/quantecon/game_theory/lemke_howson.py
+++ b/quantecon/game_theory/lemke_howson.py
@@ -13,8 +13,7 @@ TOL_PIV = 1e-10
 TOL_RATIO_DIFF = 1e-15
 
 
-def lemke_howson(g, init_pivot=0, max_iter=10**6, lex_min=False,
-                 full_output=False):
+def lemke_howson(g, init_pivot=0, max_iter=10**6, full_output=False):
     """
     Find one mixed-action Nash equilibrium of a 2-player normal form
     game by the Lemke-Howson algorithm [1]_, implemented by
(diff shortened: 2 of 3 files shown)


ArnavSood:
@oyamad: Thanks for reminding me about this.

Is the Julia analog already being worked on? If not, I'd be happy to attempt it.

jstac:
@oyamad It's beautiful code, as usual.  

The failing tests seem to be spread across the library -- perhaps due to changes in Numba?

cc7768:
I can confirm it is due to changes in `numba` -- I was trying to run some of the linear state space code and had issues until I downgraded to numba 0.28.

I'll raise a separate issue so we can address this.

jstac:
@cc7768 Thanks, that would be helpful.  I saw somewhere that numba expects new style classes now.  Could that be the issue?  (In other words, classes shouldn't inherit from `object`.)

oyamad:
@ArnavSood No I don't have a Julia version. You can continue your work in Julia.

oyamad:
Rebased on `master`.

jstac:
The code is beautifully written, everything works as advertised and the documentation is good.  Thanks!

@mmcky  My view is that this can be merged.

@oyamad Might you or one of your students be willing to add a notebook to [the notebook site](http://quantecon.org/notebooks.html) giving a couple of examples, or, alternatively, update [this notebook](http://nbviewer.jupyter.org/github/QuantEcon/QuantEcon.notebooks/blob/master/game_theory_py.ipynb) accordingly?

mmcky:
Thanks @oyamad and @jstac. I will merge this PR now.

oyamad:
Sure, I will update the game theory notebook.

Issue #278, opened 2017-01-11
Problems with `fetch_nb_dependency`


cc7768:
@thomassargent30 was trying to download files from `QuantEcon.applications` and the current dependency fetcher doesn't seem to be able to do this easily.

It might be useful to do the following two things:

* Currently, the function takes a `deps` argument because it helps you get to the dependency folder of `QuantEcon.Notebooks`. In [line 73](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/util/notebooks.py#L73) of the function it joins the `repo`, `raw`, `deps`, and other arguments together. It might make sense to remove the `deps` argument altogether (or rename it to be `folder`) and just require people to use the dictionary approach when getting files from `QuantEcon.Applications`... There could be a few other possible solutions to this, but I think the current design prevents the code from being able to fetch files from other repositories (which I think was one of the goals).
* Write some examples of how this function works.

Both of these seem like relatively easy tasks? Do we have any newer members of the team that might be capable of doing this?

Maybe @mmcky has some design thoughts to fix this?

oyamad:
I agree `deps` will better be `folder`, or `directory`.

A use case is found in Cell `[44]` in [here](http://nbviewer.jupyter.org/github/QuantEcon/QuantEcon.notebooks/blob/master/ddp_ex_career_py.ipynb#Comparison-with-CareerWorkerProblem).

mmcky:
Only thing left is to: 

- Write some examples of how this function works.

mmcky:
Closing based on PR #279 and #320

PR #279, opened 2017-01-11
improve dependency fetcher with clearer parameters

mmcky:
@oyamad Once this is merged -- unfortunately this will break the notebook. But I think you guys are right -- this will be clearer to use.

@cc7768 Is this what you mean re issue #278?


cc7768:
Yes, this is roughly what I had in mind. I think the dictionary behavior for `files` was a bit confusing because the folder name could be specified in two different places (either as the key in the dictionary or as the `deps` argument).

It would be nice if we could have someone write a short example of fetching files from each of the different repositories and include it in the documentation for this function to make it easier for others to put this function to work.

mmcky:
Before I merge this I really should write a test.

PR #280, opened 2017-01-30
upgraded tic, tac and toc to top level

jstac:
I suggest we upgrade the useful utilities `tic`, `toc` and `tac` to the top level namespace.

@mmcky If you agree would you mind to test first and then merge?


mmcky:
Thanks @jstac ... I see no problem in promoting these utilities.
Will merge now.

PR #292, opened 2017-02-24
WIP: fix doc warnings

### LINKED ISSUE #287: Fix doc warnings
* [ ] Unknown sections
  - Unknown section Return
  - Unknown section Parameters:
  - Unknown section Returns:

Example: https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/cartesian.py#L16
Change `Parameters:` to `Parameters`

shizejin:
fixes #287 

Fix unknown sections.

### DIFF
--- a/quantecon/game_theory/normal_form_game.py
+++ b/quantecon/game_theory/normal_form_game.py
@@ -277,13 +277,15 @@ class Player(object):
         Parameters
         ----------
         opponents_actions : scalar(int) or array_like
-            A profile of N-1 opponents' actions. If N=2, then it must be
-            a 1-dimensional array of floats (in which case it is treated
-            as the opponent's mixed action) or a scalar of integer (in
-            which case it is treated as the opponent's pure action). If
-            N>2, then it must be an array of N-1 objects, where each
-            object must be an integer (pure action) or an array of
-            floats (mixed action).
+            A profile of N-1 opponents' actions, represented by either
+            scalar(int), array_like(float), array_like(int), or
+            array_like(array_like(float)). If N=2, then it must be a
+            scalar of integer (in which case it is treated as the
+            opponent's pure action) or a 1-dimensional array of floats
+            (in which case it is treated as the opponent's mixed
+            action). If N>2, then it must be an array of N-1 objects,
+            where each object must be an integer (pure action) or an
+            array of floats (mixed action).
 
         tie_breaking : str, optional(default='smallest')
             str in {'smallest', 'random', False}. Control how, or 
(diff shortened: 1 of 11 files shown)


quantecon/game_theory/normal_form_game.py
```diff
@@ -276,8 +276,7 @@ def best_response(self, opponents_actions, tie_breaking='smallest',
 
         Parameters
         ----------
-        opponents_actions : array_like(int or array_like(float)) or
-                            array_like(int, ndim=1) or scalar(int)
+        opponents_actions : scalar(int) or array_like
```
mmcky: Should this ``array_like`` be ``array_like(int or float)``?

mmcky:
Thanks @shizejin. This is a good PR. Just had one question above.

quantecon/game_theory/normal_form_game.py
shizejin: Thank you for comments! I am wondering whether `array_like(int or float)` could cover `array_like(array_like(float))`? If this is the case, `array_like(int or float)` would be better.

oyamad: If we ignore the maximum length 72, it would have been
```py
        opponents_actions : scalar(int) or array_like(int, float, or array_like(foat))
```

See https://github.com/QuantEcon/QuantEcon.py/issues/287#issuecomment-286448104.

mmcky: What do you think about the following format?

```
opponents_actions : scalar(int) or array_like
    array_like can be array_like(int, float, or array_like(foat))
    A profile of N-1 opponents' actions. If N=2, then it must be
    a 1-dimensional array of floats (in which case it is treated		             
    as the opponent's mixed action) or a scalar of integer (in
```

oyamad: What about this:

```
        opponents_actions : scalar(int) or array_like
            A profile of N-1 opponents' actions, represented by either
            scalar(int), array_like(float), array_like(int), or
            array_like(array_like(float)). If N=2, then it must be a
            scalar of integer (in which case it is treated as the
            opponent's pure action) or a 1-dimensional array of floats
            (in which case it is treated as the opponent's mixed
            action). If N>2, then it must be an array of N-1 objects,
            where each object must be an integer (pure action) or an
            array of floats (mixed action).
```

mmcky: @oyamad I agree - I think that is excellent.

mmcky:
Thanks @shizejin and @oyamad.

PR #293, opened 2017-03-01
Test rouwenhorst

lbui01:
Hi @sglyon  - 

I have added a test for the rouwenhorst method.  It's largely the same as the test for the tauchen test, except i've added a control case where i've pre-calculated the MC and state space values for pre-specified parameters.

Any comments would be appreciated!

I've also changed the _ _ init _ _ files in order to load the rouwenhorst from approximation.py.

@mmcky - as discussed: publish new version because rouwenhorst isn't available on API. 

@mmcky  - migrating to sl/rouwenhorst, please close without merge.


mmcky:
Thanks @lbui01. Closing this issue and deleting the branch as you have now pushed this test to ``sl/rouwenhorst`` for travis testing. The API will be updated on the next release of ``quantecon=0.3.5``.

PR #296, opened 2017-04-02
DOC: Small correction

oyamad:
Correction according to https://github.com/QuantEcon/Games.jl/pull/26#discussion_r108597600.


mmcky:
thanks @oyamad

PR #157, opened 2015-06-22
DOC: fix doxstring for compte_sequence in lqcontrol

sglyon:
Wasn't true!

PR #297, opened 2017-04-12
implement changes from PR #157 to fix lqcontrol docstring

mmcky:
This PR

- implements changes to ``lqcontrol.py`` from PR #157 (without old version changes). 

the only variation is to continued use of ``w_path`` rather than ``w_t`` used in PR #157

### DIFF
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -238,13 +238,13 @@ class LQ(object):
         Returns
         ========
         x_path : array_like(float)
-            An n x T matrix, where the t-th column represents x_t
+            An n x T+1 matrix, where the t-th column represents x_t
 
         u_path : array_like(float)
             A k x T matrix, where the t-th column represents u_t
 
         w_path : array_like(float)
-            A j x T matrix, where the t-th column represent w_t
+            A j x T+1 matrix, where the t-th column represent w_t
 
         """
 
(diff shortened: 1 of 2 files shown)


mmcky:
@sglyon. These changes were failing due to change in return object. Previously this method was returning ``Cw_path``. Should this method return ``w_path`` as per docstring or ``Cw_path``. This implements your previous fix.

mmcky:
Based on this comment

```
I think it is still meaningful. In the docstring we claimed to be returning the w_t, but we were actually returning C * w_t

Also the number of elements in the return value was incorrectly documented.
```

This is now correct. I have added in the ``dot(c, w_t)`` step in the test.

sglyon:
Hey @mmcky Sorry but I don't understand the question. Can you repeat?

mmcky:
Hey @sglyon. The comment above came from PR #157 - which was a PR you made to implement the original change. I intend to close PR #157 and merge this PR instead after review.

mmcky:
@sglyon Just to clarify the question. 

Should this method return w_path as per docstring or Cw_path. Cw_path is ``dot(C, w_path)``

The tests on the previous PR was failing because the test assumed the function was returning Cw_path and not w_path as per the docstring for the function.  

This PR implements your previous fix that never got merged.

sglyon:
Given that the user will already have `C` inside the `LSS` instance and we can't really undo the `dot`, my preference would be to return `w_path`.

Issue #302, opened 2017-04-12
TEST: Add tests for game_theory/lemke_howson.py


mmcky:
Coverage

```
game_theory/lemke_howson.py            152    122    20%
```

This has a similar issue to #300. Coverage is driven by internal private functions. As long as all internal functions are testing through ``lemke_howson()`` calls we should probably add ``# pragma: no cover`` to internal functions to fix the coverage statistics.

oyamad:
@mmcky Thanks for opening these issues on testing.

The decrease in the coverage value is in fact not because of the privateness, but because jitted functions with `nopython=True` are not "covered"; see https://github.com/QuantEcon/QuantEcon.py/pull/144#issuecomment-98728630.

But yes, fine with me to exclude private functions that are jitted with `nopython=True`.

mmcky:
Thanks @oyamad. I have pushed a branch that updates ``.coveragerc`` to skip ``@jit`` functions in the coverage statistics. This regex approach moves ``lemke_howson.py`` to 80+%

Perhaps I should fine tune from ``@jit`` to ``@jit(.*nopython=True)``

PR #303, opened 2017-04-12
adjust coveragerc to skip @jit functions

mmcky:
This PR adjusts the coverage rc file to adopt custom regex exclusions to make ``coverage`` statistics more useful in diagnosing code that isn't covered by tests. 

This will help with PR #302

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -28,3 +28,4 @@ exclude_lines =
 
     # Don't count Numba nopython=True jit functions
     @jit\(.*nopython=True
+    @njit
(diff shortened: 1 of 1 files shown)


mmcky:
Many ``@jit`` functions are used within a wrapper function that is tested. As long as each support ``jit`` function is tested through the parent function (either directly, or via a ``kwarg`` option) then I don't see a problem with this PR as it provides more accurate coverage statistics. It does however mean that we would need to ensure ``@jit`` functions are fully tested manually.

I welcome feedback on this proposal. 

A good example of this is contained in Issue #302

oyamad:
Thanks @mmcky.

The alias `@njit` should also be excluded.

mmcky:
Thanks @oyamad. good pickup.

PR #308, opened 2017-05-12
fix compatibility with pandas v0.2

natashawatkins:
- this PR removes the ols import from pandas, which has been deprecated in pandas v0.2 and is causing the quantecon package to break
- I've used the statsmodels time series module to estimate the ar1
- I was using the lecture solutions to test the code, but this includes a simulation so I can't exactly replicate the output - would someone be able to check this gives correct results?
- I'm not sure why Newey-West covariance matrix is used, but I've included it


jstac:
@natashawatkins Many thanks.

PR #310, opened 2017-05-17
remove dependency on statsmodels and use matrices to run OLS

natashawatkins:


jstac:
@natashawatkins Thanks.
@mmcky I'm going to merge this because there's conflict between the code we import from `statsmodels` and the latest version of pandas -- hope that's OK.

mmcky:
thanks @jstac and @natashawatkins

PR #311, opened 2017-05-28
FIX: support_enumeration: Use `_numba_linalg_solve`

oyamad:
* It turned out that the workaround in item 3 in https://github.com/QuantEcon/QuantEcon.py/pull/263#issue-178525924 did not work properly in all cases.
* This fix replaces "SVD + linalg.solve" with `_numba_linalg_solve`, which calls directly Numba internal `numba_xgesv`.
* As a bonus, this leads to about 9x speedup.

```py
ns = [8, 9, 10, 11]
seed = 1234
for n in ns:
    g = gt.random_game((n, n), random_state=seed)
    %timeit gt.support_enumeration(g)
```

* Current version:
  * n = 8: `99.1 ms`
  * n = 9: `391 ms`
  * n = 10: `1.58 s`
  * n = 11: `6.56 s`

* This PR:
  * n = 8: `11.6 ms`
  * n = 9: `43 ms`
  * n = 10: `179 ms`
  * n = 11: `727 ms`

```py
numba.__version__
'0.30.1'
```

### DIFF
--- a/quantecon/util/numba.py
+++ b/quantecon/util/numba.py
@@ -1,0 +1,71 @@
+"""
+Utilities to support Numba jitted functions
+
+"""
+import numpy as np
+from numba import generated_jit, types
+from numba.targets.linalg import _LAPACK
+
+
+# BLAS kinds as letters
+_blas_kinds = {
+    types.float32: 's',
+    types.float64: 'd',
+    types.complex64: 'c',
+    types.complex128: 'z',
+}
+
+
+@generated_jit(nopython=True, cache=True)
+def _numba_linalg_solve(a, b):
+    """
+    Solve the linear equation ax = b directly calling a Numba internal
+    function. The data in `a` and `b` are interpreted in Fortran order,
+    and dtype of `a` and `b` must be the same, one of {float32, float64,
+    complex64, complex128}. `a` and `b` are modified in place, and the
+    solution is stored in `b`. *No error check is made for the inputs.*
+
+    Parameters
+    ----------
+    a : ndarray(ndim=2)
+        2-dimensional ndarray of shape (n, n).
+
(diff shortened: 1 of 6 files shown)


quantecon/util/numba.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@generated_jit(nopython=True, cache=True)
+def _numba_linalg_solve(a, b):
+    """
+    Solve the linear equation ax = b directly calling a Numba internal
+    function. The data in `a` and `b` are interpreted in Fortran order,
+    and dtype of `a` and `b` must be the same, one of {float32, float64,
+    complex64, complex128}. `a` and `b` are modified in place, and the
+    solution is stored in `b`. *No error check is made for the inputs.*
```
mmcky: @oyamad is it standard convention to return the solution in ``b`` and not explicitly return the solution and a status code from the function as a tuple?

mmcky:
Thanks @oyamad. I just have one ``inline`` question/comment.

quantecon/util/numba.py
oyamad: I am not sure if I understand your question, but [here](https://github.com/QuantEcon/QuantEcon.py/blob/20da0f8872fae93d5adc4c5e3b95582879263ba3/quantecon/game_theory/support_enumeration.py#L161-L163) is where this is used, where it is check whether matrix `A` is nonsingular (instead of `try`-`except` which is not available in nopython mode), and in case `A` is singular I don't know what is in there in `b`.

mmcky: ah. I see. In the code example ``b`` is ``out`` and it is modified in place (rather than returned). Thanks @oyamad

PR #316, opened 2017-07-17
Issue 47 - Added solver for sylvester equations

### LINKED ISSUE #47: Refactor matrix equations (Ricatti, Lyapunov, Sylvester)
Currently these equations live in different modules and some of them are not easily discoverable.

For all of them there are also several possible implementations:
- in scipy (calling linpack)
- in quantecon (i.e. `doublej`)
- in [Slycot](https://github.com/repagh/Slycot), which is based on the last opensource version of [SLICOT](http://slicot.org/)

Slycot, used to be complicated to build, but this has changed, and it is now pip-installable. It is supposed to be efficient and AFAIK is the only option for Generalized Sylvester equations.
The python-control library has a `mateqn.py` file (https://github.com/python-control/python-control/blob/master/src/mateqn.py) that is also quite complete and that we could partly reuse or depend on. It seems to do extensive error handling and input checks.

My point is that we should deal with all of these matrix equations for optimal control together in a systematic way. For instance we could
- move them to a common module (like `matrix_equations.py`)
- provide wrapper functions (`solve_lyapunov`, `solve_generalized_sylvester`, `solve_riccati`, etc...) that can switch between different implementations
- investigate which implementation is the best.

TomHase:
I want to resolve #47 with the algorithm in this paper: http://www.sciencedirect.com/science/article/pii/S0096300308006693

The paper suggests to use a generalization of the GMRES algorithm to solve generalized Sylvester equations. The algorithm works as follows: At the j-th iteration we use a Global Arnoldi algorithm to construct the bases of the j-th matrix Krylov subspace associated with the equation. Then we find the matrix in the Krylov subspace that minimizes the residual of the problem. If the residual is not small enough we repeat the procedure for iteration j+1.

For large-scale problems it becomes very expensive to store the bases of the matrix Krylov-subspace so it makes sense to restart the algorithm after a fixed number of iterations. However, this can be tricky because if we restart after too few iterations the algorithm might not converge. If you are interested in this algorithm we could also try to increase the speed of convergence by including some preconditioning strategies.

### DIFF
--- a/quantecon/matrix_eqn.py
+++ b/quantecon/matrix_eqn.py
@@ -6,17 +6,16 @@ equations.  Currently has functionality to solve:
 
 * Lyapunov Equations
 * Ricatti Equations
+* Generalized Sylvester Equations
 
-TODO: 1. See issue 47 on github repository, should add support for
-      Sylvester equations
-      2. Fix warnings from checking conditioning of matrices
+TODO: 1. Fix warnings from checking conditioning of matrices
 """
 from __future__ import division
 import numpy as np
 from numpy import dot
 from numpy.linalg import solve
 from scipy.linalg import solve_discrete_lyapunov as sp_solve_discrete_lyapunov
-
+from numpy.linalg import norm
 
 def solve_discrete_lyapunov(A, B, max_it=50, method="doubling"):
     r"""
@@ -209,3 +208,106 @@ def solve_discrete_riccati(A, B, Q, R, N=None, tolerance=1e-10, max_iter=500):
             i += 1
 
     return H1 + gamma * I  # Return X
+
+
+def solve_sylvester(A, B, C, X0=[], tol=1e-10, restart=[], maxiter=10000):
+    """
+    Computes the solution (X) to the generalized sylvester equation: 
+    
+    \sum_{i=1}^{q} A_{i} X B_{i} = C 
+
+    Parameters
+    ------------------------------------
+    A: list of q (n, n) arrays 
+         Leading matrices of the generalized sylvester equation 
+
+    B: list of q (p, p) arrays
+         Trailing matrices of the generalized sylvester equation
+
+    C: (n, p) array
+         Right hand side matrix
+
+    X0: (n, p) array, optional
+         Initial guess
+
+    tol: float, optional 
+         Error tolerance
+
+    restart: int, optional
+         Restart the algorithm every restart inner iterations
+
+    maxiter: int, optional
+         Maximum number of iterations
+
+    Returns
+    ------------------------------------
+    X: (n, p) array
+         Solution to the generalized sylvester equation
+
+    References
+    ------------------------------------
+    Bouhamidi, A., & Jbilou, K. (2008). A note on the numerical approximate 
+    solutions for generalized Sylvester matrix equations with applications. 
+    Applied Mathematics and Computation, 206(2), 687-694. Chicago.
+    """
+    n = C.shape[0]
+    p = C.shape[1]
+        
+    if X0 == []:
+        X0 = np.zeros([n,p])
+
+    R0 = C - sum([np.dot(np.dot(Ai, X0), Bi) for Ai, Bi in zip(A, B)]) 
+    beta = norm(R0) 
+    epsilon = norm(R0) 
+    V = [R0 / beta] 
+    
+    j = 0
+    k = 0
+    while epsilon > tol:
+        
+        j += 1
+        k += 1
+
+        # Arnoldi algorithm 
+        hj = np.zeros(j+1)
+        Vj = sum([np.dot(np.dot(Ai, V[j-1]), Bi) for Ai, Bi in zip(A, B)])
+
+        for i in range(j):
+           hj[i] = np.trace(np.dot(V[i].T, Vj))
+           Vj = Vj - np.dot(hj[i], V[i]) 
+       
+        hj[-1] = norm(Vj) 
+        Vj = Vj / hj[-1] 
+        V.append(Vj) 
+    
+        if j == 1:
+            H = hj.reshape(hj.shape[0], -1) 
+        else:
+            H = np.vstack([H, np.zeros([1, H.shape[1]])])
+            H = np.hstack([H, hj.reshape(hj.shape[0], -1)])
+
+        # Minimize Euclidean norm of the residual
+        e1 = np.hstack([1, np.zeros(j)])
+        y = np.linalg.lstsq(H, beta*e1)[0]
+        X = np.dot(np.hstack(V[0:-1]), np.kron(y, np.eye(p)).T) + X0
+
+        # Compute new residual 
+        R = C - sum([np.dot(np.dot(Ai, X), Bi) for Ai, Bi in zip(A, B)]) 
+        epsilon = norm(R)
+        
+        if k >= maxiter:
+            print("The algorithm did not converge after " + str(maxiter) 
+                + " iterations. The residual is " + str(epsilon) + ".") 
+            break
+
+        # Restart
+        if restart != []:
+            if j == restart:
+                j = 0
+                X0 = X
+                R0 = R 
+                beta = norm(R0) 
+                V = [R0 / beta] 
+
+    return X 
+
--- a/quantecon/tests/test_sylvester.py
+++ b/quantecon/tests/test_sylvester.py
@@ -1,0 +1,55 @@
+"""
+filename: test_sylvester.py
+author: Thomas Hasenzagl
+email: thomas.hasenzagl@gmail.com
+"""
+
+import numpy as np
+from numpy.testing import assert_allclose
+from scipy import linalg
+from solve_sylvester import *
+
+def test_sylvester(A, B, C):
+    """
+    Tests if the solution from solve_sylvester is equal
+    to the solution from scipy.linalg.solve_sylvester
+    """
+    A1 = A 
+    B2 = B
+    A2 = np.eye(A1.shape[0])
+    B1 = np.eye(B2.shape[0])
+    M1 = [A1, A2]
+    M2 = [B1, B2]
+    
+    desired = linalg.solve_sylvester(A1, B2, C)
+    computed =  solve_sylvester(M1, M2 , C)
+
+    assert_allclose(computed, desired)
+    
+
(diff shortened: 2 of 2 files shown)


jstac:
@TomHase Thanks!  It looks like you've done a great job of writing up the algorithm and the tests look sensible too.

@albop This relates to an issue that you opened a while back, #47 .  Do you have any thoughts?  Are you happy for us to close the issue once this PR is accepted?

CC @cc7768 @mmcky

albop:
Sorry I didn't see earlier. I'll have a look today.

albop:
@TomHase, @jstac , @mmcky , I have started looking at the PR. 
First, one very minor cosmetic comment: to me `None` would be better than `[]` as default arguments.

More substantial:
- problem solved is `sum_i A_i X B_i = D_i` which includes `AX+BXC=D` as a special case, which is what I need for second order perturbations. It doesn't include the special structure `AX+BX[C,C,C]=D` for higher order ones. That would still be an improvement though.
- I haven't checked implementation against algorithm described in the paper, bu I did a few tests on real and random data and the answers seem correct.
- However performances seem catastrophic w.r.t. Slycot . Comparing timings with the version in Dolo (which calls Slycot):
```

**Comment by albop** [comment]
lost my editing, rewriting...

@TomHase, @jstac, @mmcky : I started to look at the PR.  I Have one very small cosmetic comment: I would use `None` instead of `[]` as default arguments.

Then, more substentially:

- Problem solved is `sum_i A_i X B_i = D` which includes as a special case the one arising from second order approximation in discrete models `A X + B X C=D`. It doesn't solve the higher order version `AX + BX.[C,...C] = D` albeit I wonder from looking at the algo whether it could be adjusted to deal with that case (without forming the product kron(C,...,C)`)
- The result seems correct both on random matrices and random ones.
- I did a speed comparison w.r.t. the implementation in Dolo, which just calls Slycot. This shows the proposed implementation is painfully slow, even for small matrices. Actually, I suspect it would also be slower than inverting the vectorized system ( B'A vec(X) = vec(D) in the paper ). Reason I'm saying is that inverting a ( 20.20 x 20.20 ) matrix is not that long... (see below)
```
def test_sylv(n=10):
    A = np.random.random((n,n))
    B = np.random.random((n,n))
    C = np.random.random((n,n))
    D = np.random.random((n,n))
    I = np.eye(n)
    X = qme.solve_sylvester([A,B],[I,C],D)
    test = A@X + B@X@C - D
    return abs(test).max()


def test_dolo_sylv(n=10):
    from dolo.numeric.matrix_equations import solve_sylvester
    A = np.random.random((n,n))
    B = np.random.random((n,n))
    C = np.random.random((n,n))
    D = np.random.random((n,n))
    I = np.eye(n)
    X = solve_sylvester(A,B,C,-D)
    test = A@X+B@X@C - D
    return abs(test).max()

%time test_sylv(n=20)
# CPU times: user 9.86 s, sys: 28 ms, total: 9.89 s
# Wall time: 4.96 s
# 8.7190255015912044e-12
%time test_dolo_sylv(n=20)
# CPU times: user 4 ms, sys: 0 ns, total: 4 ms
# Wall time: 1.14 ms
# 9.6034291630076041e-14
```

It would be interesting to know whether the speed issue comes from the implementation (which can be corrected) or whether there are specific conditions under which the algorithm will be faster to converge and whether these can be expected in economic problems. (For instance, if Krylov subspaces decrease very fast).
It seems important to know that before merging this PR. I would also strongly suggest to implement the naive vectorized version: this one would both be a crude substitute for Slycot, as well as a benchmark to assess performance.

**Comment by albop** [comment]
Just checked the status of Slycot: it is pip and conda-installable but still not available on windows

**Comment by jstac** [comment]
Wow, that is a big difference. 

One way to tell whether the algorithm is at fault is to code this same algorithm up in Julia. If there's not much speed gain then the problem is with the algorithm.  The added benefit is that we have a potential implementation in Julia too.

I can ask an RA to do this... ?

**Comment by albop** [comment]
Sure, but I would first ask for an implementation of the vectorized algo. It is much easier to implement/understand, and it can't be useless.

**Comment by jstac** [comment]
> It is much easier to implement/understand

If you say so...

**Comment by TomHase** [comment]
Sorry, I was away last week and just saw this. 

Thank you a lot for the comments. I will try to implement the vectorized version of the algorithm this weekend. I will also look through the code to see where the speed issue might be coming from. 

@albop Sorry, I am not familiar with the notation for the higher order version `AX + BX.[C,...C] = D.` Is it just the kronecker product?

**Comment by TomHase** [comment]
@jstac, @albop, @mmcky I looked into this now. It is not straightforward to vectorize the algorithm, so I took a slightly different approach. The algorithm involves a least squares problem, which I solved using numpy's lstsq function. The more efficient way is to solve it using a QR decomposition with Givens rotations. That alone sped up the code quite a bit. I also made a few other, smaller changes that helped with the speed. 

But if I perform the same speed comparison as above, the code is still significantly slower than the dolo implementation. In fact, with (20 x 20) matrices, inverting the vectorized system directly is the fastest method.  

```
n=20
A = np.random.random((n,n))
B = np.random.random((n,n))
C = np.random.random((n,n))
D = np.random.random((n,n))
I = np.eye(n)

def test_sylv(A, B, C, D, I):
    X = solve_sylvester([A,B],[I,C],D)
    test = A@X + B@X@C - D
    return abs(test).max()

def test_dolo_sylv(A, B, C, D):
    from dolo.numeric.matrix_equations import solve_sylvester
    X = solve_sylvester(A,B,C,-D)
    test = A@X+B@X@C - D
    return abs(test).max()

def test_direct(A, B, C, D, I):
    X = np.dot(np.linalg.inv(np.kron(I, A) + np.kron(C.T, B)), D.reshape(-1,1, order="F")).reshape(n,n, order="F")
    test = A@X+B@X@C - D
    return abs(test).max()

%time test_dolo_sylv(A, B, C, D)
#CPU times: user 59 ms, sys: 8.65 ms, total: 67.6 ms
#Wall time: 71.7 ms
#2.6600943670018751e-13

%time test_sylv(A, B, C, D, I)
#CPU times: user 1.65 s, sys: 16.8 ms, total: 1.66 s
#Wall time: 1.67 s
#6.5725203057809267e-14

%time test_direct(A, B, C, D, I)
#CPU times: user 24.6 ms, sys: 5.68 ms, total: 30.3 ms
#Wall time: 27.7 ms
#9.9320551782966504e-13
```

I also looked at the speed of the Julia implementation of the algorithm. It's actually slightly shower than the python implementation. 

```
n=20;
A = rand((n,n));
B = rand((n,n));
C = rand((n,n));
D = rand((n,n));
I = eye(n);

function test_sylv(A, B, C, D, I)
    X = solve_sylvester([A,B],[I,C],D)
    test = A*X + B*X*C - D
    return maximum(abs(test))
end    

%time test_sylv(A, B, C, D, I)
# 1.769597 seconds (2.44 M allocations: 1.820 GB, 21.32% gc time)
# 3.175237850427948e-14
```

**Comment by mmcky** [comment]
Hi @albop, when you have a chance would you be able to review / comment on this PR?

**Comment by albop** [comment]
@mmcky, @jstac and mostly @TomHase : sorry for the delay. I had mostly forgotten about this PR and realize I left my comments in a half-baked state. Basically, I think quantecon should include a vectorized solution of the system which will be enough to get going and close #47. As for this PR, I'm not sure what we should do, it depends a bit on our stance on experimental algorithm, and on whether we feel qe is the right place for it (maybe we could sponsor it for scipy.sparse.linalg).

When I mentioned a "vectorized" system, I was referring to the fact that the system to solve is linear in the unknown, hence there must be a linear representation where the unknown can be solved by inverting a matrix. One can apply the formulas from https://en.wikipedia.org/wiki/Vectorization_(mathematics) to vectorize `AX+BXC+D` into `(kron(A,I)+kron(B,C'))vec(X)+vec(D)` (row-major convention) where `vec(X)` is the vectorized form of unknown matrix `X`.
One can propose a simple implementation of the solve_sylvester routine as follows:
```
from numpy.linalg import solve
from numpy import kron
vec = lambda M: M.ravel()

# no check is done on function inputs to verify they have matching sizes
def solve_sylvester_vectorized(*args):
    n = args[0][0].shape[0]
    K = vec(args[-1])
    L = sum([kron(A,B.T) for (A,B) in args[:-1]])
    X = solve(L,-K)
    return X.reshape((n,n))
```
To get a quick idea of how fast it goes:
```
n = 20
A = np.random.random((n,n))
B = np.random.random((n,n))
C = np.random.random((n,n))
D = np.random.random((n,n))
I = np.eye(n)
t1 = time.time()
X = solve_sylvester_vectorized([A,I],[B,C],D)
t2 = time.time()
print(t2-t1)
```
On my laptop it takes `0.037` seconds.

In fact, I notice now that @TomHase has already implemented it above as `test_direct` but with a full matrix inversion which is slower.

After reflecting for a while, here is my take:
1/ quantecon could include something based on the simple inversion (that would probably work fine with small systems). That would basically resolve #47 .
2/ if there is interest, there could be an option based on slycot (without dependency added, the import being made in the function). 
3/ despite being slower in the current examples, I don't think the submitted PR is uninteresting at all
     a/ it is certainly at a disadvantage on examples with random coefficients compared with dense methods . Indeed the number of steps it takes to converge is at most `n` but sometimes much smaller for some systems. With random coefficients, it is so to say guaranteed to be as slow as possible.
     b/ By contrast, iterative methods, like the one Tom proposed could potentially have an edge on big systems with lots of structure. It's not clear to me though, what difference the proposed code makes w.r.t. standard gmres like the one supplied in scipy. The latter can also potentially be used to solve Sylvester systems but it's relevance is also going to be very context dependent. Tom: do you have a specific application in mind for your code?

TomHase:
Thank you for your comments @albop. For most economic applications the simple inversion method will probably be enough. When it comes to this PR, the question is whether quantecon wants to provide a function that has an advantage in larger systems with potentially many leading and trailing matrices. I am not sure how much demand there is for that. What do you think, @jstac and @mmcky ?

TomHase:
Hi @mmcky. I think it could be useful for `QuantEcon.py` to have a Sylvester equation solver. If you are happy with the `simple inversion method` we can close this PR.  

For the documentation, we could simply cite this paper, specifically the augmented regulator problem: https://www.minneapolisfed.org/research/sr/sr198.pdf

PR #317, opened 2017-07-24
Adds stability test to stationary_distributions method of LinearStateSpace class

sebgraves:
Adds stability test of eigenvalues of the A matrix to the stationary_distributions method.

The first step is to remove the row and column from the A matrix which corresponds to the constant in the state vector (whose location is automatically detected).

The next step is to check that the eigenvalues of the submatrix have modulus less than 1 in absolute value.

### DIFF
--- a/quantecon/lss.py
+++ b/quantecon/lss.py
@@ -262,7 +262,8 @@ class LinearStateSpace(object):
     def stationary_distributions(self, max_iter=200, tol=1e-5):
         r"""
         Compute the moments of the stationary distributions of :math:`x_t` and
-        :math:`y_t` if possible.  Computation is by iteration, starting from
+        :math:`y_t` if possible.  Tests for existence of stationary distribution.
+        If stationary distribution exists, computation is then by iteration, starting from
         the initial conditions self.mu_0 and self.Sigma_0
 
         Parameters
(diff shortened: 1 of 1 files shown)


jstac:
@sebgraves Thanks, this looks nicely put together.  
Would you be willing to add a test of stationarity that validates your logic in [this file](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/tests/test_lss.py)?  Usually when we add new logic to the library we add a corresponding test.

cc7768:
Haven't had a chance to look at this carefully yet, but my initial reaction is that there should be a new method `is_stable` (or something like that). We can (and should) check stability inside of `stationary_distributions`, but this functionality is more useful than just within that method.

mmcky:
@sebgraves thanks for this PR. What do you think re: suggestions from @cc7768?

sebgraves:
@cc7768  what other uses were you thinking of?

@jstac sure, I will work on adding a test to that file

cc7768:
One immediate additional use case is for `lss.geometric_sums(beta, x_t)`, though I think even if this weren't the case, breaking it out is natural because it is an independent idea and isn't directly tied to computing the stationary distribution. The formula we use in `geometric_sums` is only valid for stable matrices -- In this case it would be that the eigenvalues of `beta * A` are stable.

Example

```

import numpy as np
import quantecon as qe


A = np.array([[0.9, 0.2], [0.2, 0.9]])
B = np.array([[0.01, 0.0], [0.0, 0.01]])
C = np.array([[1.0, 1.0]])

mu0 = np.ones(2)
sigma0 = np.eye(2)*0.001

lss = qe.LinearStateSpace(A, B, C, mu_0=mu0, Sigma_0=sigma0)

T = 500
beta_stable = 0.5
betas_stable = np.array([beta_stable**i for i in range(T)])
beta_unstable = 0.999
betas_unstable = np.array([beta_unstable**i for i in range(T)])

reps = [lss.simulate(T)[0] for i in range(250)]

sum_stable = np.mean(np.array([np.sum(r*betas_stable, axis=1)
                     for r in reps]), axis=0)

sum_unstable = np.mean(np.array([np.sum(r*betas_unstable, axis=1)
                       for r in reps]), axis=0)

gsum_stable = lss.geometric_sums(beta_stable, np.ones(2))[0]
gsum_unstable = lss.geometric_sums(beta_unstable, np.ones(2))[0]

print("Stable Difference: ", np.max(np.abs(sum_stable - gsum_stable)))
print("Unstable Difference: ", np.max(np.abs(sum_unstable - gsum_unstable)))
```

This actually makes me wonder whether the right thing to do here is add a general function `is_stable` (perhaps into the file `utils/array.py` -- I imagine this function as just taking a matrix and checking the eigenvalues and return true as long as they are less than 1) and then use it elsewhere in `QE.py`. For `LinearStateSpace` we would have a special method to the class that does the automatic removal of the constants and then checks stability of the resulting matrix like we discussed, but my guess is that having access to an `is_stable` function would also be useful in the `LQ` and `Kalman` code (among other places).

Thoughts @sebgraves @mmcky @jstac ?

If people like this idea then I think this PR should (1) add an `is_stable` method to `utils/array.py`, (2) add a method to the `LinearStateSpace` class which does the automatic removal of constants and checks the stability of that matrix, and (3) adds stability checks (using the new method) to the two `LinearStateSpace` methods they are needed (`stationary_distributions` and `geometric_sums`). We can take care of adding stability checks to other files in the future (I view that as out of scope for this PR). I'm happy to hear alternative recommendations.

jstac:
I agree with @cc7768 's argument that it should be an independent function.  Even aside from separate usage, it contains an independent chain of logic and hence lives more naturally in a distinct function that acts on an instance of `LinearStateSpace`.  

The other ideas that @cc7768 seem reasonable and I don't see any cost to putting the function in `utils/array.py`.

mmcky:
Hi @sebgraves would you mind to give me an update on this PR when you get the chance? Thanks.

mmcky:
Hi @sebgraves let me know if you intend to update this PR with the provided feedback. Thanks.

sebgraves:
Hi @mmcky. I will happily try and do the 3 things that @cc7768 mentioned at the end of his last message. It seems they've already been added to LSS.jl.

mmcky:
thanks @sebgraves. I will leave this PR open until that is done and then we can merge. Cheers.

mmcky:
This PR has been incomplete for some time. I am inclined to close and leave the branch in the event updates are made. Please re-open the PR.

PR #318, opened 2017-07-31
ENH: added routines to convert ddp between full and SA formulations

sglyon:
Sometimes I find it helpful to have convenient, even if not optimally efficient, routines for converting between the SA and full formulations. 

This PR implements such routines.

### DIFF
--- a/quantecon/markov/ddp.py
+++ b/quantecon/markov/ddp.py
@@ -456,6 +456,73 @@ class DiscreteDP(object):
                     'violated for state {s}'.format(s=s)
                 )
 
+    def to_sa_formulation(self, sparse=True):
+        """
+        Convert this instance of `DiscreteDP` to SA-pair form
+
+        Parameters
+        ----------
+        sparse : bool, optional(default=True)
+            Should the `Q` matrix be stored as a sparse matrix.
+            If true the CSR format is used
+
+        Returns
+        -------
+        ddp_sa : DiscreteDP
+            The correspnoding DiscreteDP instance in SA form
+
+        Notes
+        -----
+        If this instance is already in SA form then it is returned
+        un-modified
+        """
+
+        if self._sa_pair:
+            return self
+        else:
+            s_ind, a_ind = np.where(np.isfinite(self.R))
+            RL = self.R[s_ind, a_ind]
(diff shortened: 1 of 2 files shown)


quantecon/markov/ddp.py
```diff
@@ -456,6 +456,73 @@ def _check_action_feasibility(self):
...
+        Notes
+        -----
+        If this instance is already in full form then it is returned
+        un-modified
+
+        """
+        if self._sa_pair:
+            ns = self.num_states
+            na = self.a_indices.max() + 1
+            R = np.zeros((ns, na))
```
oyamad: Maybe this should be `np.full((ns, na), -np.inf)`?

quantecon/markov/ddp.py
```diff
@@ -456,6 +456,73 @@ def _check_action_feasibility(self):
...
+        Notes
+        -----
+        If this instance is already in SA form then it is returned
+        un-modified
+        """
+
+        if self._sa_pair:
+            return self
+        else:
+            s_ind, a_ind = np.where(np.isfinite(self.R))
```
oyamad: Better to write `self.R > -np.inf` instead of `np.isfinite(self.R)`?
(`self.R` may contain `np.inf`.)

oyamad:
A comment on terminology:
I guess the terminologies should be consistent between the  code and the documentation: "product formulation" should be replaced "full formulation" in the documentation, or vice versa.

(My view is that an SA formulation is already a "full" description of the model when the sets `A(s)` of available actions at state `s` are given excluding infeasible actions.)

sglyon:
Thanks for the review. I 100% agree on all accounts and will make the changes in a few hours when I'm back at a computer

oyamad:
In the test code, maybe better to assign `-np.inf` to some entry in `R` and `0` in `Q`.
Otherwise the code looks good to me.

sglyon:
Good call @oyamad -- I've added these tests.

mmcky:
Thanks @sglyon and @oyamad for review. This looks good to me. I will leave open one more day and then merge.

mmcky:
Thanks @sglyon.

PR #321, opened 2017-08-04
Fix class style

lbui01:
This PR changes the class style (issue: jstac/QuantEcon.lectures#497).

Specifically, the PR changes `class Foo(object):` to `class Foo:`.

Note: `object` used to be required to provide some class infrastructure like `super` and `@property` in `python2.7` which is no longer required in `python3.X` as inheritance of these things are now implicit.

### DIFF
--- a/quantecon/arma.py
+++ b/quantecon/arma.py
@@ -13,7 +13,7 @@ import matplotlib.pyplot as plt
 from scipy.signal import dimpulse, freqz, dlsim
 
 
-class ARMA(object):
+class ARMA:
     r"""
     This class represents scalar ARMA(p, q) processes.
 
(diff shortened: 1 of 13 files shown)


jstac:
Thanks @lbui01

jstac:
@mmcky Sorry, I thought this was still lecture code, not QuantEcon.py!

Are you OK with these changes?  It means we're no longer supporting 2.x, but perhaps that was already the case?

If you are OK I'll leave you to delete the branch.

mmcky:
@jstac this is fine. ``QuantEcon.py`` only supports ``3.5+`` as of version ``0.3.3``.

PR #323, opened 2017-08-09
Test lemke howson exceptions

### LINKED ISSUE #302: TEST: Add tests for game_theory/lemke_howson.py

QBatista:
Added tests for exceptions to lemke_howson.py and raise type error when init_pivot type is not an integer instance.

Fix #302

### DIFF
--- a/quantecon/game_theory/lemke_howson.py
+++ b/quantecon/game_theory/lemke_howson.py
@@ -5,6 +5,7 @@ Compute mixed Nash equilibria of a 2-player normal form game by the
 Lemke-Howson algorithm.
 
 """
+import numbers
 import numpy as np
 from numba import jit
 from .utilities import NashResult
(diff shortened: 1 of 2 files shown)


mmcky:
thanks for the PR @QBatista and for the review @oyamad. 
I will leave this PR open for one more day for comment and merge. 

It appears that ``travis`` is in fact ``green`` and passing but the web integration isn't updating.

mmcky:
I reset travis and now showing green so will merge. Thanks again @QBatista and @oyamad

PR #327, opened 2017-08-18
game_theory.Player: Add is_dominated

oyamad:
* If `method=None`, `quantecon.game_theory. lemke_howson ` is used, while if `method='simplex'`, `scipy.optimize.linprog` is used with `method='simplex'`.
* [For a future PR] Accept `method='interior-point'` once scipy/scipy#7123 gets released.


mmcky:
thanks @oyamad. As soon as travis has rechecked this PR I will merge.

Issue #325, opened 2017-08-17
Remove plotting functionality from ARMA


jstac:
I propose that we remove the plot methods (the last 5 methods) from [the ARMA class](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/arma.py) and embed them in [the lecture](https://lectures.quantecon.org/py/arma.html) instead.

This seems a more natural place for code that produces plots and it also would bring the Python lecture more in line with the Julia lecture --- see the first code block here: https://lectures.quantecon.org/jl/arma.html#code

sglyon:
I think this would also remove the dependency on matplotlib, which would make quantecon even lighter-weight

oyamad:
Now [build is failing](https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/268220128) for some unknown reason, but triggered by `matplotlib` anyway, it is a good timing to remove the dependency on `matplotlib`. Is the task just to remove all the plot function from `arma.py`?

jstac:
I'll work on this now.   The code has to be shifted into the lecture site.

oyamad:
OK, thanks.

jstac:
Umm, so I just accidentally committed to master ;-)

https://github.com/QuantEcon/QuantEcon.py/commit/985d9b8bacf0841ff871540c3f19923ad6465355

But it looks like the build is still failing

jstac:
OK, so I also forgot to remove the import.  I'll do this in a moment.

mmcky:
This looks complete. Closing this issue.

PR #330, opened 2017-08-25
removed dep on matplotlib

jstac:
See #325


oyamad:
I think you should remove `matplotlib` from config files too. See 5567f27dde05cd3b0936ac047b435d470f0511ef.

jstac:
@oyamad Thanks.

PR #337, opened 2017-08-28
issue release of version 0.3.6.2 due to issues with 0.3.6.1 in setupt…

mmcky:
…ools


mmcky:
this will be the release of ``0.3.6`` to pypi and conda-forge. There was an issue with setup tools that needed to be resolved from ``0.3.6.1``.

PR #344, opened 2017-09-27
Add simplex_grid and simplex_index

oyamad:
From my old code https://github.com/oyamad/simplex_grid.
To be used in `game_theory`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -28,6 +28,7 @@ exclude_lines =
     if __name__ == .__main__.:
 
     # Don't count Numba nopython=True jit functions
+    @jit
     @jit\(.*nopython=True
     @njit
     @generated_jit\(.*nopython=True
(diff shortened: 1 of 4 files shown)


mmcky:
@oyamad do you know why coverage has decreased with this PR? It looks like there are tests that should cover the additional routines.

oyamad:
It is because of the [`@jit`](https://github.com/QuantEcon/QuantEcon.py/blob/37ff3458167e2ad2e7c9911f5f69c5c7cf819fc5/quantecon/gridtools.py#L127) decorator. It seems that a jitted function is not covered even without `nopython=True`.

mmcky:
ah - I should add the simple case for ``.coveragerc``

Currently we have:
```bash
    # Don't count Numba nopython=True jit functions
    @jit\(.*nopython=True
    @njit
```

mmcky:
is there a reason not to use ``njit``?

oyamad:
Because `num_compositions` is not jitted which uses `scipy.misc.comb`.

I could have implemented `scipy.misc.comb` with Numba, but I decided not to because it is not really necessary (`@jit` just works fine) and it is a nontrivial job to handle overflow.

PR #352, opened 2017-10-19
ENH: jit the 1d quadrature routines

sglyon:
This just jits the 1-dimensional quadrature routines.

The multi-dim ones are implemented using higher order functions, so they still will not work in nopython mode.

### DIFF
--- a/quantecon/quad.py
+++ b/quantecon/quad.py
@@ -595,14 +595,14 @@ def qnwgamma(n, a=1.0, b=1.0, tol=3e-14):
     n : int or array_like(float)
         A length-d iterable of the number of nodes in each dimension
 
-    mu : scalar or array_like(float), optional(default=zeros(d))
-        The means of each dimension of the random variable. If a scalar
-        is given, that constant is repeated d times, where d is the
-        number of dimensions
+    a : scalar or array_like(float) : optional(default=ones(d))
+        Shape parameter of the gamma distribution parameter. Must be positive
 
-    sig2 : array_like(float), optional(default=eye(d))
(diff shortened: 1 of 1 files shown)


sglyon:
This is now ready for review/merge

quantecon/quad.py
```diff
@@ -573,7 +586,7 @@ def qnwbeta(n, a=1.0, b=1.0):
     return _make_multidim_func(_qnwbeta1, n, a, b)
 
 
-def qnwgamma(n, a=None):
+def qnwgamma(n, a=1.0, b=1.0, tol=3e-14):
```
mmcky: @sglyon the function signature seems to no longer correspond to the docstring.

mmcky:
Nice -- thanks @sglyon.

mmcky:
Does this need any additional tests. 
This isn't usually the case as the `jit` is usually tested as part of an overall function.

mmcky:
@sglyon do you by chance have any performance comparison. No worries if you don't but if you had a ROM then I would include that in the release notes.

sglyon:
I don't have performance comparisons.

I did this not because of potential performance gains from calling it, but rather so I can call these routines inside other `nopython`  jitted functions.

Good catch regarding the docstring, I'll fix that up

mmcky:
Thanks @sglyon - ok cool - that's a good reason. When (!) I get performance tracking up and running then it will pick up these kind of commits anyway from the git history.

sglyon:
Ok, docstrings updated!

mmcky:
thanks @sglyon.

PR #353, opened 2017-10-19
fix kalman class docstring

natashawatkins:
Needed a new line after an equation

### DIFF
--- a/quantecon/kalman.py
+++ b/quantecon/kalman.py
@@ -19,7 +19,7 @@ class Kalman:
 
     .. math::
 
-        x_{t+1} = A x_t + C w_{t+1}
+        x_{t+1} = A x_t + C w_{t+1} \\
         y_t = G x_t + H v_t
 
     Here :math:`x_t` is the hidden state and :math:`y_t` is the measurement.


quantecon/kalman.py
```diff
@@ -19,7 +19,7 @@ class Kalman:
 
     .. math::
 
-        x_{t+1} = A x_t + C w_{t+1}
+        x_{t+1} = A x_t + C w_{t+1} \\
```
mmcky: @natashawatkins should these be aligned?

natashawatkins: No I wouldn't say so - they're different equations

jstac: Not a big deal but it would be natural to align these equations on the equals sign.

mmcky: I ask as they will be right aligned (by default)

image

natashawatkins: I think it will make it quite messy when you view the docstring in jupyter for example (which simply displays the raw string and not latex)

I can't see it done anywhere else in the docs

You would need to use the align environment right?

jstac: mmm, good point.  perhaps it's best as is.  probably not worth spending more time on.

mmcky: I was just about to say the same thing :).

PR #355, opened 2017-10-21
RBLQ: add pure forecasting case 

szokeb87:
Adds the following case:
If Q and B are zero matrices, RBLQ assumes that the user wants to solve a "pure forecasting problem" defined on p171 in Robustness (Hansen and Sargent, 2008).

It is assumed that the user sets up the problem in a way that "eliminates" F.  This involves redefining the R and A matrices in equations (1) and (2) (of [this lecture](https://lectures.quantecon.org/py/robustness.html)) as 
R_adj = R + F'QF 
A_adj = A - BF
Then, using R_adj, A_adj, C with Q=B=0 (zero matrices), RBLQ returns a K which is the worst-case with respect to the proposed policy F.

### DIFF
--- a/quantecon/tests/test_robustlq.py
+++ b/quantecon/tests/test_robustlq.py
@@ -1,5 +1,5 @@
 """
-Author: Chase Coleman
+Author: Chase Coleman and Balint Szoke
 Filename: test_robustlq
 
 Tests for robustlq.py file
--- a/quantecon/tests/test_robustlq.py
+++ b/quantecon/tests/test_robustlq.py
@@ -1,8 +1,8 @@
 """
 Author: Chase Coleman
-Filename: test_lqcontrol
+Filename: test_robustlq
 
-Tests for lqcontrol.py file
+Tests for robustlq.py file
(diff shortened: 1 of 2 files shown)


jstac:
@szokeb87 Many thanks.
@mmcky For background, @szokeb87 has added this functionality in response to

http://discourse.quantecon.org/t/robust-lqc-special-case/281

mmcky:
Thanks @szokeb87. Can we please add this case to the ``test_robustlq`` suite?

szokeb87:
@mmcky Done

quantecon/tests/test_robustlq.py
```diff
@@ -1,8 +1,8 @@
 """
 Author: Chase Coleman
```
cc7768: Seems like your name would fit better here :wink:

szokeb87:
@cc7768 this was just the cherry on the cake :)

mmcky:
Thanks @szokeb87 this is looking great.

Issue #354, opened 2017-10-20
Getting rid of the LinearStateSpace argument in Kalman


natashawatkins:
Wouldn't it be easier to just have the arguments that a `LinearStateSpace` object takes available when creating a `Kalman` object? It's somewhat cumbersome to first create the LSS object and then put that into the Kalman filter (you also need to import two classes).

We could just create a LSS object using the arguments in the kalman filter.

jstac:
Yeah, fair point.  I had a look at the  code and the Kalman class never actually uses any of the functionality of the `LinearStateSpace` class internally.  So from a coding perspective there's no advantage linking them.

Plus, as you say, if we do need such functionality down the road we can always build an instance of LSS inside the class.

Could you please put together a PR that closes this issue?  We'll also need to coordinate on changing the lecture.  Although that shouldn't happen until the next release of the library. (CC @mmcky )

mmcky:
Thanks @jstac and @natashawatkins. 
I was thinking of putting a release together but I will do so after this is resolved.

cc7768:
I think this is a reasonable thing to do.

If you wanted to make sure you going from `LSS` to `Kalman` was low cost then one natural thing to do would be to add a method to `LSS` that created a `Kalman` object.

thomassargent30:
Lets talk with John before changing this please. It makes sense not to
change it.  Tom

On Oct 23, 2017 9:40 AM, "Chase Coleman" <notifications@github.com> wrote:

I think this is a reasonable thing to do.

If you wanted to make sure you going from LSS to Kalman was low cost then
one natural thing to do would be to add a method to LSS that created a
Kalman object.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/QuantEcon/QuantEcon.py/issues/354#issuecomment-338681143>,
or mute the thread

jstac:
Hi @thomassargent30, I think @natashawatkins has a good point here: To build an instance of `Kalman` you need to know about `LinearStateSpace`.  This makes sense on a mathematical level but forms a barrier to entry when coding, since users need to look up the documentation for two classes.

If in the future we need to use the functionality of `LinearStateSpace` inside `Kalman`, we can build an instance of `LinearStateSpace` inside `Kalman`, rather than outside.

So on balance I support the change proposed by @natashawatkins.

thomassargent30:
Go with what John and Natasha think is best.  There will be some general
[...]

PR #368, opened 2017-10-26
WIP: Update Kalman Interface to accept setup matrices directly (in addition to LSS)

mmcky:
a proof of concept trial for accepting multiple interfaces for Kalman object, 

**Note:** no updated docstrings yet. Setting up in Branch for collaboration. 

This would allow `kalman` to accept the lecture example:

```python
# == parameters == #
theta = 10  # Constant value of state x_t
A, C, G, H = 1, 0, 1, 1
ss = LinearStateSpace(A, C, G, H, mu_0=theta)

# == set prior, initialize kalman filter == #
x_hat_0, Sigma_0 = 8, 1
kalman = Kalman(ss, x_hat_0, Sigma_0)
```
in addition to

```python
# == parameters == #
theta = 10  # Constant value of state x_t
A, C, G, H = 1, 0, 1, 1
x_hat_0, Sigma_0 = 8, 1
kalman = Kalman(A, C, G, H, mu_0=theta, x_hat_0, Sigma_0)
```

the key assumption is to check if the first element of the arguments list is an ``LSS`` object then it init's the object class (as per previously implemented) otherwise it constructs an ``LSS`` object internally with the matrices and then set's up using an ``LSS`` object.

@natashawatkins is this what you had in mind for #354? I think you were suggesting to replace the acceptance of ``LSS`` altogether but the benefit here is that old code won't break and Kalman will continue to work with those packaging up A, C, and G in an ``LSS`` object.

The downside to this approach is it makes the introspected interface to `Kalman` harder to read due to accepting arbitrary arguments and keyword arguments to parse for various init methods. 

The **key assumption** here is that the argument list is checked and if the first element of the passed argument list is type ``LinearStateSpace`` then ``_init_lss`` is used to populate the object otherise it is assumed that a user has passed ``A, C, and G`` matrices.

### DIFF
--- a/quantecon/kalman.py
+++ b/quantecon/kalman.py
@@ -61,11 +61,30 @@ class Kalman:
 
     """
 
-    def __init__(self, ss, x_hat=None, Sigma=None):
+    def __init__(self, *args, **kwargs):
+        args = list(args)
+        if type(args[0]) is LinearStateSpace:
+            self._init_lss(*args, **kwargs)
+        else:
+            self._init_matrix(*args, **kwargs)
+
+    def _init_lss(self, ss, x_hat=None, Sigma=None):
+        """ setup object using a LinearStateSpace as the first argument
+        """
         self.ss = ss
         self.set_state(x_hat, Sigma)
         self.K_infinity = None
         self.Sigma_infinity = None
+
+    def _init_matrix(self, *args, **kwargs):
+        """ setup object by passing A, C, and G matrices directly
+        """
+        kwargs = dict(**kwargs)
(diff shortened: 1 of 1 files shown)


natashawatkins:
Yeah I was thinking of just getting rid lss argument

I'm not sure - it kind of overcomplicates the code, and I think our library generally is useful as template

But it is handy in case someone wants to use both a linear state space and the kalman object.

mmcky:
I am cleaning up some old PR's and it doesn't seem like this is that useful. It is probably best to choose either `matrix` or `LSS` entry point to reduce complexity of the code for maintenance. I propose to close this **without** merge.

Issue #304, opened 2017-04-26
Migrate to using setuptools rather than distutils for package setup


mmcky:
``setuptools`` seems to be a better solution, with some additional features over ``distutils`` that make packaging python packages easier. 

For example it will allow ``develop`` environment

```
Deploy your project in “development mode”, such that it’s available on sys.path, yet can still be edited directly from its source checkout.
```

mmcky:
https://github.com/kennethreitz/setup.py

mmcky:
See PR #369

PR #369, opened 2017-10-26
add ability to use setuptools in preference to distutils.core

mmcky:
One nice feature is ability to use ``python setup.py develop``. 
This sets up a link between ``cwd`` and ``site-packages`` for ``install`` and enables quick testing etc.


mmcky:
Fixes #304

PR #375, opened 2017-12-23
FIX: Update `num_compositions`

QBatista:
Use `scipy.special.comb` instead of `scipy.misc.comb` as `scipy.misc.comb` is deprecated in scipy 1.0.0.

### DIFF
--- a/quantecon/gridtools.py
+++ b/quantecon/gridtools.py
@@ -280,5 +280,5 @@ def num_compositions(m, n):
         Total number of m-part compositions of n.
 
     """
-    # docs.scipy.org/doc/scipy/reference/generated/scipy.misc.comb.html
+    # docs.scipy.org/doc/scipy/reference/generated/scipy.special.comb.html
     return scipy.special.comb(n+m-1, m-1, exact=True)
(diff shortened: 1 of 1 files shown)


quantecon/gridtools.py
```diff
@@ -281,4 +281,4 @@ def num_compositions(m, n):
 
     """
     # docs.scipy.org/doc/scipy/reference/generated/scipy.misc.comb.html
```
oyamad: Please update this commit as well.

sglyon:
LGTM

Do we need to update a requirements.txt or something like that to note that these changes require spicy >= 1.0?

oyamad:
`comb` was moved to `scipy.special` some time ago, by https://github.com/scipy/scipy/commit/c430bed181b507a24edab51b77a7a05acd87c023.

mmcky:
@sglyon most of our requirements files don't specify versions (so they should default to latest). But are you suggesting we should put ``scipy > 1.0``?

oyamad:
@mmcky I don't think we need specify a version for `scipy`. `comb` has been there in `scipy.special` since `0.14`, and the current Anaconda has `scipy` `0.19.1`.

PR #376, opened 2017-12-23
FEAT: Add `expected_payoff` and `expected_payoff_2p`

QBatista:
Adds a feature to calculate the expected payoff for a player's action given his opponents' actions. Here is the performance:

```
coordination_game_matrix = np.array([[4, 0], [3, 2]])
player0 = gt.Player(coordination_game_matrix)
s0 = np.array([1/2, 1/2])
s1 = np.array([2/3, 1/3])

%timeit expected_payoff(player0, s0, [s1])
%timeit expected_payoff_2p(coordination_game_matrix, s0, s1)
```
```
28.6 µs ± 515 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
682 ns ± 6.77 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

### DIFF
--- a/quantecon/game_theory/__init__.py
+++ b/quantecon/game_theory/__init__.py
@@ -3,7 +3,7 @@ Game Theory SubPackage
 
 """
 from .normal_form_game import Player, NormalFormGame
-from .normal_form_game import pure2mixed, best_response_2p
+from .normal_form_game import pure2mixed, best_response_2p, expected_payoff_2p
 from .random import random_game, covariance_game
 from .pure_nash import pure_nash_brute, pure_nash_brute_gen
 from .support_enumeration import support_enumeration, support_enumeration_gen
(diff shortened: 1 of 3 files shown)


oyamad:
I am skeptical whether we really need these functions.

* We already have `Player.payoff_vector()`: one can do

  ```py
  s0 @ player0.payoff_vector([s1, s2])
  ```

  where `s1` and `s2` may be a scalar (pure action) or an array (mixed action).

* For two-player case, one can do

  ```py
  s0 @ player0.payoff_array @ s1
  ```

QBatista:
I didn't know that you could calculate it this way which is why I wrote these. Using functions seems more intuitive to me and would make it more straightforward for the user. The performance of `expected_payoff_2p` should also be better as long as the payoff matrix is not large.

QBatista:
@mmcky This PR can be closed.

PR #379, opened 2018-01-04
Re-implement `next_k_array`; add `k_array_rank`

oyamad:
1. Re-implement `next_k_array` which is now based on "Algotirhm T" in [Generating All Combinations](http://www.kcats.org/csci/464/doc/knuth/fascicles/fasc3a.pdf) by Knuth.

2. Add `k_array_rank` and `k_array_rank_jit`:
   * `k_array_rank` works with pure Python's arbitrary precision integer.
   * `k_array_rank_jit` is Numba jit compiled and limited by the range of `np.intp`.

3. `_next_k_array` in `support_enumeration` is replaced by the new `next_k_array`:
   * The previous `_next_k_array` is implemented by "Gosper's hack" (see https://github.com/QuantEcon/QuantEcon.py/pull/263#issue-178525924) which has a limit in the size of the input array by the number of bits of `np.int`.
   * The new `next_k_array` has no such limitation.
   * The performance does not change almost at all.

### DIFF
--- a/quantecon/game_theory/support_enumeration.py
+++ b/quantecon/game_theory/support_enumeration.py
@@ -14,6 +14,7 @@ Tardos, and V. Vazirani eds., Algorithmic Game Theory, 2007.
 import numpy as np
 from numba import jit
 from ..util.numba import _numba_linalg_solve
+from ..util.combinatorics import next_k_array
 
 
 def support_enumeration(g):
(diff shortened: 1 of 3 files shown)


mmcky:
thanks @oyamad I like the approach to using ``util/combinatorics``. I will merge later this afternoon.

PR #387, opened 2018-02-03
 Fix tac, toc, loop_timer to return float

### LINKED ISSUE #386: Return value of timing functions
* I think `toc` and `tac` should return a `float`, and there is no point of returning a `str`.
* `digits` should be used to control the number of digits of the **printed** value, not of the returned value.

See also #322.

oyamad:
Close #386

### DIFF
--- a/quantecon/util/tests/test_timing.py
+++ b/quantecon/util/tests/test_timing.py
@@ -12,57 +12,26 @@ from quantecon.util import tic, tac, toc, loop_timer
 
 class TestTicTacToc():
     def setUp(self):
-        self.h = 0.1 
+        self.h = 0.1
         self.digits = 10
-    
+
     def test_timer(self):
         tic()
 
         time.sleep(self.h)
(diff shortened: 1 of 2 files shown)


mmcky:
@oyamad from a backwards compatibility perspective are there any changes to usage? 
I don't think so as I see you've used `verbose=True` as a default so the floats will print.

mmcky:
thanks @oyamad this looks good.

oyamad:
Yes, this is backward incompatible:

| Version  | Output type     | Print option   | Output option |
|----------|-----------------|----------------|---------------|
| -0.3.6.2 | `float`         | -              | -             |
| 0.3.7    | `str` or `None` | `print_output` | `output`      |
| This PR  | `float`         | `verbose`      | -             |

mmcky:
@oyamad thanks. This is mainly used as a utility so I think that is fine. I will just do a search across the repo to make sure there are no other changes.

PR #389, opened 2018-02-04
MAINT: Use `np.ix_` to extract submatrix

oyamad:
Replace `matrix[indices, :][:, indices]` with `matrix[np.ix_(indices, indices)]`.

I happened to find [this fact](https://stackoverflow.com/questions/4257394/slicing-of-a-numpy-2d-array-or-how-do-i-extract-an-mxm-submatrix-from-an-nxn-ar#comment4623697_4257708).

### DIFF
--- a/quantecon/graph_tools.py
+++ b/quantecon/graph_tools.py
@@ -351,7 +351,7 @@ class DiGraph:
             A DiGraph representing the subgraph.
 
         """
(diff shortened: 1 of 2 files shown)


mmcky:
reference to docs: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ix_.html

PR #391, opened 2018-03-07
FEAT: Add jitted function for drawing

### LINKED ISSUE #390: Replace DiscreteRV with jitted function
It would be nice to make this functionality useful by replacing the class with a simple jitted function --- which can then be embedded in other jitted functions.

Suggested new syntax:

`draws = discrete_rv(p_vals, x_vals, size=1000)`

QBatista:
Adds a jitted function for drawing discrete random variables. Note that `DiscreteRV.draw` no longer uses `check_random_state` as Numba supports top-level functions from the numpy.random module, but does not allow you to create individual RandomState instances.

Close #390

### DIFF
--- a/quantecon/discrete_rv.py
+++ b/quantecon/discrete_rv.py
@@ -11,7 +11,8 @@ specified vector of probabilities.
 import numpy as np
 from numpy import cumsum
 from numpy.random import uniform
-from .util import check_random_state
+from numba import jit
+
 
 class DiscreteRV:
     """
@@ -21,13 +22,13 @@ class DiscreteRV:
     Parameters
     ----------
     q : array_like(float)
-        Nonnegative numbers that sum to 1
+        Nonnegative numbers that sum to 1.
 
     Attributes
     ----------
     q : see Parameters
     Q : array_like(float)
-        The cumulative sum of q
+        The cumulative sum of q.
 
     """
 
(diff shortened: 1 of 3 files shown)


oyamad:
Some comments:

* Why not keep the `DiscreteRV` class as is, just adding the new jitted `discrete_rv` function? A `DiscreteRV` instance cannot be passed to a jitted function anyway (unless it is implemented as a `jitclass`).

* To me, `discrete_rv` sounds like returning some object representing a discrete random variable. What about `random_choice`? Note that this is similar to [`np.random.choice`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.choice.html), where Numba does not support the `p` option (see [this](http://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#permutations)).

quantecon/discrete_rv.py
```diff
@@ -68,20 +69,46 @@ def draw(self, k=1, random_state=None):
...
-            An array of k independent draws from q
+            An array of k independent draws from q.
 
         """
-        random_state = check_random_state(random_state)
+        return discrete_rv(self._q, self.Q, size=k, seed=seed)
+
+
+@jit(nopython=True)
+def discrete_rv(p_vals, cum_sum=None, size=1000, seed=0):
```
oyamad: The default value for `seed` must be `None`.

QBatista:
I agree with those comments. @jstac What do you think?

jstac:
@QBatista Many thanks.
@mmcky Do you mind if we merge this now?  It looks good for me and I need it to finish what I'm doing in the master branch of the lectures...

jstac:
@mmcky I'm going to go ahead and merge this, i hope that's OK.

Issue #367, opened 2017-10-26
kalman.py: Add `method` option


oyamad:
See https://github.com/QuantEcon/QuantEcon.py/pull/362#issuecomment-338985350

(The attributes `Sigma_infinity` and `K_infinity` should be implemented as `property`. EDIT: This part done by #396)

oyamad:
@mmcky Sorry I should have clarified, but #396 has done the task described in the second line in the initial comment.

The issue "Add `method` option" is still open.

PR #396, opened 2018-03-10
Implement Sigma_infinity and K_infinity as properties

mcsalgado:
As specified in https://github.com/QuantEcon/QuantEcon.py/issues/367#issue-268592941


mmcky:
Thanks @mcsalgado and @oyamad. Merging.

PR #397, opened 2018-03-11
FEAT: Add random.draw

### LINKED ISSUE #393: BUGs in random_choice
I tried the recently merged `random_choice` #391 and found the following:

1. This doesn't work with Numba v.0.36 with the following error:

   ```
   TypeError: Failed at nopython (inline calls to locally defined closures)
   sequence item 0: expected str instance, NoneType found
   ```

   It works with Numba v.0.37, so this should be a bug in Numba v.0.36. The problem for us is that the latest Anaconda contains Numba v.0.36.2.

2. Setting a seed with `np.random.seed` within a library (hidden from the user) is problematic: it seems to affect Numba's global random state.

   Suppose you have the following jit function:

   ```py
   @numba.jit(nopython=True)
   def my_rand(size):
       return np.random.uniform(0, 1, size)
   ```

   Call it after calling `random_choice` with the same seed (with Numba v.0.37):

   ```py
   qe.random_choice(np.array([0.1, 0.9]), seed=0)
   my_rand(5)
   ```

   Then it always returns the same sequence:

   ```py
   array([0.71518937, 0.60276338, 0.54488318, 0.4236548 , 0.64589411])
   ```

@jstac What is your use case where you need a jitted version of this functionality? What's wrong with generating random numbers outside your jit function and then using `np.searchsorted` (or [`qe.searchsorted`](http://quanteconpy.readthedocs.io/en/latest/util/array.html#quantecon.util.array.searchsorted)) inside the jit function?

oyamad:
Close #393.

(Also revert the changes by #391, as we should not have both `random_choice` and `random.draw`.)

### DIFF
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -167,7 +167,7 @@ def sample_without_replacement(n, k, num_trials=None, random_state=None):
 def draw(cdf, size=None):
     """
     Generate a random sample according to the cumulative distribution
-    given by `cdf`. Jit-complied by Numba in nonptython mode.
+    given by `cdf`. Jit-complied by Numba in nopython mode.
 
     Parameters
     ----------
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -3,9 +3,9 @@ Utilities to Support Random Operations and Generating Vectors and Matrices
 """
 
 import numpy as np
-from numba import jit, guvectorize
+from numba import jit, guvectorize, generated_jit, types
 
-from ..util import check_random_state
+from ..util import check_random_state, searchsorted
 
 
 # Generating Arrays and Vectors #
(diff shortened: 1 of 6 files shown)


jstac:
@mmcky I'm fully in favor of this PR and it would help me clean up master if we could move quickly on it.  Thanks.

quantecon/random/utilities.py
```diff
@@ -161,3 +161,46 @@ def sample_without_replacement(n, k, num_trials=None, random_state=None):
         return result[0]
     else:
         return result
+
+
+@generated_jit(nopython=True)
+def draw(cdf, size=None):
+    """
+    Generate a random sample according to the cumulative distribution
+    given by `cdf`. Jit-complied by Numba in nonptython mode.
```
QBatista: Did you mean "Jit-compiled by Numba in nopython mode"?

oyamad: Ah good catch, thanks.

mmcky:
I think this PR can be merged. @oyamad can you apply `ready` label when you're happy with it.

PR #399, opened 2018-03-13
RFC: Remove authors' name

### LINKED ISSUE #394: remove authors' names
I propose we remove authors' names from all files in line with standard open source conventions.

Example: https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/discrete_rv.py

It might be worth flagging this on QuantEcon.jl

QBatista:
Close #394 

The format of the description is not consistent across files. Here's my proposal to make it consistent:

```
"""
Filename: <filename>

<Description>

References (if applicable)
----------
 <References> 
"""
```

### DIFF
--- a/quantecon/util/timing.py
+++ b/quantecon/util/timing.py
@@ -1,8 +1,6 @@
 """
 Filename: timing.py
 
-Authors: Pablo Winant
-
 Date: 10/16/14
 
 Provides Matlab-like tic, tac and toc functions.
--- a/quantecon/util/timing.py
+++ b/quantecon/util/timing.py
@@ -1,8 +1,4 @@
 """
-Filename: timing.py
-
-Date: 10/16/14
-
 Provides Matlab-like tic, tac and toc functions.
 
 """
--- a/quantecon/arma.py
+++ b/quantecon/arma.py
@@ -1,6 +1,4 @@
 """
-Filename: arma.py
-
 Provides functions for working with and visualizing scalar ARMA processes.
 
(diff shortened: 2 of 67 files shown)


oyamad:
I would say we should also remove the `Filename` entry, as well as `Date` and `@date`.

For example: http://quanteconpy.readthedocs.io/en/latest/game_theory/random.html
"Filename: random.py" adds no information given the header "**random**".

QBatista:
That's fine for me too.

jstac:
Thanks @QBatista, much appreciated.
@oyamad I agree

quantecon/util/timing.py
```diff
@@ -1,8 +1,6 @@
 """
 Filename: timing.py
```
mmcky: this could be deleted also.

quantecon/util/timing.py
```diff
@@ -1,8 +1,6 @@
...
 
-Authors: Pablo Winant
-
 Date: 10/16/14
```
mmcky: and this

QBatista: Done, thanks!

mmcky:
thanks @QBatista when you're happy with this PR can you attach the `ready` label?

QBatista:
@mmcky I think I don't have access to label PRs yet

mmcky:
oh - I didn't realize that wasn't accessible. I have added the label and will merge tomorrow. I usually leave it a day or so at ``ready`` status.

PR #405, opened 2018-03-23
Add hamilton filter

Shunsuke-Hori:
This PR adds "Hamilton filter" from [_Why You Should Never Use the Hodrick-Prescott Filter_](http://econweb.ucsd.edu/~jhamilto/hp.pdf)

Proposition 4 in the paper shows that the error term v_{t+h} of the regression

y_{t+h} = beta_0 + beta_1 y_{t} + ... + beta_p y{t-p+1} + v_{t+h}

gives cyclical component.

If `*arg` (p in the paper) is not passed, the true process is assumed to be a random walk. Then, we no longer need to do regression and we just take the difference.

Example:
~~~
df['y_c'], df['y_t'] = hamilton_filter(df['y'], 8, 4) # apply to data `y` in pd.dataframe
df['y_c'], df['y_t'] = hamilton_filter(y, 8, 4) # apply to data `y` in np.array

# random walk
df['y_c'], df['y_t'] = hamilton_filter(df['y'], 8) # apply to data `y` in pd.dataframe
df['y_c'], df['y_t'] = hamilton_filter(y, 8) # apply to data `y` in np.array
~~~


Shunsuke-Hori:
@jstac If more filters are added, we may be better to make it. Al least, I don't have any plan to add more filters now.

Two questions:
- Is there any way to test locally? Would anyone mind telling me how?
- Below is the difference between python and matlab output; python gives larger output for all periods. Is it wierd? too large? If so, does anyone know why?
dif

Shunsuke-Hori:
Thank you @mmcky, I can run the test locally now.

It's ready for review.

Shunsuke-Hori:
Regarding the difference with matlab output, original matlab code is using `inv(X'*X)*X'y` instead of `(X'*X)\X'y`. With `\`, output becomes close.

mmcky:
@Shunsuke-Hori can you please update the `description` to the PR to document what this routine does, and any references of how it was implemented etc, a simple example for how it can be used etc.

Shunsuke-Hori:
@mmcky Thank you for pointing it out. It's updated.

mmcky:
thanks @Shunsuke-Hori this is looking good. I think we can leave it in ``filter`` for now and when the time comes to add new filters we may setup a ``filters`` sub-module. I have marked this as ready and will allow a day for comments.

mmcky:
thanks @Shunsuke-Hori for this contribution. Merging now.

PR #407, opened 2018-05-07
FIX: fix release year

mirca:


mmcky:
thanks @mirca. Good catch.

PR #409, opened 2018-05-14
add minimum version number for numba support

mmcky:
This PR adds ``0.36.2`` as a minimum numba version required for installation in ``setup.py``

``numba`` is one of the key dependencies for QuantEcon.py. Given our user base -- it may be helpful to alternatively write a version check in the base ``__init__.py`` file which would give us the opportunity to issue a more useful error message based on ``conda``.

Could check if ``numba.__version__ > minimum value`` otherwise print out instructions to upgrade such as ``conda install numba=0.37``.  Any comments welcome.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,6 @@ setup(name='quantecon',
       author_email='john.stachurski@gmail.com',
       url='https://github.com/QuantEcon/QuantEcon.py',  # URL to the repo
       download_url='https://github.com/QuantEcon/QuantEcon.py/tarball/' + VERSION,
-      keywords=['quantitative', 'economics']
+      keywords=['quantitative', 'economics'],
       install_requires=['numba>=0.36.2']
       )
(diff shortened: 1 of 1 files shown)


oyamad:
I think adding a minimum version in `setup.py` (i.e., the current status of the PR) is sufficient (isn't it always the case that the user who updates `QE` to the version that contains this patch will automatically update Numba to the minimum version, if below it?).

Another issue: should we list all the requirements, including `numpy` and `scipy`, in `install_requires` in `setup.py`?

mmcky:
> isn't it always the case that the user who updates QE to the version that contains this patch will automatically update Numba to the minimum version, if below it?

thanks @oyamad. You're right - that should be true although I haven't tested it. 

I guess I was thinking of only ``minimum`` dependencies that require version constraints, such as numba where the project uses many of the latest features available. But you're probably right - we should have the full list of dependencies (just without version number if not required).

oyamad:
Another related issue: What is the status of [`pip-requirements.txt`](https://github.com/QuantEcon/QuantEcon.py/blob/master/pip-requirements.txt)? (And what is the role of that file?)

For example, `statsmodels` is no longer used in the library.

mmcky:
Merging now will address other comments in #410

Issue #395, opened 2018-03-09
Adding inequality/segregation indices


cdagnino:
Hi,

Do you think adding ways to calculate inequality/segregation would fit the scope of the quantecon project?
I already have one [in this repository](https://github.com/cdagnino/ReardonSegregationIndex) that I could contribute.

jstac:
@cdagnino My instinct is yes, and the code looks well written.  thanks for offering.  My hesitation is that this code is code is quite isolated from what we have so far.  If it was part of a broader set of code on inequality and segregation measures that would certainly be better.

@mmcky ?

mmcky:
@cdagnino I am keen to diversify the library to include a distributed set of tools that are useful to economists. I will put some time this week into thinking how best we could organise this material, as I have been interested in assembling a ``datalab`` submodule that packages data related routines together.  

My main question is how best to organise the code to sensibly accomodate a variety of different functions that are available from research that may compute the same (or similar) type of output through different methodologies. It would also be good to lay the foundation to allow for a collection of tools for computing inequality and segregation measures (as @jstac suggests). @cdagnino do you use other metrics in your work that would be useful to have in a library?

cdagnino:
Hi,

@jstac @mmcky  I agree 100% with the isolation problem of the code I linked.
If y'all have some thoughts on the interface/syntax for this `datalab` submodule (which I'm guessing would share some ideas from the more general `Quantecon` module) then I can adapt my code to give a first taste of inequality/segregation measures.

I think there are a bunch of inequality and segregation indices that could be interesting to people. Besides the one I coded, a Gini index, a Generalized entropy measure (with Between + Within decompositions) and Duncan Index might be interesting and could be easily made compatible with pandas dataframes. [See this documents for some reference on inequality indices](http://siteresources.worldbank.org/PGLP/Resources/PMch6.pdf)

The "point" estimates should be fine and relatively easy to code. Adding standard errors could be interesting, but more difficult and maybe outside the scope of this package.

jstac:
I think these indices are a great example of what we could usefully code up for economists.  And if there's a number of them we can consider making a subpackage.

@cdagnino Might you be willing to take the lead in developing this subpackage, if we decide to go ahead?  We can probably devote RA resources to helping out, in that case...

cdagnino:
@jstac I'd be interested, but I promised myself not to undertake anything semi-big until I finish my dissertation. I'd be happy to contribute with code and ideas though.

[...]

PR #414, opened 2018-06-04
Added basic inequality mesasures: lorenz curve and gini

cdagnino:
First stab at creating inequality indices, as discussed in #395 

Tests added and  `nosetests` passed.

### DIFF
--- a/quantecon/inequality.py
+++ b/quantecon/inequality.py
@@ -4,7 +4,7 @@ Implements inequality and segregation measures such as Gini, Lorenz Curve
 """
 
 import numpy as np
-from numba import njit, prange, jitclass, float64, int64
+from numba import njit, prange
 
 
 @njit
@@ -56,16 +56,8 @@ def lorenz_curve(y):
     return cum_people, cum_income
 
 
-
-# Gini
-
-spec = [
-    ('n', int64),
-    ('y', float64[:])
-]
-
-@jitclass(spec)
-class Gini(object):
+@njit(parallel=True)
+def gini_coefficient(y):
     r"""
     Implements the Gini inequality index
 
@@ -74,11 +66,9 @@ class Gini(object):
     y : array_like(float)
         Array of income/wealth for each individual. Ordered or unordered is fine
 
-    Attributes
-    ----------
-    y : as above
-    n : scalar(int)
-        length of array y
+    Returns
+    -------
+
 
 
     References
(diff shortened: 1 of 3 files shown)


jstac:
Thanks @cdagnino 

For me, `prange` is not working once you wrap computation of the Gini coefficient in a `jitclass`.  I can see that by monitoring system resources.

Since `jitclass` is still experimental and seems to conflict with `prange`, I recommend that we keep computation of the Gini coefficient as a function with `@njit(parallel=True)` as a decorator.

I can see your motivation for wrapping the functionality in a class, since you propose additional methods, but the cost of preserving these operations as individual functions is not high.

mmcky:
thanks @cdagnino this is looking good. I agree with @jstac comment re `jitclass`. 

In other code I often write `jit` functions that do a lot of heavy lifting which makes them accessible (through import) at the function level, and when there is a use case I often use those fast functions to construct a convenience class that holds data and methods and in turn make calls the collection of general functions.

jstac:
@cdagnino Thanks for your help to date.  Should we leave this to you or do you not have time?

cdagnino:
@jstac @mmcky  Apologies for the (really) long silence. I've pushed the changes you suggested. Let me know if I should do a pull request now

quantecon/inequality.py
```diff
@@ -0,0 +1,86 @@
...
+    Implements the Gini inequality index
+
+    Parameters
+    -----------
+    y : array_like(float)
+        Array of income/wealth for each individual. Ordered or unordered is fine
+
+    Returns
+    -------
+
```
rebekahanne: Maybe consider creating a variable called "gini" or something similar and explicitly including it as a variable that's returned in the documentation.

quantecon/tests/test_inequality.py
```diff
@@ -0,0 +1,66 @@
...
+    y[4] = 100000
+    pop_cum, income_cum = lorenz_curve(y)
+    expected_income_cum = np.repeat(0., n + 1)
+    expected_income_cum[-1] = 1.
+    assert_allclose(expected_income_cum, income_cum, atol=1e-4)
+
+
+def test_gini_coeff():
+    """
+    Tests how the `Gini` object calculates the Gini coefficient
```
rebekahanne: Small note, since you now have functions rather than a class (to allow for jitting), perhaps the documentation should refer to the function, not the object

quantecon/inequality.py
cdagnino: Agreed!

quantecon/tests/test_inequality.py
cdagnino: Yep! I forgot to change that. Will change that too

mmcky:
thanks @cdagnino and @rebekahanne -- look forward to merging this.

cdagnino:
@mmcky @rebekahanne I pushed the suggested changes to the docs. Let me know if anything else is missing :-D

mmcky:
thanks @cdagnino and @rebekahanne. I have tagged this as ready and it will be merged later today.

PR #424, opened 2018-07-25
Add bisection and brent's method for root finding

### LINKED ISSUE #422: Add robust jitted root finding routine
I would like to have a jitted robust root finding method such as brent or bisection, following on from PR #417.  Bisection at least would be easy.

spvdchachan:
Hi @jstac and @mmcky 

I would like to resolve #422 in this PR.

I've added two new robust root finding methods `bisect` and `brentq` which are based on Scipy's version.
These procedures were written in C originally and I rewrote them in Python to jit compile.

### DIFF
--- a/quantecon/optimize/__init__.py
+++ b/quantecon/optimize/__init__.py
@@ -3,4 +3,4 @@ Initialization of the optimize subpackage
 """
 
 from .scalar_maximization import brent_max
-from .root_finding import newton, newton_halley, newton_secant
+from .root_finding import *
--- a/quantecon/optimize/__init__.py
+++ b/quantecon/optimize/__init__.py
@@ -3,4 +3,4 @@ Initialization of the optimize subpackage
 """
 
 from .scalar_maximization import brent_max
-from .root_finding import *
+from .root_finding import newton, newton_halley, newton_secant, bisect, brentq
--- a/quantecon/optimize/root_finding.py
+++ b/quantecon/optimize/root_finding.py
@@ -1,5 +1,5 @@
 import numpy as np
-from numba import jit, njit
+from numba import njit
 from collections import namedtuple
 
 __all__ = ['newton', 'newton_halley', 'newton_secant', 'bisect', 'brentq']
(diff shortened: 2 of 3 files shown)


quantecon/optimize/__init__.py
```diff
@@ -3,4 +3,4 @@
 """
 
 from .scalar_maximization import brent_max
-from .root_finding import newton, newton_halley, newton_secant
+from .root_finding import *
```
mmcky: hi @spvdchachan I would rather list the items here rather than using `*`. That way if someone adds a support function that isn't prepended with `_` it won't get promoted through `import`.

mmcky: actually I just realised you have used ``__all__`` in ``root_finding.py``. Please ignore.

spvdchachan:
@mmcky  I've fixed `import` to list items as requested.

quantecon/optimize/root_finding.py
```diff
@@ -2,26 +2,31 @@
 from numba import jit, njit
```
oyamad: `jit` seems not to be used in this file.

mmcky:
thanks @spvdchachan this is looking great.

I agree with @oyamad it would be nice to add the `optional` default values to the documentation. This is often done with ``, optional(rtol = {{ value }})`` or adding it to the description. I see that many of these are set at the top of the file, i.e. using `_rtol`. Is there a way to fill the documentation from the set variable?

spvdchachan:
@oyamad @mmcky I agree with adding default values to the documentation. Hence, I've added them to all functions (including other root finding procedures to be consistent).

I also fixed issues with trailing whitespaces to follow pep8 (checked through `pycodestyle`).

With regards to using variables inside docstrings, one option is to use a decorator as discussed in https://stackoverflow.com/questions/10307696/how-to-put-a-variable-into-python-docstring.

mmcky:
thanks for the updates @spvdchachan. I am happy with values for now. 

that link is interesting - I think the decorator approach would be the best but there is not need to implement this before merging. I'll think a bit more about if we want to pursue filling the docstrings from variables for default values.

PR #427, opened 2018-08-18
update to new rtd requirements spec

### LINKED ISSUE #423: update RTD
Read the Docs has recently started failing. 

It looks like they have changed the way requirements are specified

https://docs.readthedocs.io/en/latest/yaml-config.html#requirements-file

mmcky:
Adjusts `readthedocs.yml` to latest rtd documentation

fixes #423


mmcky:
moved over to use `pip-requirements.txt` as docs seem to render the same way and it is one less requirements file we need to maintain.

PR #431, opened 2018-09-14
DOC: Fix the doc of `root_finding.py` to display nicely

QBatista:
Fixes the documentation of `newton`, `newton_halley` and `newton_secant`.


mmcky:
thanks @QBatista

PR #433, opened 2018-09-17
ENH: Add AS algorithm.

shizejin:
A Python version implementation of AS (Abreu and Sannikov 2014) algorithm. (Julia version is QuantEcon/Games.jl#65).

Comparing to the Julia version, the running time is significantly reduced here by using `scipy.spatial.ConvexHull`. On my computer, the example from AS runs for 30 ms on average, which is even quicker than the original Java implementation (usually runs for 50 ms, sometimes more than 100 ms). (Downloadable from [here](http://econtheory.org/supp/1302/supplement.zip))

A [demonstration](https://nbviewer.jupyter.org/github/shizejin/Notebooks/blob/master/AS%20algorithm-Python.ipynb) shows the AS example and Prisoner's dilemma example.

Detailed docstring needed to be added. Will see if it's possible to improve the efficiency further.

### DIFF
--- a/quantecon/game_theory/repeated_game.py
+++ b/quantecon/game_theory/repeated_game.py
@@ -121,7 +121,7 @@ def _equilibrium_payoffs_abreu_sannikov(rpg, tol=1e-12, max_iter=500,
         msg = "this algorithm only applies to repeated two-player games."
         raise NotImplementedError(msg)
 
-    best_dev_gains = _best_dev_gains(sg, delta)
+    best_dev_gains = _best_dev_gains(rpg)
     C = np.empty((4, 2))
     IC = np.empty(2)
     action_profile_payoff = np.empty(2)
@@ -176,39 +176,31 @@ def _equilibrium_payoffs_abreu_sannikov(rpg, tol=1e-12, max_iter=500,
     return hull
 
 
-def _best_dev_gains(sg, delta):
+def _best_dev_gains(rpg):
     """
     Calculate the normalized payoff gains from deviating from the current
     action to the best response for each player.
 
     Parameters
     ----------
-    sg : NormalFormGame
-        The stage game.
-
-    delta : scalar(float)
-        The common discount rate at which all players discount the future.
+    rpg : RepeatedGame
+        Two player repeated game.
 
     Returns
     -------
-    best_dev_gains0 : ndarray(float, ndim=2)
+    best_dev_gains : tuple(ndarray(float, ndim=2))
         The normalized best deviation payoff gain arrays.
-        best_dev_gains[0][a0, a1] is normalized payoff gain
-        player 0 can get if originally players are choosing
-        a0 and a1, and player 0 deviates to the best response action.
-
-    best_dev_gains1 : ndarray(float, ndim=2)
-        The normalized best deviation payoff gain arrays.
-        best_dev_gains[1][a1, a0] is normalized payoff gain
-        player 1 can get if originally players are choosing
-        a1 and a0, and player 1 deviates to the best response action.
+        best_dev_gains[i][ai, a-i] is normalized payoff gain
+        player i can get if originally players are choosing
+        ai and a-i, and player i deviates to the best response action.
     """
-    best_dev_gains0 = (1-delta)/delta * \
-        (np.max(sg.payoff_arrays[0], 0) - sg.payoff_arrays[0])
-    best_dev_gains1 = (1-delta)/delta * \
-        (np.max(sg.payoff_arrays[1], 0) - sg.payoff_arrays[1])
+    sg, delta = rpg.sg, rpg.delta
 
-    return best_dev_gains0, best_dev_gains1
+    best_dev_gains = ((1-delta)/delta *
+                      (np.max(sg.payoff_arrays[i], 0) - sg.payoff_arrays[i])
+                      for i in range(2))
+
+    return tuple(best_dev_gains)
 
 
 @njit
--- a/quantecon/game_theory/repeated_game.py
+++ b/quantecon/game_theory/repeated_game.py
@@ -5,7 +5,7 @@ Tools for repeated game.
 
 import numpy as np
 from scipy.spatial import ConvexHull
-from numba import jit, njit
+from numba import njit
 
 
 class RepeatedGame:
@@ -15,11 +15,24 @@ class RepeatedGame:
     Parameters
     ----------
     stage_game : NormalFormGame
-                 The stage game used to create the repeated game.
+        The stage game used to create the repeated game.
 
     delta : scalar(float)
-            The common discount rate at which all players discount the future.
+        The common discount rate at which all players discount the future.
 
+    Attributes
+    ----------
+    sg : NormalFormGame
+        The stage game. See Parameters.
+
+    delta : scalar(float)
+        See Parameters.
+
+    N : scalar(int)
+        The number of players.
+
+    nums_actions : tuple(int)
+        Tuple of the numbers of actions, one for each player.
     """
     def __init__(self, stage_game, delta):
         self.sg = stage_game
--- a/quantecon/game_theory/repeated_game.py
+++ b/quantecon/game_theory/repeated_game.py
@@ -39,13 +39,13 @@ class RepeatedGame:
             Two player repeated game.
 
         tol : scalar(float), optional(default=1e-12)
-            Tolerance for convergence checkinsg.
+            Tolerance for convergence checking.
 
         max_iter : scalar(int), optional(default=500)
             Maximum number of iterations.
 
         u_init : ndarray(float, ndim=1), optional(default=np.zeros(2))
-            The initial threat points.
+            The initial guess of threat points.
 
         Returns
         -------
@@ -57,8 +57,8 @@ class RepeatedGame:
         C = np.empty((4, 2))
         IC = np.empty(2)
         action_profile_payoff = np.empty(2)
-        # array for checking if payoff is inside the polytope or not
-        # the last entry is set to be 1
+        # auxiliary array for checking if payoff is inside the convex hull
+        # first two entries for payoff point, and the last entry is 1.
         extended_payoff = np.ones(3)
         # array to store new points of C in each intersection
         # at most 4 new points will be generated
@@ -131,9 +131,9 @@ def _best_dev_gains(sg, delta):
 
     best_dev_gains1 : ndarray(float, ndim=2)
         The normalized best deviation payoff gain arrays.
-        best_dev_gains[1][a0, a1] is normalized payoff gain
+        best_dev_gains[1][a1, a0] is normalized payoff gain
         player 1 can get if originally players are choosing
-        a0 and a1, and player 1 deviates to the best response action.
+        a1 and a0, and player 1 deviates to the best response action.
     """
     best_dev_gains0 = (1-delta)/delta * \
         (np.max(sg.payoff_arrays[0], 0) - sg.payoff_arrays[0])
@@ -164,8 +164,8 @@ def R(delta, nums_actions, payoff_arrays, best_dev_gains, points,
 
     best_dev_gains : tuple(ndarray(float, ndim=2))
         Tuple of the normalized best deviation payoff gain arrays.
-        best_dev_gains[i][a0, a1] is payoff gain player i
-        can get if originally players are choosing a0 and a1,
+        best_dev_gains[i][ai, a-i] is payoff gain player i
+        can get if originally players are choosing ai and a-i,
         and player i deviates to the best response action.
 
     points : ndarray(float, ndim=2)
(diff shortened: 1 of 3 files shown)


shizejin:
Thanks @cc7768 ! I will add detailed doc string these days so that it makes easier for you to review.

About the Julia version, yes it has been suspending for a really long time... Actually, after writing up this Python version, I realized I made some stupid bugs in the Julia version. After you review this Python version I will go back there and modify it.

oyamad:
@shizejin Looks great!

Some initial comments:

* Consider setting up a class `RepeatedGame` and implementing the function as a method of that class; refer to #347.
* Look for a more informative name than `AS`. I don't have a good idea though: `RepeatedGame.get_equilibrium_payoff_set(method='abreu_sannikov')` would be too long...

shizejin:
I have added `RepeatedGame` class and use `AS` as a function of the class. 

[Here](https://nbviewer.jupyter.org/github/shizejin/Notebooks/blob/master/AS%20algorithm-Python.ipynb) is the newest demonstration, with a Cournot Duopoly Game example added.

To do:

- [ ] Add docstrings for the internal functions.

- [ ] Think about a more informative name for `AS`.

shizejin:
This is ready for review.

Before the merge, we may want to decide whether to use `RepeatedGame.get_equilibrium_payoff_set(method='abreu_sannikov')` as suggested by @oyamad, or `RepeatedGame.abreu_sannikov()` (or just `RepeatedGame.AS()` which I am using now. `AS` is the abbreviation of the algorithm they proposed in the paper.)

The former one is favorable in the sense that it is easy to add `outerapproximation` algorithm in the future, but I am also concerned about it as `outerapproximation` and `AS` take very different arguments, and I am not sure if it is possible to use one `get_equilibrium_payoff_set()` for both of them.

mmcky:
@oyamad would you have time to review this PR?

mmcky:
- [ ] @mmcky to update documentation to include `repeated_games` once ready to be merged.

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,395 @@
...
+                    break
+
+            # update threat points
+            update_u(u, W_new[:n_new_pt])
+
+        hull = ConvexHull(W_new[:n_new_pt])
+
+        return hull
+
+@jit()
```
oyamad: Remove this decorator, which is of no use?

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,395 @@
...
+        max_iter : scalar(int), optional(default=500)
+            Maximum number of iterations.
+
+        u_init : ndarray(float, ndim=1), optional(default=np.zeros(2))
+            The initial guess of threat points.
+
+        Returns
+        -------
+        hull : scipy.spatial.ConvexHull
+            The convex hull of feasible payoff pairs.
```
oyamad: Isn't it the convex hull of *equilibrium* payoff pairs?

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,411 @@
...
+            The common discount rate at which all players discount the future.
+
+    """
+    def __init__(self, stage_game, delta):
+        self.sg = stage_game
+        self.delta = delta
+        self.N = stage_game.N
+        self.nums_actions = stage_game.nums_actions
+
+    def AS(self, tol=1e-12, max_iter=500, u_init=np.zeros(2)):
```
mmcky: @oyamad @shizejin what about using ``compute_as()`` as a method name?

oyamad:
* An alternative is to implement this method as a function that takes `RepeatedGame` as (one of) the argument like `lemke_howson` (possibly in a separate file) and call it just `abreu_sannikov`.

* Later, we may add a method, `RepeatedGame.get_equilibrium_payoff_set(method='abreu_sannikov')` or `RepeatedGame.equilibrium_payoffs(method='abreu_sannikov')`, allowing also an alias `method='AS'`.

* Note that passing different options for different methods is not a big problem. It is usual if you have a unified interface that unifies several different methods; see for example the `options` argument in [scipy.optimize.minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html).

shizejin:
- I think `RepeatedGame.equilibrium_payoffs(method='abreu_sannikov')` would be a good idea. It is clear and not too long.

- `scipy.optimize.minimize` takes `fun` and `x0` as key parameters for all methods, while there is no such common parameters for our case (for `abreu_sannikov` and `outerapproximation`). In that sense, we need to put all the parameters in `options`, which is a little bit strange to me. But I guess it's fine.

quantecon/game_theory/repeated_game.py
mmcky: @oyamad I think you're right.

shizejin:
A unified interface `RepeatedGame.equilibrium_payoffs()` has been added, following the style of `scipy.optimize.minimize`.

The usage would be as following now:

1. Abreu and Sannikov 2014 example

```python
>>> p1 = gt.Player([[16, 3, 0], [21, 10, -1], [9, 5, -5]])
>>> p2 = gt.Player([[9, 1, 0], [13, 4, -4], [3, 0, -15]])
>>> sg = gt.NormalFormGame([p1, p2])
>>> rpg = gt.RepeatedGame(sg, 0.3)
>>> hull = rpg.equilibrium_payoffs()
>>> hull.points[hull.vertices]
array([[7.33770472e+00, 1.09826253e+01],
       [1.12568240e+00, 2.80000000e+00],
       [7.33770472e+00, 4.44089210e-16],
       [7.86308964e+00, 4.44089210e-16],
       [1.97917977e+01, 2.80000000e+00],
       [1.55630896e+01, 9.10000000e+00]])
```

2. Prisoner Dilemma example
```python
>>> pd_payoff = [[9.0, 1.0], [10.0, 3.0]]
>>> A = gt.Player(pd_payoff)
>>> B = gt.Player(pd_payoff)
>>> sg = gt.NormalFormGame((A, B))
>>> rpg = gt.RepeatedGame(sg, 0.9)
>>> hull = rpg.equilibrium_payoffs(options={'u_init': np.array([3, 3])})
>>> hull.points[hull.vertices]
array([[3.  , 3.  ],
       [9.75, 3.  ],
       [9.  , 9.  ],
       [3.  , 9.75]])
```

Currently, only `abreu_sannikov` method is supported. Therefore, if we pass `method='outerapproximation'`, `NotImplementedError` will be raised.

```python
>>> hull = rpg.equilibrium_payoffs(method='abreu_sannikov', options={'u_init': np.array([3, 3])})
>>> hull = rpg.equilibrium_payoffs(method='AS', options={'u_init': np.array([3, 3])})
>>> hull = rpg.equilibrium_payoffs(method='outerapproximation', options={'u_init': np.array([3, 3])})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/hyde/QuantEcon.py/quantecon/game_theory/repeated_game.py", line 69, in equilibrium_payoffs
    raise NotImplementedError(msg)
NotImplementedError: method outerapproximation not supported.
```

It will be very easy for us to unify `outerapproximation` into this interface in the future.

Would this way of implementation be better? @oyamad @mmcky

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
+"""
+Tools for repeated game.
+
+"""
+
+import numpy as np
+from scipy.spatial import ConvexHull
+from numba import jit, njit
```
oyamad: Remove `jit`.

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
...
+
+    Parameters
+    ----------
+    stage_game : NormalFormGame
+                 The stage game used to create the repeated game.
+
+    delta : scalar(float)
+            The common discount rate at which all players discount the future.
+
+    """
```
oyamad: Add docstring for attributes.

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
...
+            options = {}
+
+        if method in ('abreu_sannikov', 'AS'):
+            return _equilibrium_payoffs_abreu_sannikov(self, **options)
+        else:
+            msg = f"method {method} not supported."
+            raise NotImplementedError(msg)
+
+
+def _equilibrium_payoffs_abreu_sannikov(rpg, tol=1e-12, max_iter=500,
```
oyamad: Does it make sense to export this function, under the name, say, `abreu_sannikov` (this is what I meant in my previous comment)? In the current version, no documentation is exposed for the AS algorithm implementation.

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
...
+
+        # update threat points
+        _update_u(u, W_new[:n_new_pt])
+
+    hull = ConvexHull(W_new[:n_new_pt])
+
+    return hull
+
+
+def _best_dev_gains(sg, delta):
```
oyamad: What's wrong with passing a `RepeatedGame` (which is a container of `sg` and `delta`) as an argument to this function?

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
...
+    best_dev_gains1 : ndarray(float, ndim=2)
+        The normalized best deviation payoff gain arrays.
+        best_dev_gains[1][a1, a0] is normalized payoff gain
+        player 1 can get if originally players are choosing
+        a1 and a0, and player 1 deviates to the best response action.
+    """
+    best_dev_gains0 = (1-delta)/delta * \
+        (np.max(sg.payoff_arrays[0], 0) - sg.payoff_arrays[0])
+    best_dev_gains1 = (1-delta)/delta * \
+        (np.max(sg.payoff_arrays[1], 0) - sg.payoff_arrays[1])
```
oyamad: Tuple comprehension seems to be a better approach here.

quantecon/game_theory/repeated_game.py
```diff
@@ -0,0 +1,454 @@
...
+            method = 'abreu_sannikov'
+
+        if options is None:
+            options = {}
+
+        if method in ('abreu_sannikov', 'AS'):
+            return _equilibrium_payoffs_abreu_sannikov(self, **options)
+        else:
+            msg = f"method {method} not supported."
+            raise NotImplementedError(msg)
```
oyamad: I would expect a `ValueError` if I pass `method='something_nonsense'`.

quantecon/game_theory/repeated_game.py
shizejin: Would it be a little bit redundant if we both export `abreu_sannikov` and `RepeatedGame.equilibrium_payoffs(method='abreu_sannikov')`?

What `scipy.optimize.minimize` does is to make documentation page for each internal function corresponding to each method (e.g. [minimize(method='COBYLA')](https://github.com/scipy/scipy/blob/8dba340293fe20e62e173bdf2c10ae208286692f/doc/source/optimize.minimize-cobyla.rst)) and then add a reference link in the documentation of `scipy.optimize.minimize` (e.g. [here](https://github.com/scipy/scipy/blob/8dba340293fe20e62e173bdf2c10ae208286692f/scipy/optimize/_minimize.py#L74)). Do you think we can follow this way?

quantecon/game_theory/repeated_game.py
shizejin: That's True. I shall change this exception.

Considering that 'outerapproximation' method is going to be added in the future, shall we set a separate `NotImplementedError` exception for it?

mmcky:
@oyamad this PR is looking close. Should I wait to release a new version through `PyPI` after we merge this PR?

mmcky:
thanks @oyamad and @shizejin for this contribution. I will merge now and make a new PyPI release this afternoon.

quantecon/game_theory/tests/test_repeated_game.py
```diff
@@ -0,0 +1,57 @@
...
+             'delta': 0.9,
+             'vertices': vertices,
+             'u': np.array([3., 3.])}
+        self.game_dicts.append(d)
+
+    def test_abreu_sannikov(self):
+        for d in self.game_dicts:
+            rpg = RepeatedGame(d['sg'], d['delta'])
+            for method in ('abreu_sannikov', 'AS'):
+                hull = rpg.equilibrium_payoffs(options={'u_init': d['u']})
```
natashawatkins: @shizejin just wondering, where is `method` used here?

shizejin: @natashawatkins Thanks for reviewing. I should have added `method='abreu_sannikov'` and `method='AS'` here. The test passed because currently only "AS" method is supported and is set to be default.

I will fix this in minutes.

natashawatkins: No problem, I'm just writing a news item about the latest release

Issue #411, opened 2018-05-25
Kalman filter docstring mistake?


mirca:
Hi,

I believe there is a mistake in the docstring of ``x_hat``. Where it says ``mean x_hat and covariance
matrix Sigma of the prior/predictive density``, it should say only ``mean of the prior/predictive density.`` Let me know if that's legit.
```
    x_hat : scalar(float) or array_like(float), optional(default=None)
        An n x 1 array representing the mean x_hat and covariance
        matrix Sigma of the prior/predictive density.  Set to zero if
        not supplied.
```
https://github.com/QuantEcon/QuantEcon.py/blob/35f41b118b26c12e104d412ff33a252e903c0bfd/quantecon/kalman.py#L41

mmcky:
thanks @mirca for the report. I will look into this when I'm back in the office on Tuesday.

PR #438, opened 2018-10-02
Fix docstring of x_hat in Kalman class

mirca:
This PR fixes issue #411 which points out a minor inconsistency on the docstring of ``x_hat``, the mean vector of ``Kalman``.


mmcky:
thanks @mirca for the update.

PR #439, opened 2018-10-04
Fix issue#367

### LINKED ISSUE #367: kalman.py: Add `method` option

Yuya-Furusawa:
Add `method` option in Kalman class.

Close #367

### DIFF
--- a/quantecon/kalman.py
+++ b/quantecon/kalman.py
@@ -240,7 +240,7 @@ class Kalman:
         self.prior_to_filtered(y)
         self.filtered_to_forecast()
 
-    def stationary_values(self,method='doubling'):
+    def stationary_values(self, method='doubling'):
         """
         Computes the limit of :math:`\Sigma_t` as t goes to infinity by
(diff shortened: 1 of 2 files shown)


oyamad:
@Yuya-Furusawa Thanks.

I should have noticed earlier, this is a bit tricky: You can't pass an additional argument to a method with `@property`. Let's discuss later today.

quantecon/kalman.py
```diff
@@ -240,12 +240,19 @@ def update(self, y):
         self.prior_to_filtered(y)
         self.filtered_to_forecast()
 
-    def stationary_values(self):
+    def stationary_values(self,method='doubling'):
```
oyamad: missing whitespace after ','

PR #441, opened 2018-10-16
FEAT: Add the Nelder-Mead algorithm

### LINKED ISSUE #419: Add multivariate optimization to quantecon.optimize
It would be nice to have at least one multivariate maximizer, akin to our Brent based scalar routine in

https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/optimize/scalar_maximization.py

This is a topic I know very little about but I'm thinking of a local routine such as this one:

https://docs.scipy.org/doc/scipy/reference/optimize.minimize-neldermead.html

Or perhaps this one, which seems to have the option to supply a bounding box:

https://docs.scipy.org/doc/scipy/reference/optimize.minimize-lbfgsb.html

One problem is that some of the SciPy code is wrapped C / Fortran.

@cc7768 I seem to remember to you had some Nedler-Mead code.  Might you be willing to share?  Would this be a good starting point?
@sglyon @albop @oyamad Any thoughts?

CC @chrishyland @spvdchachan

I should note that @chrishyland and @spvdchachan will do the leg work on this so from others I'm just hoping for a few pointers.

QBatista:
Adds a jitted implementation of the Nelder-Mead algorithm for multivariate optimization. 

The two main references are [Lagarias et. al (1998)](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.120.6062&rep=rep1&type=pdf) for the algorithm and [Singer and Singer (2004)](https://onlinelibrary.wiley.com/doi/abs/10.1002/anac.200410015) for efficiency. 

It might be a good idea to have a common interface for `scalar_maximization.py` and this multivariate version. Please let me know if you have any suggestions for setting up this interface and choosing 
 names.

### Performance
```python
import quantecon as qe
from numba import njit
import numpy as np

@njit
def rosenbrock(x):
    return -(100 * (x[1] - x[0] ** 2) ** 2 + (1 - x[0])**2)

x0 = np.array([-2, 1]) 
```
```python
qe.optimize.maximize(rosenbrock, x0, tol_x=1e-20)
%timeit qe.optimize.maximize(rosenbrock, x0, tol_x=1e-20)
```
```
106 µs ± 6.83 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
```python
from scipy.optimize import minimize

%timeit minimize(rosenbrock, x0, method='Nelder-Mead')
```
```
6.23 ms ± 308 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Close #419

### DIFF
--- a/quantecon/optimize/__init__.py
+++ b/quantecon/optimize/__init__.py
@@ -3,5 +3,5 @@ Initialization of the optimize subpackage
 """
 
 from .scalar_maximization import brent_max
-from .multivar_maximization import maximize, nelder_mead_algorithm
+from .nelder_mead import nelder_mead
 from .root_finding import newton, newton_halley, newton_secant, bisect, brentq
--- a/quantecon/optimize/__init__.py
+++ b/quantecon/optimize/__init__.py
@@ -3,4 +3,5 @@ Initialization of the optimize subpackage
 """
 
 from .scalar_maximization import brent_max
+from .multivar_maximization import maximize, nelder_mead_algorithm
 from .root_finding import newton, newton_halley, newton_secant, bisect, brentq
--- a/quantecon/optimize/multivar_maximization.py
+++ b/quantecon/optimize/multivar_maximization.py
@@ -10,7 +10,7 @@ from collections import namedtuple
 results = namedtuple('results', 'x fun success nit final_simplex')
 
 
-#@njit
+@njit
 def maximize(fun, x0, bounds=np.array([[], []]).T, args=(), tol_f=1e-10,
              tol_x=1e-10, max_iter=1000):
     """
@@ -107,7 +107,7 @@ def maximize(fun, x0, bounds=np.array([[], []]).T, args=(), tol_f=1e-10,
     return results
 
 
-#@njit
+@njit
 def nelder_mead_algorithm(fun, vertices, bounds=np.array([[], []]).T,
                           args=(), ρ=1., χ=2., γ=0.5, σ=0.5, tol_f=1e-8,
                           tol_x=1e-8, max_iter=1000):
@@ -274,7 +274,7 @@ def nelder_mead_algorithm(fun, vertices, bounds=np.array([[], []]).T,
                    vertices)
 
 
-#@njit
+@njit
 def _initialize_simplex(x0, bounds):
     """
     Generates an initial simplex for the Nelder-Mead method.
@@ -314,7 +314,7 @@ def _initialize_simplex(x0, bounds):
     return vertices
 
 
-#@njit
+@njit
 def _check_params(ρ, χ, γ, σ, bounds, n):
     """
     Checks whether the parameters for the Nelder-Mead algorithm are valid.
@@ -357,7 +357,7 @@ def _check_params(ρ, χ, γ, σ, bounds, n):
         raise ValueError("Lower bounds must be greater than upper bounds.")
 
 
-#@njit
+@njit
 def _check_bounds(x, bounds):
     """
     Checks whether `x` is within `bounds`.
@@ -383,7 +383,7 @@ def _check_bounds(x, bounds):
                 (x <= np.atleast_2d(bounds)[:, 1]).all())
 
 
-#@njit
+@njit
 def _neg_bounded_fun(fun, bounds, x, args=()):
     """
     Wrapper for bounding and taking the negative of `fun` for the
--- a/quantecon/optimize/multivar_maximization.py
+++ b/quantecon/optimize/multivar_maximization.py
@@ -1,0 +1,419 @@
+"""
+Implements the Nelder-Mead algorithm for maximizing a multivariate function
+
+"""
+
+import numpy as np
+from numba import njit
+from collections import namedtuple
+
+results = namedtuple('results', 'x fun success nit final_simplex')
+
+
+#@njit
+def maximize(fun, x0, bounds=np.array([[], []]).T, args=(), tol_f=1e-10,
+             tol_x=1e-10, max_iter=1000):
+    """
+    .. highlight:: none
+
+    Maximize a scalar function with multiple variables using the Nelder-Mead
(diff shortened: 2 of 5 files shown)


natashawatkins:
I agree with having a common interface which can be used for either scalar or multivariate

Maybe we could do something similar to scipy and have both a common interface to the function and individual functions, ie. `nelder_mead` and `brentq`, and a function `maximize` with a method argument

natashawatkins:
But if we don't decide for go for a common interface for now, I think we should call this function something more like `nelder_mead` to differentiate from `brent_max`

natashawatkins:
I've tried adding this to the lecture but don't seem to get convergence.. here is a look at the output from this function (stored in `solver`) vs. the 'brute-force' method we're currently using

screen shot 2018-10-19 at 11 48 23 am

Here is the piece of code just in case you spot something wrong with it:

image

Also why is the bounds parameter set to have a shape of two by default? Is this needed?

Thanks

QBatista:
The bounds parameter is set this way to get Numba to work without using `generated_jit`. Would it be possible for you to upload the notebook so that I can have a look at it?

QBatista:
It looks like the problem is coming from the algorithm. Both my version and Scipy often seem to cycle. I can sometimes get convergence by tweaking the parameters depending on the  `JVWorker` instance, but this occurs quite rarely. Most other methods in Scipy seem to fail, but TNC sometimes work. However, this also doesn't seem to be very robust -- changing the initial guess causes failure.

natashawatkins:
So in the live lecture at the moment, it's using the COBYLA method - do you think we should implement this one in the QuantEcon library?

QBatista:
Actually, after checking the live lecture, the problem seems to be that the constraints are missing. By passing `bounds = np.array([[0., 1.], [0., 1.]])` as in the live lecture, it seems to be converging consistently. The graph is slightly different though. I tried a naive way of encoding `s + ϕ <= 1` by changing the objective function to return `-np.inf` if the constraint is not satisfied, but the algorithm seems to eventually get stuck.

chrishyland:
Hey @natashawatkins , 
Just thought to jump in but in a previous iteration I had of going about implementing the bounded constraints just involved shifting a point to its closest boundary if a point is outside the feasible region. This is touched on in [here](https://github.com/QuantEcon/QuantEcon.py/issues/419) if you scroll towards the end.

chrishyland:
Hi @QBatista ,
I believe that in the other link that the box constraints implementation implementation would be sufficient for most problems and I recall seeing an implementation somewhere (will have to look again) in a paper that uses something like this. I’ve struggled as well to find a linear inequality constraint solution that you mentioned as it appears that the Nelder-Mead algorithm is an unconstrained optimisation problem, so I’m not sure if adding such constraints is possible unless you alter the algorithm? However, I recall that @jstac saying that just having box constraints is sufficient for now for Quantecon’s needs. Sorry I couldn’t be of more help but I’d be happy to discuss things with you if you had any other ideas on what to do.

QBatista:
@jstac @natashawatkins Having a common interface for the Nelder-Mead algorithm and `brent_max` would require choosing a common set of required arguments. However, the Nelder-Mead algorithm requires an initial guess of the solution `x0` while `brent_max` doesn't. 

One option would be to follow Scipy, which uses two different interfaces, one dedicated to function minimization with only one variable (called `minimize_scalar`) and a more general one which handles functions with both one or more variables (called `minimize`). The algorithms behind both interfaces are different: brent is used in `minimize_scalar`, while Nelder-Mead is used in `minimize`.

Another option would be to make the initial guess argument optional and raise an error message if the method used requires an initial guess. Scipy uses a similar idea for its L-BFGS-B method in the `minimize` interface, where the Jacobian of the function is a keyword argument, but an error message is raised if the Jacobian is not passed because this particular method requires it.

In terms of maintenance, the first option would have a more straightforward implementation, especially since we want the interface to be jitted, and therefore should be easier to maintain. 

@oyamad Please let me know if you have any suggestions for designing this interface.

QBatista:
@jstac In Scipy, the methods that do not require an initial guess are methods which can be exclusively applied to functions with only one variable, and therefore, they all fall in `scalar_minimization`. Meanwhile, the more general methods that can be used for functions with one or more variables all require an initial guess, and are accessed through `minimize`. Therefore, the `minimize` interface requires an initial guess to be passed, while `minimize_scalar` has no option for passing an initial guess. This means that a method which requires an initial guess but can be applied exclusively to functions with only one variable would not fit into any of these two interfaces.

jstac:
Thanks for the explanation.  I guess something analogous would be OK for us.

natashawatkins:
Should we just stick to having separate functions for different algorithms? What was the issue regarding generated jit again?

QBatista:
Yes, that's fine too. Just one thing to note: we currently also have root finding methods in `qe.optimize` and these would not fit in `qe.maximize`. For example, `brentq` is a root finding method while `brent` is a maximization method. SciPy puts both root finding methods and optimization methods in `optimize`, so we could either put all of them in `qe.optimize` along with maximization methods, or have both `qe.maximize` and `qe.root_finding` (or something similar).

natashawatkins:
sorry I meant `qe.optimize.brentq`

jstac:
I'm in favor of @natashawatkins's suggestion.

QBatista:
Sounds good -- I'll make the changes soon.

QBatista:
This is now ready for review.

jstac:
Cleverly written, well documented, excellent tests.  Great job @QBatista.

mmcky:
thanks @QBatista this is looking good. I will attach a `ready` tag and merge tomorrow.

mmcky:
Once this is merged -- I will issue a new release via PyPI

natashawatkins:
@mmcky are you releasing a new version of the library? I might use this in a lecture

PR #443, opened 2018-11-04
FIX: Disallow Player with 0 actions

oyamad:


mmcky:
thanks @oyamad I will merge this today.

PR #446, opened 2018-11-27
FIX: test `method` keyword of `RepeatedGame.equilibrium_payoffs()`.

shizejin:
Fix the bug that `method` keyword is not tested, as mentioned [here](https://github.com/QuantEcon/QuantEcon.py/pull/433#discussion_r236513038).


mmcky:
thanks @shizejin

PR #447, opened 2018-12-07
Add Shorrocks inequality index

natashawatkins:
Adds the Shorrocks inequality index as described in this paper on page 13:      https://www.econ.nyu.edu/user/bisina/RevisionAugust.pdf

### DIFF
--- a/quantecon/__init__.py
+++ b/quantecon/__init__.py
@@ -23,7 +23,7 @@ from .estspec import smooth, periodogram, ar_periodogram
 # from .game_theory import <objects-here> 							#Place Holder if we wish to promote any general objects to the qe namespace.
 from .graph_tools import DiGraph, random_tournament_graph
 from .gridtools import cartesian, mlinspace, simplex_grid, simplex_index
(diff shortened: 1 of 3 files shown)


mmcky:
- [ ] update ``docs`` to include

natashawatkins:
I was trying to follow the instructions here: https://github.com/QuantEcon/QuantEcon.py/tree/master/docs but wasn't able to compile the docs - was getting an import error

Is my commit sufficient though?

natashawatkins:
Also I think we need to add documentation for the `nelder_mead` algorithm added by @QBatista ?

mmcky:
probably need to add it to the index?
we do - but please don't add that in this PR. We will make a separate PR to add `nelder_mead`

jstac:
Nice work @natashawatkins.  Nice code and tests.

Let's try to remember to get this in a notebook somehow, so we make it known.

@cdagnino FYI, since I seem to remember you were looking at mobility indices.

PR #450, opened 2018-12-10
ENH: Add errors for invalid inputs for `brent_max`

### LINKED ISSUE #442: Function value convergence test for `brent_max`
Without a function value convergence test, `brent_max` can fail on simple functions, for instance:

```python
>>> import quantecon as qe
>>> import numpy as np
>>> from numba import njit
>>> @njit
... def f(x): return 1.
... 
>>> qe.optimize.brent_max(f, -np.inf, np.inf)
(nan, 1.0, (0, 1))
```

Adding a function value convergence test should solve this issue.

QBatista:
Adds error messages for invalid inputs for `brent_max`.

Close #442


mmcky:
thanks @QBatista

PR #454, opened 2018-12-11
Update Documentation

mmcky:
Update the documentation to include

- [x] inequality
- [x] nelder mead (source)
- [x] repeated_game (source)

### DIFF
--- a/quantecon/inequality.py
+++ b/quantecon/inequality.py
@@ -98,7 +98,7 @@ def shorrocks_index(A):
     --------
     Shorrocks index: float
         The Shorrocks mobility index calculated as 
-        :math`s(A) = \frac{m - \sum_j a_{jj}}{m - 1} \in (0, 1)`
+        :math:`s(A) = \frac{m - \sum_j a_{jj}}{m - 1} \in (0, 1)`
 
         An index equal to 0 indicates complete immobility.
 
--- a/quantecon/inequality.py
+++ b/quantecon/inequality.py
@@ -30,7 +30,7 @@ def lorenz_curve(y):
 
     References
     ----------
-    https://en.wikipedia.org/wiki/Lorenz_curve
+    .. [1] https://en.wikipedia.org/wiki/Lorenz_curve
 
     Examples
     --------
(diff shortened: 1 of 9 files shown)


rht:
@mmcky https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/466292136 is the test flake I encountered. It is indeed non-deterministic? The nelder mead code is recent, isn't it?

quantecon/inequality.py
```diff
@@ -106,10 +106,11 @@ def shorrocks_index(A):
         An index equal to 0 indicates complete immobility.
```
mmcky: @natashawatkins could you check this math equation. It seems to be breaking on the `html` build.

mmcky:
the build of this branch is viewable: https://quanteconpy.readthedocs.io/en/update-docs/

mmcky:
thanks @rht. @QBatista can you take a look at the issue with `nelder_mead` test?

quantecon/inequality.py
mmcky: https://quanteconpy.readthedocs.io/en/update-docs/tools/inequality.html

mmcky:
@natashawatkins I tried a few things re: `sharrock` but there are still issues with the build. 

```bash
/home/mmcky/anaconda3/lib/python3.6/site-packages/quantecon-0.4.2-py3.6.egg/quantecon/inequality.py:docstring of quantecon.inequality.shorrocks_index:34: WARNING: Error in "rubric" directive:
no content permitted.

.. rubric:: References

    .. [R23] Benhabib, Bison, and Luo, Wealth distribution and social mobility 
           in the US:  A quantitative approach (2017), https://www.econ.nyu.edu/user/bisina/RevisionAugust.pdf
```
If you can take a look at the RST that would be great.

quantecon/inequality.py
natashawatkins: How can I actually build the docs myself because I was having problems when I followed the instructions?

natashawatkins:
I think the above should have fixed it, but I can't test because docs won't compile

I'm not sure about the math error - it renders perfectly fine in latex

natashawatkins:
I'll try double escaping `\f` since it seems to have a problem with that

natashawatkins:
Actually I just added a raw flag

PR #456, opened 2018-12-11
Travis: Install numpy with OpenBLAS instead of MKL

rht:
By default, conda installs the non-FLOSS version of numpy, which bloats the installation size by 204.4 MB compressed (~600 MB uncompressed). This slows down the cache creation, even though now the `conda install` time has decreased from ~113s to ~11s.

See also: https://github.com/conda-forge/numpy-feedstock/issues/84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,9 +48,11 @@ before_install:
   #- sudo ln -s /run/shm /dev/shm
 
 install:
+  - pip install coveralls coverage flake8
+  # catch flake8 error before anything else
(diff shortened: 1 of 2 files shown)


rht:
https://travis-ci.org/QuantEcon/QuantEcon.py/builds/428771113#L539

rht:
I added the flake8 commit for the purpose of testing how long it takes for the cache to download (~44s, https://travis-ci.org/QuantEcon/QuantEcon.py/jobs/466303295#L443). It looks like the download size hasn't decreased significantly because the binaries from conda-forge are still fat. I suppose the remaining solution is to use pip.

PR #458, opened 2018-12-11
Travis: Move dependency installs with wheels available to pip

rht:
This supersedes #456.


mmcky:
thanks @rht -- this is a nice improvement.

PR #461, opened 2018-12-11
Travis: Disable performance tests

### LINKED ISSUE #109: TEST: Tests for grid_tools take a long time
I just ran the tests on my machine and got the following printout:

```
Ran 247 tests in 73.629s
```

I checked and if I run only [this test](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/tests/test_cartesian.py#L51-L83) in test_cartesian.py I get:

```
$ nosetests quantecon/tests/test_cartesian.py:test_performance_C
.
----------------------------------------------------------------------
Ran 1 test in 23.669s
```

Similar timings for the `test_performance_F` function:

```
$ nosetests quantecon/tests/test_cartesian.py:test_performance_F
.
----------------------------------------------------------------------
Ran 1 test in 24.902s
```

These two tests account for about 2/3 of the total testing time. Is there a way we can not run these tests by default, or at least have their run time not take so long?

We could probably still update the travis script to make sure they are checked by travis.

@albop  any ideas?

rht:
They should be run only from time to time, e.g. between releases, or
when the commit specifically states the perf modification of the
relevant code.

This fixes #109.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,9 @@ install:
   - python setup.py install
 
 script:
-  - nosetests --with-coverage --cover-package=quantecon
+  - nosetests --with-coverage -a "!slow" --cover-package=quantecon
 
 after_success:
   - coveralls
+  # Enable this to occasionally test performance
+  # - nosetests -a "slow"


mmcky:
thanks @rht. 

@oyamad -- QuantEcon.py.benchmark is currently defunct. Never got it running on a stable build platform for consistent results. I'll assign an RA to look into setting this up to support as it would be nice to have.

https://github.com/QuantEcon/QuantEcon.py.benchmark

PR #462, opened 2018-12-13
fix for lorenz documentation

natashawatkins:
Need these arrows for code like rendering


natashawatkins:
Ah I might have to add import statements I guess?

mmcky:
thanks @natashawatkins. I'll merge once CI is finished.

PR #468, opened 2018-12-26
DOC: Update example for `nelder_mead`

QBatista:
Fixes the example in the documentation of `nelder-mead`.


mmcky:
thanks @QBatista

PR #476, opened 2019-03-12
FEAT: Add `AMF_LSS_VAR`

QBatista:
Adds an `AMF_LSS_VAR` class to support the lectures on additive and multiplicative functionals.

### Class design

- I've deconstructed `construct_ss` from the class in the lecture into 6 functions for two reasons: (i) to be able to test the construction of each matrix separately and (ii) to satisfy the [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle).

- The functions `additive_decomp` and `multiplicative_decomp` were recomputing the additive  and multiplicative decomposition every time they were called. I've made them attributes of the class that are computed only once (when the class is initialized).

- I haven't added setter methods for the attributes because ensuring that new values are valid inputs and updating the related variables is delicate.

- I agree with @oyamad that the plotting code should be separated from the class because the purpose of the class is to use the first-order VAR representation of an additive functional to build the LSS representation. Adding visualization methods would give another distinct responsibility to the class. That being said, there are different options for testing the plotting code if you want to include it in the class. One option is to write an [image comparison test](https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test). Another option is to run numerical tests against the data returned by the plotting functions.

### TODO

- [x] Add `loglikelihood_path` 

- [x] Add helper function to compute first order stacked VAR representation from pth order representation

- [x] Add helper function to apply the Blanchard and Quah method

### DIFF
--- a/quantecon/__init__.py
+++ b/quantecon/__init__.py
@@ -18,7 +18,7 @@ from . import random
 from . import optimize
 
 #-Objects-#
-from .amf import AMF_LSS_VAR
+from .amf import AMF
 from .compute_fp import compute_fixed_point
 from .discrete_rv import DiscreteRV
 from .dle import DLE
--- a/quantecon/__init__.py
+++ b/quantecon/__init__.py
@@ -18,8 +18,7 @@ from . import random
 from . import optimize
 
 #-Objects-#
-from .amf import (AMF_LSS_VAR, pth_order_to_stacked_1st_order,
-                  compute_BQ_restricted_B_0)
+from .amf import AMF_LSS_VAR
 from .compute_fp import compute_fixed_point
 from .discrete_rv import DiscreteRV
 from .dle import DLE
--- a/quantecon/__init__.py
+++ b/quantecon/__init__.py
@@ -18,6 +18,8 @@ from . import random
 from . import optimize
 
 #-Objects-#
+from .amf import (AMF_LSS_VAR, pth_order_to_stacked_1st_order,
+                  compute_BQ_restricted_B_0)
 from .compute_fp import compute_fixed_point
 from .discrete_rv import DiscreteRV
 from .dle import DLE
--- a/quantecon/amf.py
+++ b/quantecon/amf.py
@@ -1,0 +1,362 @@
+"""
+A module for working with additive and multiplicative functionals.
+
+"""
+
+import numpy as np
+import scipy.linalg as la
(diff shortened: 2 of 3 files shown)


jstac:
Thanks @QBatista. This is looking good --- I like the class design.  I'm in favor of pulling out the plotting code and embedding it in the lectures, even if this means we end up with similar plotting code in a few different lectures.

mmcky:
@QBatista is this still be worked on? Can we close this PR?

QBatista:
@mmcky Tom would like to move forward with this to (i) support his book with LPH and (ii) support some QuantEcon lectures such as https://python-advanced.quantecon.org/additive_functionals.html#Simulation that currently define an `AMF_LSS_VAR` directly in the lecture. I'll try to wrap up this PR for merging with basic functionalities needed to support the lectures this week.

QBatista:
@mmcky This is ready for review. It contains basic features for supporting the additive and multiplicative functionals lecture. Once the PR is merged and a new version of `QuantEcon.py` released, I can modify the lecture to use this class. The modifications needed to the lecture are minimal.

jstac:
Well written and tested.  I think this is niche but still valuable.  Would it be OK if we ping you in the future if we get PRs on this @QBatista ?  I double there will be many.

QBatista:
@jstac Yes, of course.

oyamad:
One minor comment on class naming:

According to [PEP8](https://www.python.org/dev/peps/pep-0008/#class-names):

> Class names should normally use the CapWords convention.

See also https://github.com/QuantEcon/QuantEcon.py/pull/489#issuecomment-508338109.

QBatista:
@oyamad Thank you for pointing this out! I kept the original name from the class in the lecture because `AMF_LSS_VAR` would become `AMFLSSVAR` or `AmfLssVar` under PEP8 which I find hard to read. We could also use the name `AdditiveMultiplicativeFunctional`. This name is less informative than `AMF_LSS_VAR` but it shouldn't be an issue given that the documentation takes care of explaining the purpose of this class in detail. Please let me know if you have any other ideas.

jstac:
+1 for following PEP8.  I suspect the corresponding lectures would need to be edited though, since they use the existing class name.

mmcky:
Closing this PR as `stale` > 12 months.

PR #484, opened 2019-05-23
Add drift term keyword to `markov.tauchen`.

shizejin:
Try to make `markov.tauchen` able to handle non-zero drift term in the AR(1) process.

[Here](https://nbviewer.jupyter.org/github/shizejin/Notebooks/blob/master/AR%281%29%20with%20drift%20term.ipynb) discusses the logic of modification made.

### DIFF
--- a/quantecon/markov/approximation.py
+++ b/quantecon/markov/approximation.py
@@ -135,20 +135,22 @@ def rouwenhorst(n, ybar, sigma, rho):
     return MarkovChain(theta, bar)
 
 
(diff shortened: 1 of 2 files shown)


jstac:
@shizejin Great work and great analysis.

I was surprised the standard deviation wasn't a bit closer so I tried with m=4 and n=150.  It was much better.

(I remember now that the approximation actually gets worse when m gets large.)

@mmcky I approve this PR.  I would be glad if we could get this merged and a new version of QE.py released quickly to fix the bug @AnjuJoon found.

Issue #460, opened 2018-12-11
docs won't compile local version


natashawatkins:
I'm getting this error:

![image](https://user-images.githubusercontent.com/21038129/49778096-84fdae80-fd57-11e8-9d7b-0987819679a0.png)

rht:
Strange that the `matplotlib` error happens 1. only recently, 2. only at a specific `:math:` line. Possibly `matplotlib.sphinxext.only_directives` used to be builtin, then it is no longer the case at most recent sphinx version.

mmcky:
@natashawatkins I am not getting this error on my machine.

oyamad:
I have the same problem with `Python 3.7.2` and `Sphinx 2.0.0`.

oyamad:
Any update on this issue??

mmcky:
@oyamad I will be back in the office on Tuesday so will try and replicate this issue using `sphinx=2.0`. I have been using `sphinx=1.7.8` with the latest python with no issues.

oyamad:
No longer relevant?

mmcky:
Confirmed -- I can run the docs build locally. Thanks @oyamad

PR #488, opened 2019-06-24
FIX: DOC: Remove `matplotlib.sphinxext.only_directives`

oyamad:
See #460.
Just removed `matplotlib.sphinxext.only_directives` in `docs/source/conf.py`. It works fine on my environment with Anaconda.


mmcky:
thanks @oyamad

PR #489, opened 2019-07-02
Add `LQMarkov`.

shizejin:
Integrating the code for solving Markov jump LQ dynamic programming problems from the [notebook](https://github.com/QuantEcon/python_lecture_sandpit/blob/master/source/rst/tax_smoothing_1.rst) in the sandpit repository.

The modifications I have made comparing to the original code:

1. modify the notations to be consistent with the existing `LQ` class,
2. discard the usage of `dict` and adopt `ndarray` to store transition, payoff matrices, policies, and value function arrays,
3. arrange the part of matrix calculation to make it easier to read and more efficient.
4. add tests.

[Here](https://nbviewer.jupyter.org/github/shizejin/Notebooks/blob/master/Demonstration%20of%20%60qe.LQ_Markov%60.ipynb) is a demonstration for this `qe.LQ_Markov` class.

Two points that I would like to have your opinions before I move on:

1. the original code only applies to the case with infinite horizon. I am wondering if I should try to add code for finite horizon case as well?
2. parallelization over Markov states during iterations of the stacked system of discrete riccati equations could be very beneficial for the cases with large number of Markov states. Shall we try to explore whether we want to implement this or not?

I will add more tests in the next few days.

### DIFF
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -1,14 +1,16 @@
 """
 Provides a class called LQ for solving linear quadratic control
-problems.
+problems, and a class called LQMarkov for solving Markov jump
+linear quadratic control problems.
 
 """
 from textwrap import dedent
 import numpy as np
 from numpy import dot
 from scipy.linalg import solve
-from .matrix_eqn import solve_discrete_riccati
+from .matrix_eqn import solve_discrete_riccati, solve_discrete_riccati_system
 from .util import check_random_state
+from .markov import MarkovChain
 
 
 class LQ:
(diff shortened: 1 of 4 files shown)


quantecon/lqcontrol.py
```diff
@@ -327,3 +329,279 @@ def compute_sequence(self, x0, ts_length=None, method='doubling',
...
+    Here :math:`x` is n x 1, :math:`u` is k x 1, :math:`w` is j x 1 and the
+    matrices are conformable for these dimensions.  The sequence :math:`{w_t}`
+    is assumed to be white noise, with zero mean and
+    :math:`\mathbb{E} [ w_t' w_t ] = I`, the j x j identity.
+
+    If :math:`C` is not supplied as a parameter, the model is assumed to be
+    deterministic (and :math:`C` is set to a zero matrix of appropriate
+    dimension).
+
+    The optiaml value function :math:`V(x_t, s_t)` takes the form
```
jstac: "optimal"

quantecon/lqcontrol.py
```diff
@@ -327,3 +329,279 @@ def compute_sequence(self, x0, ts_length=None, method='doubling',
...
+    m : scalar(int)
+        The number of Markov states
+    k, n, j : scalar(int)
+        The dimensions of the matrices as presented above
+
+    """
+
+    def __init__(self, Π, Qs, Rs, As, Bs, Cs=None, Ns=None, beta=1):
+
+        # == Make sure all matrices can be treated as 3D arrays == #
```
jstac: 2D?

jstac:
Beautifully written code.  Great work @shizejin 

Regarding suggestions 1--2, let's wait and see if there's demand before we make those changes (but thanks for the suggestions).  

Rather than those additions, please have a think about the tests and see if you can keep coverage high.

quantecon/lqcontrol.py
shizejin: Here I was trying to see that all the arrays (`Qs`, `Rs`, etc.) consist of 2d arrays, but I guess it is confusing. Would "Make sure all matrices for each state are 2D arrays" be a little better?

quantecon/lqcontrol.py
shizejin: Thanks!

quantecon/lqcontrol.py
jstac: No, I just mean this: In your comment on line 423, you say `3D` but do you actually intend to say `2D`?

shizejin: I actually intended to say 3D since I was referring to the shape of `Qs` array (and `Rs` etc.) and `Qs` has shape m (number of Markov states) x n (number of state variables) x n, but after you pointed this out, I realized `Qs` is not a matrix and saying "matrices are 3D arrays" does not make any sense.. so I think maybe the comment I proposed above would be a little more clear

shizejin:
Hi @jstac, I added more tests and the newly added lines in 4 changed files are fully covered. I think this is ready to be merged :)

shizejin:
Screen Shot 2019-07-03 at 23 53 03

It seems that the coverage rates increase or stay the same for changed files, but the total coverage rate decreases by 0.9%. I am not sure what is happening here...

jstac:
Thanks @shizejin .

As far as I can tell, you don't have a test for `solve_discrete_riccati_system`.  Could that be the reason coverage goes down?

shizejin:
Thanks @jstac for your suggestion! I had a look at the coverage report and the lines in `solve_discrete_riccati_system` are fully covered. This is because this function is called in `LQ_Markov` and I am testing for all the possible situations (raising erros etc.), so I think this should be fine.

I think the following is what causes our confusion:

For each PR, we run two build jobs on Travis, one for `linux` and one for `osx`. Both jobs submit coverage reports after success, and the coverage badge is updated by the one that is submitted earlier. Usually the `linux` build job runs faster, but yesterday the coverage report from `osx` build job is submitted earlier. In details, this "Add `LQ_Markov`" PR will update the coverages by

- `linux`: 93.95 -> 94.11,
- `osx`: 92.86 -> 93.05.

But since `osx` coverage report is submitted earlier, the coverage badge is updated by

- badge: 93.95 -> 93.05.

This is why we observe a decrease in the coverage.

I am trying to fix this by #490, which modifies Travis file to make sure that we only update coverage badge by reports from the `linux` build job. By doing so, I think the coverage updates can be more consistent.

oyamad:
Just a comment on naming: I guess in Python usually underscores are not used in a class name; if that is the case, would `LQMarkov` be better?

According to [PEP8](`https://www.python.org/dev/peps/pep-0008/#id41`):

> Class names should normally use the CapWords convention.

Does "CapWords convention" exclude use of an underscore?

jstac:
@shizejin Good plan!

@oyamad You are right.  @shizejin , could you please change the class name as @oyamad suggests?

shizejin:
Thanks @oyamad for your suggestion!

@jstac sure, I will modify the PR after #490 is merged.

shizejin:
Hi @jstac @oyamad, would you mind to let me know if you have any other comments or suggestions? Otherwise I think this is ready to merge :)

jstac:
Very nice work.  I approve this PR.

jstac:
@mmcky It would be good to have this merged soon, as we'll follow it up with some new lectures.  Please have a look when you can.

Thanks.

mmcky:
thanks @shizejin and @jstac.

PR #490, opened 2019-07-04
TRAVIS: Set coverage branch as `linux`.

shizejin:
Only submit coverage reports from the `linux` build job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,9 @@ script:
   - nosetests --with-coverage -a "!slow" --cover-package=quantecon
 
 after_success:
-  - coveralls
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @shizejin. Do you find coveralls not working on an other platform?

shizejin:
Hi @mmcky, coveralls is working in both `linux` and `osx`, but I found that submitting coverage reports from both platforms causes inconsistent updates (see discussion #489), and I guess this PR could solve this problem.

mmcky:
I see. thanks @shizejin happy to accept coverage reports from linux builds.

PR #498, opened 2019-07-19
FIX: two minor modifications in `lqcontrol`

shizejin:
I made two fixes  for `LQ` and `LQMarkov` classes.

1. There was a typo in calculating the values of the last period states for `LQMarkov` models. I forgot to left multiply the `C` matrix to the vector of shocks. I have fixed this and modified tests correspondingly.

2. The code was computing stationary values each time we call `compute_series`, which is unnecessary since the optimal policy `F` are stored as class attribute and do not change. This increases running time especially for lectures that simulate multiple paths for the same model, e.g. "How to Pay for a War: Part 1". I modified it to only compute stationary values once when `compute_series` method is called for the first time. By doing so, the running time of "How to Pay for a War: Part 1" decreases from `42s` to `12s`.

The first point is closely related to the work about the series of Markov Jump LQ DP lectures. I feel sorry that I fail to spot this typo. It would be great if we can make another release before we publish those lectures. However, since this bug only matters for values in the last period and almost has no influence on the patterns of the graphs that we draw, I guess it will also be fine that we publish lectures first and include these fixes in a release with other commits later.

### DIFF
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -605,6 +605,6 @@ class LQMarkov:
             u_path[:, t] = - (Fs[state[t]] @ x_path[:, t])
         Ax = As[state[T]] @ x_path[:, T-1]
         Bu = Bs[state[T]] @ u_path[:, T-1]
-        x_path[:, T] = Ax + Bu + w_path[:, T]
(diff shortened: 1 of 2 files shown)


jstac:
Thanks @shizejin, I appreciate the fix.

@oyamad @mmcky This looks good to me.

PR #518, opened 2019-12-06
Adds a rank size plot to inequality

jstac:
Generates a rank size plot from supplied data.  This is quite useful and the code has been written to avoid new external dependences.

No test and I don't have time to write one now.  Would appreciate if we can merge this subject to review and then think about proper tests later.


mmcky:
thanks @jstac this looks great.

PR #521, opened 2020-01-06
Add matplotlib as required

gliptak:
Allows running https://hub-binder.mybinder.ovh/user/quantecon-quant-otebooks-python-whx3pz3v/notebooks/additive_functionals.ipynb

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,7 @@ setup(name='quantecon',
           'requests',
           'scipy>=1.0.0',
           'sympy',
+          'matplotlib',
           ],
       include_package_data=True
       )


mmcky:
thanks @gliptak -- that is a notebook from the lecture series. 

We have kept `matplotlib` out of the code library as it isn't directly required and previously slowed the import of `quantecon` so it was removed. 

`matplotlib` is included as an import in that lecture. The key assumption behind the lecture series is that the lectures will run on the latest `anaconda` python distribution. Perhaps I need to update the `mybinder` spec?

gliptak:
@mmcky thank you for the pointer

https://github.com/jupyterhub/binderhub/blob/master/requirements.txt

I think I does pick up a ```requirements.txt``` from the project loaded

mmcky:
thanks @gliptak.

PR #524, opened 2020-01-24
remove old test commands from Makefile

### LINKED ISSUE #523: Make test-all
Hi all,

There are two issues with the "make test-all" command. First, nose tests pass but then test-examples fails since there is no scripts/test-examples.py directory (or file) or test-solutions fails since there is no scripts/test-solutions.py (directory or file) in the home directory (under Quantecon.py/ directory). However the individual tests pass when I go to, say for example, quantecon/game_theory/tests/ and run a specific test*.py test. 

Will be helpful if you can guide where the missing scripts/ directory is or how it gets generated.

thanks

mmcky:
fixes #523

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,16 +3,6 @@
 install:
 	python setup.py install
 
-test-all: test test-examples test-solutions
-
 test:
 	@echo "Running nosetests on test suite ..."
 	nosetests -v
-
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @oyamad

PR #542, opened 2020-05-10
Fix variables never used lgtm warnings in dle.py.

duncanhobbs:
This PR addresses [these](https://lgtm.com/projects/g/QuantEcon/QuantEcon.py/snapshot/b54e85c50aa0535bdea6982134df2936c0805a03/files/quantecon/dle.py?sort=name&dir=ASC&mode=heatmap) lgtm alerts for `dle.py`.

### DIFF
--- a/quantecon/dle.py
+++ b/quantecon/dle.py
@@ -69,7 +69,7 @@ class DLE(object):
         self.nb, self.nh = self.llambda.shape
         self.nd, self.nc = self.phic.shape
         self.nz, self.nw = self.c2.shape
-        self.junk, self.ng = self.phig.shape
+        _, self.ng = self.phig.shape
         self.nk, self.ni = self.thetak.shape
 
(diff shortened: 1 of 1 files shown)


quantecon/dle.py
```diff
@@ -69,15 +69,14 @@ def __init__(self, information, technology, preferences):
         self.nb, self.nh = self.llambda.shape
         self.nd, self.nc = self.phic.shape
         self.nz, self.nw = self.c2.shape
-        junk, self.ng = self.phig.shape
+        self.junk, self.ng = self.phig.shape
```
oyamad, line 72: I would say this should be

```suggestion
        _, self.ng = self.phig.shape
```
```

duncanhobbs: Thank you for pointing this out. I updated the PR accordingly.

PR #544, opened 2020-05-11
Add interpolation methods

najuzilu:
This PR adds interpolation methods provided by @jstac in order to replace our lectures' dependency on the [interpolation](https://pypi.org/project/interpolation/) package. This solution was suggested [here](https://github.com/QuantEcon/lecture-python/issues/198).

The methods can be found under the `interpolation` directory. I have removed the test method as tests will need to be implemented under a `test` directory.

The following need to be implemented:
- [ ] test cases

Feedback would be greatly appreciated.

### DIFF
--- a/quantecon/interpolation/__init__.py
+++ b/quantecon/interpolation/__init__.py
@@ -1,0 +1,6 @@
+# flake8: noqa
+"""
+Initialization of the interpolation subpackage
+"""
+
+from .utilities import quantile, lininterp_1d, lininterp_2d, lininterp_3d, lininterp_4d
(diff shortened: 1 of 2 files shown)


mmcky:
The `interpolations` package was updated to cater to earlier version of `numba` so this should only go ahead if we would like to support some `interpolation` routines. 

Thanks for putting this together @najuzilu but let's put this on hold given there is a nice solution already out there that we can use.

mmcky:
@jstac I think in general -- the `interpolation` routines in the `interpolations` package will reduce our maintenance burden.  Are you happy to keep using those or do we want to revive this PR?

jstac:
Agree, thanks @mmcky , closing.

Issue #508, opened 2019-09-07
ValueError: `LQMarkov` with `beta=1`


oyamad:
```py
Π = np.array([[0.8, 0.2],
              [0.2, 0.8]])
Qs = np.array([[[0.9409]], [[0.870489]]])
Rs = np.array([[[1., 0., 1.],
                [0., 0., 0.],
                [1., 0., 1.]],
               [[1., 0., 1.],
                [0., 0., 0.],
                [1., 0., 1.]]])
Ns = np.array([[[-0.97, 0., -0.97]],
               [[-0.933, 0., -0.933]]])
As = np.array([[[0., 0., 0.],
                [0., 1., 0.],
                [0., 5., 0.8]],
               [[0., 0., 0.],
                [0., 1., 0.],
                [0., 5., 0.8]]])
B = np.array([[1., 0., 0.]]).T
Bs = [B, B]
C = np.array([[0., 0., 1.]]).T
Cs = [C, C]

lq_markov_mat1 = qe.LQMarkov(Π, Qs, Rs, As, Bs,
                             Cs=Cs, Ns=Ns, beta=1.)
lq_markov_mat1.stationary_values()
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-15-022c46ef29a8> in <module>
----> 1 lq_markov_mat1.stationary_values()

/usr/local/lib/python3.7/site-packages/quantecon/lqcontrol.py in stationary_values(self)
    506 
    507         # == Solve for P(s) by iterating discrete riccati system== #
--> 508         Ps = solve_discrete_riccati_system(Π, As, Bs, Cs, Qs, Rs, Ns, beta)
    509 
    510         # == calculate F and d == #

/usr/local/lib/python3.7/site-packages/quantecon/matrix_eqn.py in solve_discrete_riccati_system(Π, As, Bs, Cs, Qs, Rs, Ns, beta, tolerance, max_iter)
    290 
    291         if iteration > max_iter:
--> 292             raise ValueError(fail_msg.format(max_iter))
    293 
    294         else:

ValueError: Convergence failed after 1000 iterations.
```

duncanhobbs:
I tested this same example with different values of `beta` below. I still get a convergence error when `beta` is larger than 0.972.


**beta**   error after 1000 iterations:

**1**      51.302186197106494

**0.99**   0.0022754969468223862

**0.98**   1.0506482794880867e-07
   
[...]

PR #550, opened 2020-06-05
FIX: ValueError `LQMarkov` convergence failed

bktaha:
Hi!

For #508 I noticed the solver method matrix_eqn.solve_discrete_riccati_system() allowed for setting an upper bound on the number of iterations, but the stationary_values() method of LQMarkov didn't make use of this arg so it threw an error if it failed to converge within the preset number of iterations, 1000.

I've added an optional arg to the LQMarkov init method to increase the upper bound on the number of iterations, and it was able to converge for the betas mentioned in #508 ( converged for 0.973 in 1001 iterations, for 0.974 in 1050 and for 0.99 in 2700 ).

I didn't add a test for this since this was a small enough change, but I can add it if needed.

### DIFF
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -517,8 +517,8 @@ class LQMarkov:
         Qs, Rs, Ns = self.Qs, self.Rs, self.Ns
 
         # == Solve for P(s) by iterating discrete riccati system== #
-        Ps = solve_discrete_riccati_system(
-            Π, As, Bs, Cs, Qs, Rs, Ns, beta, max_iter=max_iter)
+        Ps = solve_discrete_riccati_system(Π, As, Bs, Cs, Qs, Rs, Ns, beta,
+                                           max_iter=max_iter)
 
         # == calculate F and d == #
         Fs = np.array([np.empty((k, n)) for i in range(m)])
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -406,8 +406,6 @@ class LQMarkov:
         k x n for each Markov state,
     beta : scalar(float), optional(default=1)
         beta is the discount parameter
-    max_iter : scalar(int), optional(default=1000)
-        The maximum number of iterations allowed
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @bktaha. 

@oyamad are you happy with adding `max_iter=1000` as a default value. This is outside of my domain knowledge so wanted to check with you on this.

mmcky:
@bktaha can you update the `docstring` with this optional argument

https://github.com/QuantEcon/QuantEcon.py/blob/9d0fb5bf253648255b4d2e6029320a7795355a95/quantecon/lqcontrol.py#L66

oyamad:
I would like to delegate it to the author @shizejin.

@shizejin Will you review this PR?

shizejin:
Sure @oyamad.

Thanks @bktaha, this is going to be a very meaningful improvement of the code. One minor point is that `max_iter` is purely a computational parameter, which means it should be a keyword argument of the method `stationary_values` instead of the `LQMarkov` class. It would be even better if you would be willing to take care of the indentations in [line432](https://github.com/QuantEcon/QuantEcon.py/pull/550/files#diff-dd708f84523a1e5c729e0725ebb7d5b8R432) and [line521](https://github.com/QuantEcon/QuantEcon.py/pull/550/files#diff-dd708f84523a1e5c729e0725ebb7d5b8R521).

I agree with you that a test for this is not needed.

bktaha:
Thanks @shizejin for the kind comments, and for reviewing this PR.

As you suggested, I've moved the `max_iter` parameter to the `stationary_values` method and fixed the `docstring` to reflect that change. The line indent issue is taken care of as well.

shizejin:
Thanks @bktaha for your quick response. The PR looks great to me except that I think the following way of breaking a line might be favorable:

```python
Ps = solve_discrete_riccati_system(Π, As, Bs, Cs, Qs, Rs, Ns, beta,
                                   max_iter=max_iter)
```

The difference between this and the previous version of your code is that one space before `max_iter` is deleted so that the starting of the second line is aligned with `(`. But this is no more than my personal opinion and is really a minor point which does not affect the good quality of your commit. It would be great to have opinions from @oyamad and @mmcky, and I recommend merge for this PR after this is set.

mmcky:
thanks @bktaha, @shizejin and @oyamad.

I agree with @shizejin style comment on line breaks. It is in line with common `python` practice. 
@bktaha thanks so much for submitting and being willing to iterate on this PR. It is greatly appreciated. Once that minor update has been done I will merge.

bktaha:
@mmcky I've made the suggested style fix.

mmcky:
thanks @bktaha for the PR -- really appreciate the contribution!

Thanks @shizejin and @oyamad for your input and comments.

PR #551, opened 2020-06-16
REF and TEST: rank_size in inequality.py

bktaha:
Based on the discussion in #519, refactor the method `rank_size_plot` in `inequality.py` so the output now is the data to be plotted instead of a pyplot `Axes` object. This is in line with the rest of the methods in this module. The method is renamed to `rank_size` to reflect this change.

Also added test for the new method, which checks that
- The `x` and `y` arrays are the same length.
- This length is `(c x 100)%` of the input, as specified.
- The output log-log plot is linear for a Pareto draw, and not otherwise.

Edited the `docstring` for the other two methods in the module for pep8 compliance.

This closes #519

PR #556, opened 2020-06-30
FIX: rank-size test by inc. sample size

### LINKED ISSUE #553: Issue with test for rank_size
@bktaha would you have any time to look into this issue with the rank_size test? 

`Travis` was broken due to the migration to `.com` from the `.org` site and QuantEcon.py was using legacy connections -- which has now been fixed. 

I'll hold off on issuing a new release until this is fixed. 

```
FAIL: Tests `rank_size` function, which generates rank-size data for
896----------------------------------------------------------------------
897Traceback (most recent call last):
898  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
899    self.test(*self.arg)
900  File "/home/travis/build/QuantEcon/QuantEcon.py/quantecon/tests/test_inequality.py", line 119, in test_rank_size
901    assert_allclose(r_sqval, 1, rtol=1e-4)
902  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 1528, in assert_allclose
903    verbose=verbose, header=header, equal_nan=equal_nan)
904  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 840, in assert_array_compare
905    raise AssertionError(msg)
906AssertionError: 
907Not equal to tolerance rtol=0.0001, atol=0
908
909Mismatched elements: 1 / 1 (100%)
910Max absolute difference: 0.00188538
911Max relative difference: 0.00188538
912 x: array(0.998115)
913 y: array(1)
````

bktaha:
As mentioned in PR #551 the test for `rank_size` method checks for linearity in the output log-log plot. I did this by performing linear regression of `log(size_data)` on `log(rank_data)` and comparing the resulting `r_squared` value to 1.

Turns out an error tolerance of `1e-4` for a sample size of `1000` was a little optimistic. I have increased the sample size to `10000`, reduced the tolerance to `1e-3` and the comparison works as expected now.

Another mistake was setting the seed for exponential draw but not the Pareto draw, I've fixed that as well. I'm not sure if setting a seed here is a good practice in the first place, if not I can remove it and increase sample size / reduce tolerance further.

Please let me know if this is an acceptable fix. This should close #553.

### DIFF
--- a/quantecon/tests/test_inequality.py
+++ b/quantecon/tests/test_inequality.py
@@ -103,7 +103,8 @@ def test_rank_size():
     of the size of the distribution.
     """
 
-    sample_size = 1000
+    np.random.seed(15)
+    sample_size = 10000
     c = 0.74
 
     # Tests Pareto; r_squared ~ 1
@@ -116,10 +117,9 @@ def test_rank_size():
     _, _, r_value, _, _ = linregress(np.log(rank_data), np.log(size_data))
     r_sqval = r_value**2
 
-    assert_allclose(r_sqval, 1, rtol=1e-4)
+    assert_allclose(r_sqval, 1, rtol=1e-3)
 
     # Tests Exponential; r_squared < 1
-    np.random.seed(13)
     z = np.random.randn(sample_size)
 
     exp_draw = np.exp(z)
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @bktaha for this well documented PR and for looking into the test failure. It is greatly appreciated. 
@jstac would you mind to cast a quick eye over this if you have time?

jstac:
Nice work @bktaha.  Many thanks. This looks good to me @mmcky.

mmcky:
thanks @jstac

PR #560, opened 2020-12-16
Update root_finding.py

alanlujan91:
when x0 >= 0, q0 is not initially assigned, but since this function is jitted, I believe q0 = 0.0 when referenced before assignment

to notice the error, remove @njit decorator and set x0 >= 0


mmcky:
thanks @alanlujan91 for the PR. @jstac would you have any time to take a quick look at this?

sglyon (approved):
LGTM!

mmcky:
thanks @sglyon

PR #561, opened 2021-01-01
[GitHub Actions] Setup Tests via Github Actions

rht:


rht:
(whoops, I was about to make this PR on my fork instead of here!)
Oh well, anyway, since travis-ci.org is gone, we need to switch to another CI.

rht:
@Harveyt47 @mmcky I finally got 94.15% from coveralls.io: https://github.com/QuantEcon/QuantEcon.py/pull/561/checks?check_run_id=1633166620#step:7:15 -> https://coveralls.io/jobs/72880434.

rht:
My fix is to do according to this comment: https://github.com/coverallsapp/github-action/issues/4#issuecomment-727600521.

rht:
Let me know if the commits need to be squashed. They are kind of messy.

mmcky:
thanks @rht!. We will definitely Squash these commits. One question before we merge -- I see you added a GitHub Repo Secret for the Tokens Coveralls. Just wondering if we should use a `quantecon` account -- I am guessing that is a personal token?

rht:
> I see you added a GitHub Repo Secret for the Tokens Coveralls.

Nope, I was in fact removing it.
According to the readme of https://github.com/AndreMiras/coveralls-python-action
this is the default value of `GITHUB_TOKEN`:
> Default: ${{ github.token }}

rht:
Also, I removed .coveralls.yml instead, so it is no longer tied to travis-ci.

rht:
I added another commit to remove pytest since it is no longer used in the latest version of this PR.

Harveyt47:
Thanks @rht I struggled with the coveralls.io a bit. Is this still running `nose` and not `pytest`

rht:
Yes, this is still running `nose`. If you want to migrate to `pytest`, I think we can do that in another PR.

rht:
Ready to be squashed.

mmcky:
Thanks @rht -- it will be nice to have tests via GitHub Actions.

PR #562, opened 2021-01-01
[BUGFIX] kalman: Always initialize self.Sigma and self.x_hat

rht:


mmcky:
thanks @rht -- good catch. Thanks for your contributions -- in particular setting up smaller focused PR's -- its a big help with merging.

PR #574, opened 2021-03-14
TEST: Add python=3.9

### LINKED ISSUE #559: Compatibility with Python=3.9
It looks like we may have some compatibility issues with `python=3.9`

https://github.com/conda-forge/quantecon-feedstock/pull/23#issuecomment-739217657

oyamad:
`numba=0.53` has been released: https://numba.discourse.group/t/numba-0-53-0-and-llvmlite-0-36-0-final-release

Close #559

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ on:
     branches:
     - master
   pull_request:
-   
+
 jobs:
   tests:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


oyamad:
There seems to be an issue in installing `tables` for `python=3.9` on macos.
I propose to remove dependency on `tables`. It is called only in `get_h5_data_file`, `get_h5_data_group`, and `write_array` which are not used anywhere.

mmcky:
thanks @oyamad

PR #579, opened 2021-04-30
EHN: Add minmax solver

oyamad:
This PR adds a `minmax` solver to the `quantecon` package. 

This solver is also used by default in `normal_form_game` module in the `is_dominated` method


mmcky:
thanks @oyamad once this passes tests I will merge this in today. Just fixed a minor `merge` conflict re: documentation.

Issue #466, opened 2018-12-17
Review LGTM Code Quality Suggestions


mmcky:
We currently have a very high mark on code quality thanks to testing and linting etc. 

However we should review the code suggestions provided by LGTM

https://lgtm.com/projects/g/QuantEcon/QuantEcon.py/alerts?mode=list

nshea3:
Hi @mmcky! Would be interested in taking on this work. 

I noticed the issue is a little old, is there still an interest in getting this done?

mmcky:
hi @nshea3 that would be great - thanks. I'm keen to see what suggestions it has. But `on other projects` I have grown to like some of the `pre-commit` style looks like `black` etc. for code-reformatting.

oyamad:
@mmcky This has been done and can be closed?

mmcky:
thansk @nshea3 and @oyamad

PR #588, opened 2021-10-02
LGTM code quality suggestions

nshea3:
Hi all! 

This PR addresses the code quality alerts in #466 

Three different things that LGTM picked up in the repo: 

1. An `import *` in quantecon/game_theory/__init__.py. This one was a relatively straightforward fix, just updated with the specific imports required.
2. "wrong number of arguments" in quantecon/util/timing.py. I think this one is a false alarm - there is branching logic to distinguish between the no argument, one argument, and many argument cases. Triggered by a call in TestTicTacToc.test_loop.test_function_one_arg. **Have not made any changes to this code since I believe LGTM is mistaken here.**
3. Module (numpy) is imported with 'import' and 'import from' in a couple different files. This is triggered by an `import numpy as np` followed by `from numpy import dot` etc. which is pretty reasonable especially with big formulas with lots of linear algebra. I see two possible approaches here: 
      - Replace all `dot(A,B)` calls with `np.dot(A,B)`
      - Replace all `dot(A,B)` calls with `A.dot(B)` 

For completeness here I defaulted to the first approach since it's a one-to-one drop in replacement and won't require any significant rewrites. However it does not help with readability or brevity, so I can certainly drop those commits from the PR if need be. 
Happy to give the second approach a try as well. 

LGTM list for reference: https://lgtm.com/projects/g/QuantEcon/QuantEcon.py/alerts?mode=list

Looking forward to your feedback on this! Thanks,

Nick

### DIFF
--- a/quantecon/game_theory/__init__.py
+++ b/quantecon/game_theory/__init__.py
@@ -13,5 +13,7 @@ from .support_enumeration import support_enumeration, support_enumeration_gen
 from .lemke_howson import lemke_howson
 from .mclennan_tourky import mclennan_tourky
 from .vertex_enumeration import vertex_enumeration, vertex_enumeration_gen
-from .game_generators import *
+from .game_generators import (
+    blotto_game, ranking_game, sgc_game, tournament_game, unit_vector_game
+)
 from .repeated_game import RepeatedGame
--- a/quantecon/lqcontrol.py
+++ b/quantecon/lqcontrol.py
@@ -6,7 +6,6 @@ linear quadratic control problems.
 """
 from textwrap import dedent
 import numpy as np
-from numpy import dot
 from scipy.linalg import solve
 from .matrix_eqn import solve_discrete_riccati, solve_discrete_riccati_system
 from .util import check_random_state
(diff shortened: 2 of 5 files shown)


jstac:
Thanks @nshea3 , we appreciate the contribution!

Actually my preference is that all of those `np.dot` calls are replaced by `@` syntax, which makes the matrix algebra more readable.  Some of that code was written before `@` was implemented as matrix multiplication.

I'll let the maintainers (@mmcky and @oyamad) weigh in as well.

mmcky:
thanks @nshea3 these look like good changes. Sorry for my delay on this. 

I like @jstac suggestion of switching over to using `@` instead of `np.dot`. But happy for us to merge these LGTM code quality suggestions and then migrate `np.dot` to `@` in another PR.

nshea3:
Hi @mmcky ! No worries! 
I have explored switching over to the `@` instead of `np.dot`. However that is going to be a little more involved - it's not a drop-in replacement since it is *much* more particular about the datatypes and matrix dimensions it will accept as arguments. I'll open up a separate issue and PR to discuss those changes.

Happy to merge this in for the time being or close without merging, whichever seems best to you and @jstac

mmcky:
thanks @nshea3 -- Thanks for looking into the `@` migration. This explanation is helpful when thinking about the differences

https://stackoverflow.com/questions/34142485/difference-between-numpy-dot-and-python-3-5-matrix-multiplication

Once an issue is open I will go ahead and merge these improvements and then we can loop back around with `@` operator.

mmcky:
@nshea3 I have opened this issue

https://github.com/QuantEcon/QuantEcon.py/issues/589

mmcky:
thanks @nshea3 for these changes in code quality. Greatly appreciated.

PR #596, opened 2022-01-11
MAINT: Replace `np.int`

oyamad:
To remove ``DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`.``


mmcky:
thanks @oyamad

Issue #539, opened 2020-05-09
Convert test suite from nosetests to pytest 


mmcky:
It would be good to migrate our test suite from `nosetests` to `pytest`

https://docs.pytest.org/en/latest/

@Harveyt47 do you have time to look into this?

Harveyt47:
Yep will do @mmcky .

Harveyt47:
`pytest` has support for `nosetests` see [here](https://docs.pytest.org/en/latest/nose.html). For example this [test](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/tests/test_approximation.py#L12) will work as is.

However when there is not support we will have to change functions like `setUp` to `setup_method` see [here](https://docs.pytest.org/en/latest/xunit_setup.html?highlight=setup#method-and-function-level-setup-teardown). For example the `setUp` function [here](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/tests/test_core.py#L179) needs to be changed to `setup_method`

Harveyt47:
yield tests were removed in pytest 4.0, see [here](https://docs.pytest.org/en/latest/deprecations.html#yield-tests) for needed switch

oyamad:
@mmcky What is the status of this issue? It seems `nose` will no longer work on Python 3.10 (nose-devs/nose#1099).

mmcky:
thanks @oyamad sounds like this can no longer sit on the back burner. 

I'll need to reserve some time to make the migration.

mmcky:
@oyamad we have a new RA coming on board this month (hopefully) -- I might ask him to help me with this task (cc: @jstac)

Smit-create:
Should I work on this?

oyamad:
@Smit-create Thanks, that would be great.

I converted one file in the `game_theory` module in #597. The resulting tests work with both `nose` and `pytest`.

oyamad:
Regarding PR #610 (thanks @Smit-create!), I didn't realize the `PytestCollectionWarning`s: [`yield tests were removed in pytest 4.0 - test_XXX will be ignored`](https://github.com/QuantEcon/QuantEcon.py/runs/5546271326#step:7:44). We need to find a substitution for `yield` in test code.

Smit-create:
Yeah, will have a look into it.

Smit-create:
I found the fix for that using the following:
```py
test_classes = [
    StationaryDistSumOne,
[...]

PR #597, opened 2022-01-11
MAINT: TEST: Remove `nose` functions

oyamad:
Replaced with `numpy.testing` functions

Only for `game_theory/game_generators/tests/test_bimatrix_generators.py`

Ref: #539


mmcky:
thanks @oyamad

PR #600, opened 2022-02-10
CI: Add a CI job for testing using conda

### LINKED ISSUE #514: Update GitHub Actions to check against latest Anaconda Distribution
As we recommend users to use `anaconda` we should update ~~`travis`~~ `github` to use the latest `anaconda` distribution. This will be slower for `CI` but more consistent with user reports of versions (particularly relating to `numba` errors and issues)

Smit-create:
Fixes #514 

cc @mmcky

### DIFF
--- a/.github/workflows/conda_ci.yml
+++ b/.github/workflows/conda_ci.yml
@@ -37,9 +37,9 @@ jobs:
         conda info
         conda list
 
-    - name: Setup and Install - QuantEcon
+    - name: Setup QuantEcon
       run: |
-        python setup.py install
+        python setup.py build
 
     - name: Run Tests
       run: |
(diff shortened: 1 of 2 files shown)


mmcky:
thanks @Smit-create this is looking great -- really appreciate it.

I have approved for `GitHub actions` to run this PR.

Smit-create:
> this is looking great

Thanks! This seems to be working fine. We can also use `conda` environment cache for faster CI (not that effective in case of `mac`) but we can do it as the follow up of this PR.

mmcky:
Thanks @Smit-create -- the Conda cache idea is also great 👍 - thanks for your contribution.

mmcky:
- [ ] We might also deprecate the old `testing`

PR #602, opened 2022-02-11
ENH: Vectorize ECDF's `__call__` method

### LINKED ISSUE #97: ecdf function needs to be vectorized
The **call** method in ecdf.py needs to be vectorized so that it works as a ufunc (acts pointwise on arrays)

Smit-create:
Closes #97 

cc @jstac

### DIFF
--- a/quantecon/ecdf.py
+++ b/quantecon/ecdf.py
@@ -51,4 +51,4 @@ class ECDF:
         def f(a):
             return np.mean(self.observations <= a)
         vf = np.frompyfunc(f, 1, 1)
-        return vf(x).astype(np.float)
+        return vf(x).astype(float)
--- a/quantecon/ecdf.py
+++ b/quantecon/ecdf.py
@@ -48,4 +48,6 @@ class ECDF:
             Fraction of the sample less than x
 
         """
-        return np.mean(self.observations <= x)
+        f = lambda a, b: a <= b
+        vf = np.vectorize(f)
+        return np.mean(vf(self.observations, x))
--- a/quantecon/ecdf.py
+++ b/quantecon/ecdf.py
@@ -51,4 +51,4 @@ class ECDF:
         def f(a):
             return np.mean(self.observations <= a)
         vf = np.frompyfunc(f, 1, 1)
-        return vf(x)
+        return vf(x).astype(np.float)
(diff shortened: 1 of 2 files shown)


Smit-create:
> I'd be glad if you take another try.

Thanks for the review! I tried to fix that in some other way.

Smit-create:
> the output looks good, but the dtype of `e(t)` is `np.object`

Oh, I think that's easy to fix. We can use the following diff to fix that:
```diff
diff --git a/quantecon/ecdf.py b/quantecon/ecdf.py
index 12a2014..647ce7a 100644
--- a/quantecon/ecdf.py
+++ b/quantecon/ecdf.py
@@ -51,4 +51,4 @@ class ECDF:
         def f(a):
             return np.mean(self.observations <= a)
         vf = np.frompyfunc(f, 1, 1)
-        return vf(x)
+        return vf(x).astype(np.float)
```
With this above diff I get:
```py
>>> import numpy as np
>>> from quantecon import ECDF
>>> obs = np.random.randn(100)
>>> e = ECDF(obs)
>>> t = np.linspace(0, 1, 10)
>>> print(e(t))
[0.49 0.56 0.6  0.65 0.69 0.71 0.75 0.77 0.8  0.83]
>>> e(t)
array([0.49, 0.56, 0.6 , 0.65, 0.69, 0.71, 0.75, 0.77, 0.8 , 0.83])
>>> t
array([0.        , 0.11111111, 0.22222222, 0.33333333, 0.44444444,
       0.55555556, 0.66666667, 0.77777778, 0.88888889, 1.        ])
```




> What do you think about using these methods?

I am not exactly sure that we need to use `vectorize` here, because NumPy docs say that it will use broadcasting rules, which in our case may not be applicable always. This commit https://github.com/QuantEcon/QuantEcon.py/pull/602/commits/22d7f5864dd57412358bf67fa0482e9896ae508a used the numpy's vectorize function which was failing on out test in https://github.com/QuantEcon/QuantEcon.py/pull/602#issuecomment-1038845675

jstac:
Excellent job @Smit-create .  Many thanks for this (and sorry for the slow response)!

Smit-create:
Thanks for the review!

PR #610, opened 2022-03-11
MAINT, TST: Remove nose imports and use pytest

Smit-create:
cc @mmcky


oyamad (changes requested):
@Smit-create Thanks! I have a few comments.

oyamad (approved):
Thanks! All look good to me.

mmcky:
thanks @oyamad and @Smit-create.

PR #613, opened 2022-03-24
MAINT: Remove nose imports and use pytest

Smit-create:
Fixes #603  
Fixes #539

PR #614, opened 2022-03-29
[TMP] FIX: Specify dtype=np.int64 to avoid overflow on Windows

oyamad:
Let's see if db2066e765bb04823a75bc882d813ba12ece2eff fixes the problem.


oyamad:
(Incorporated in #613.)

PR #615, opened 2022-04-01
MAINT: Fix pytest warnings and use pytest in Conda CI

Smit-create:
https://github.com/QuantEcon/QuantEcon.py/issues/539#issuecomment-1084615625

PR #618, opened 2022-04-03
MAIN: Use official coveralls github action

mmcky:
Use official coveralls action


mmcky:
@Smit-create I have reconfigured `coveralls` in this PR to change to the `primary` coverall.io` GitHub action which uses coverage.py.

Is there a reason why we aren't using pytest yet in `conda_ci.yml`?

Smit-create:
> Is there a reason why we aren't using pytest yet in `conda_ci.yml`?

I have made that change in #615

Issue #607, opened 2022-02-24
CI failing on Master Branch


mmcky:
It looks like `Conda` tests are failing on the `master` branch. 

This will need to be fixed before any other PR's are merged 

cc @Smit-create

Smit-create:
The tests seems to be failing because of precision error. Might be the issue when we test using random numbers (without seed). We just need to raise the error limit to fix it. I will have a look at it soon.

oyamad:
Also `TestTicTacToc.test_loop` in `util/tests/test_timing.py` is [failing](https://github.com/QuantEcon/QuantEcon.py/runs/5747742400#step:6:66) on `windows-latest, 3.8`.

Smit-create:
> Also `TestTicTacToc.test_loop` in `util/tests/test_timing.py` is [failing](https://github.com/QuantEcon/QuantEcon.py/runs/5747742400#step:6:66) on `windows-latest, 3.8`.

I guess that is a random failure. We can slightly raise the time bar if there are frequent failures due to this.

mmcky:
Closing this now -- given #620

PR #620, opened 2022-04-05
TEST: Add Python 3.10, remove 3.7

oyamad:
I propose to drop 3.7, as testing against 3.7-3.10 would be too much, and it has been dropped in NumPy.


oyamad:
https://github.com/QuantEcon/QuantEcon.py/runs/5826143256#step:7:67: Same failure as https://github.com/QuantEcon/QuantEcon.py/issues/607#issuecomment-1084602050.

Going to re-run jobs.

mmcky:
I agree - thanks @oyamad

PR #622, opened 2022-04-05
TEST: Exclude windows-latest python-version 3.8

oyamad:
I propose to exclude windows-latest python-version 3.8, where we often have a failure for `TestTicTacToc.test_timer`.


mmcky:
> I propose to exclude windows-latest python-version 3.8, where we often have a failure for `TestTicTacToc.test_timer`.

Good idea. Thanks @oyamad

PR #624, opened 2022-04-05
MAINT: rework README and move changes to CHANGELOG.txt

### LINKED ISSUE #517: Updates to readme pages for QuantEcon.py and QuantEcon.jl
There's at least one spelling error in README.md

But more importantly, to reduce the number of pages we need to maintain, I propose that we take the content of [this page](https://quantecon.org/quantecon-py/) and insert it into the README page of the repo.

It might not look as nice but the price is probably worth paying.

More generally, it's not a good idea to have two pages with the same role, and GH readmes are now the established way to describe a library.

Similar comments apply to the Julia side.  CC @sglyon @mmcky @DrDrij

mmcky:
fixes #517 (for python)


mmcky:
@oyamad I propose to relocate the `major` changes from the `README` into the `CHANGELOG.txt` as a central place for looking at the major changes.

PR #631, opened 2022-04-07
FIX: issue with publish via tags

mmcky:
This should fix the issue with `tagged` releases and `automatic publishing to pypi`

### DIFF
--- a/.github/workflows/conda_ci.yml
+++ b/.github/workflows/conda_ci.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - 'v*'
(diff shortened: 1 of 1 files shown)


mmcky:
@Smit-create do you know why `coverage` keeps reporting small decreases in coverage values? I don't see why this PR would change anything to do with coverage. I guess `coverage` thinks its a `code increase` and I should exclude `yaml` files?

Smit-create:
> I don't see why this PR would change anything to do with coverage. I guess `coverage` thinks its a `code increase` and I should exclude `yaml` files?

Not sure why. Need to look more into it.

Issue #639, opened 2022-10-05
Add code for VAR approximation


jstac:
Replicate Matlab routines from ``Finite-State Approximation Of  VAR Processes:  A Simulation Approach'' by Stephanie Schmitt-Grohé and Martín Uribe, July 11, 2010. 

The Matlab code is in the zip file [Discrete_VAR.zip](https://github.com/QuantEcon/QuantEcon.py/files/9712903/Discrete_VAR.zip)

In that same zip file is Python code that replicates the results of the Matlab routine, provided by @crondonm.  Next stages are

1. create a PR
2. shift code to PEP8
3. try to make code more numerically efficient
4. add tests

I've asked @crondonm to submit a PR.

oyamad:
Done by #640!

Issue #633, opened 2022-05-05
TST: Test failure on windows CI


Smit-create:
See the failure:
```
___________________________ TestTicTacToc.test_loop ___________________________

self = <quantecon.util.tests.test_timing.TestTicTacToc object at 0x000001A0[69](https://github.com/QuantEcon/QuantEcon.py/runs/6306536611?check_suite_focus=true#step:7:69)8597C0>

    def test_loop(self):
        if platform == 'darwin':
            # skip for darwin
            return
    
        def test_function_one_arg(n):
            return time.sleep(n)
    
        def test_function_two_arg(n, a):
            return time.sleep(n)
    
        test_one_arg = \
            loop_timer(5, test_function_one_arg, self.h, digits=10)
        test_two_arg = \
            loop_timer(5, test_function_two_arg, [self.h, 1], digits=10)
        for tm in test_one_arg:
            assert(abs(tm - self.h) < 0.01), tm
        for tm in test_two_arg:
>           assert(abs(tm - self.h) < 0.01), tm
E           AssertionError: 0.110115003585[81](https://github.com/QuantEcon/QuantEcon.py/runs/6306536611?check_suite_focus=true#step:7:81)543
E           assert 0.0101150035[85](https://github.com/QuantEcon/QuantEcon.py/runs/6306536611?check_suite_focus=true#step:7:85)815421 < 0.01
E            +  where 0.010115003585815421 = abs((0.11011500358581543 - 0.1))
E            +    where 0.1 = <quantecon.util.tests.test_timing.TestTicTacToc object at 0x000001A0698597C0>.h

quantecon\util\tests\test_timing.py:56: AssertionError
```

This might be because of platform dependencies. Should we raise the bar slightly to fix it?

Smit-create:
I think we should fix this on priority so that the upcoming PRs have Green CI and it would be easy to review them.

jstac:
Thanks @Smit-create, agreed.  @mmcky , I'll leave this to you to and @Smit-create .

PR #640, opened 2022-10-05
ENH: Adding Discrete Approximation of VAR Methods

crondonm:
Hi, 

As discussed with @jstac, I am submitting the routine discrete_var.py discussed in #639.

The routine is based on the work `Finite-State Approximation Of VAR Processes: A Simulation Approach'' by Stephanie Schmitt-Grohé and Martín Uribe, July 11, 2010.

I am attaching the Authors' Matlab file for comparison. Also, I am attaching a notebook to facilitate testing

 
[Discrete_VAR.zip](https://github.com/QuantEcon/QuantEcon.py/files/9717747/Discrete_VAR.zip)


jstac:
Many thanks @crondonm , this will be a very valuable contribution.

@chien-y and @shlff,  would you mind to take a first pass at reviewing this PR?

You'll need to quickly review the underlying paper, get some idea of what the code does, and write up a small notebook that tries to explain it.  Also, 

* can you make the Python code more Pythonic?  For example, function names should be longer and more descriptive.
* I think we should break up the large function `tpm`.  In particular, simulation of the process should be separate.
* Should `tpm` return an instance of `MarkovChain`, like `qe.tauchen`?
* If so, do we need a separate simulation routine or can we use the one in `MarkovChain`?

@Smit-create , I know that you have quite a lot on but I'm hoping that you can finish this PR in the next month or two, reviewing whether the code can be made more efficient and turning the example into a test.

Smit-create:
> @Smit-create , I know that you have quite a lot on but I'm hoping that you can finish this PR in the next month or two, reviewing whether the code can be made more efficient and turning the example into a test.

Thanks, @jstac. Once the code is reviewed and verified, I can look into this and try to modify and make the APIs more user-friendly and clean ups.

shlff:
Thanks @jstac and @Smit-create .

@chien-y and I chatted about this just now. We will work on this, meet again on Monday and find a time to report it to you later.

shlff:
Thanks @jstac . Here is a list of reviewing tasks:
- [ ] understand the Tauchen method in the QuantEcon library
- [ ] quickly read the paper: http://www.columbia.edu/~mu2166/tpm/tpm.pdf
- [ ] read and understand the notebook written by Carlos
- [ ] make the code more Pythonic following PEP8: https://peps.python.org/pep-0008/
- [ ] break the large functions into smaller pieces, especially the simulation part
- [ ] improve the code's efficiency
- [ ] make the return of ``tpm`` consistent with the return of ``qe.tauchen``, including three components: state space, stochastic matrix and simulation for the stochastic process.

jstac:
@shlff and @chien-y, please update me on your progress.

shlff:
Hey @jstac please refer to the list above, I've designed and been implementing some experiments to test different parts of the code:
computing unconditional covariance $\sigma_i$
- [ ] learn doubling algorithm 
- [ ] compare DA with the algebraic algorithm (before/after speeding up with JIT and JAX)
- [ ] simplify macro variable denotations
- [ ] investigate why we need ``SiggJ``
compute state space matrix $S$:
- [ ] make code and math more consistent
- [ ] make code more pythonic
simulating $\Pi$
- [ ] use time magic to measure the time complexity of each part and identify the most time-consuming part
- [ ] figure out a way to speed it up

Goal: compete with the Matlab code
- [ ] compile Matlab code
- [ ] compare Matlab code with python code
- [ ] see what's more if we can do if python code runs longer than matlab one.

jstac:
Adding to the last comment,

jstac:
`tpm` should have a default seed in the argument list in order to help reproducibility.

jstac:
The function `mom` solves a discrete Lyapunov equation.  Replace with existing QE code.

quantecon/discrete_var.py
```diff
@@ -0,0 +1,159 @@
...
+    Pi = np.zeros([n, n])
+    x0 = np.zeros([m, 1])
+    xx = np.zeros([n, m])
+    d = np.sum((S-xx)**2, 1)
+    ind_i = np.argmin(d)
+    Xvec = np.zeros([m, T])
+    mean = [0] * m
+    
+    for t in range(T+Tburn):
+        drw = multivariate_normal(mean, omega).reshape(1,2)
```
jstac, line 97: @crondonm , does this line assume that `m = 2` ?  I'm guessing `2` should be `m` in my edit.

quantecon/discrete_var.py
```diff
@@ -0,0 +1,159 @@
...
+    S  = S[indx[0], :]
+    
+    sum_row = np.sum(Pi,1)
+    for i in range(len(Pi)):
+        Pi[i,:] = Pi[i,:]/sum_row[i]
+    
+    return Pi, Xvec, S
+
+
+def mom(gx, hx, varshock, J=0, method=True):
```
jstac, line 126: @crondonm , FYI, this code is already available in `quantecon` as `solve_discrete_lyapunov`.  I've changed it to use that function (after confirming identical results for the test matrices you use.

jstac:
@shlff I couldn't see your code (due to lack of access).  I edited @crondonm's code and pushed my own commits.  Pleaese comment on the new version if I've missed obvious changes.

We still need to accelerate the simulation, which is very slow (as well as add tests, etc.).

I suggest that we also return a multi-index version of the transition probability matrix P.  Then S doesn't need to contain a lot of unnecessary values --- we can just return the grid.

shlff:
Thanks @jstac . I've updated the access and now you and others should be able to view, edit and comment on it:
- https://colab.research.google.com/drive/1116BYYOw0XiJjvedRNUSf5DrSQkeRchK?usp=sharing

I've updated the new-version code, commented on them (please see comments 1-7), broke down the code by parts and identified the most time-consuming parts in the simulation. For example this part:
```
drw = multivariate_normal(mean, Omega).reshape(m, 1)
x = A @ x0 + drw
```

Next I will do two things:
- modify the code according to comments except for comment 6
- for comment 6, design some experiments to see why it is so slow and then think about some ways to improve it.

jstac:
@crondonm, FYI, the documentation states that the code discretizes $x_t = A x_{t-1} + \Omega e_t$ but in fact it discretizes $x_t = A x_{t-1} + u_t$ where $u_t$ is zero-mean normal with variance-covariance matrix $\Omega$.

(Note that $\Omega e_t$ has variance-covariance matrix $\Omega \Omega^\top$.)

If this is a mistake in the original MATLAB documentation perhaps we should let the authors know.

jstac:
@crondonm , FYI, following up from our conversation at the CBC, to avoid tiling and hence creating unnecessary arrays, 

``` python
xx = np.tile(x.T, (n, 1))                                                                                                                                             
d = np.sum((S - xx)**2, axis=1)     
```
can be replaced by 

``` python
xx = np.reshape(x, (1, m))                                                                                                                                            
d = np.sum((S - xx)**2, axis=1)     
```

jstac:
Efficiency issues are now solved.  I chose Numba to accelerate the code because it's largely serial.  With jit compilation and some memory savings mentioned above (`reshape` vs `tile`), execution time for the example provided by @crondonm drops from 1 min 13 seconds to 2.65 seconds.

quantecon/discrete_var.py
crondonm: Hi @jstac,

m should be equal to the number of states you are discretizing. In the example, m = 2.

quantecon/discrete_var.py
crondonm: Noted. Thanks for the suggestion.

quantecon/discrete_var.py
jstac: Thanks @crondonm .  In the specified example that's true but for the code to work for arbitrary examples we need to replace 2 by m.  I've taken care of this.

crondonm:
Hi @jstac,

This is fantastic. Thanks a lot for all your work optimizing the code.

Thanks a lot for pointing out the issue with the variance-covariance matrix, 

In fact, the  MATLAB code does compute the variance-covariance as $\Omega \Omega^\top$, which requires the input to be a Cholesky-Decomposition of $\omega$. I missed using the right matrix when producing the simulation. 

Regarding your question about XVEC, I find it helpful to get it as it saves time when you are simulating the full model. However, in fairness,  it is not always needed and it could be an option. I guess the minimum output should be the states and the stochastic matrix.

jstac:
> Regarding your question about XVEC, I find it helpful to get it as it saves time when you are simulating the full model. However, in fairness, it is not always needed and it could be an option. I guess the minimum output should be the states and the stochastic matrix.

Good idea, I'll make that change now.

crondonm:
> Is this OK with you @crondonm ?

Sounds great!

jstac:
OK @crondonm , I've changed the name and also provided options to construct the state space using row major order, as well as column major order.  (The first will be a better fit for most Python code, whereas the second is consistent with the original).

jstac:
This is about ready.  I've added tests within the file so the code just needs to be reviewed and then embedded properly in the library.

I suggest that we add the code to `markov/approximation`.

All help for these last steps is appreciated (CC @Smit-create @mmcky)

Smit-create:
> This is about ready. I've added tests within the file so the code just needs to be reviewed and then embedded properly in the library.

Sure, will try to add the final touch-ups this weekend.

Smit-create:
@jstac I have added the tests and the `discrete_var` in their respective files. I have also added you and @crondonm as co-author of the commits.
@crondonm, Please let me know if the GitHub-id is correct (I found that from your previous commit) or I can change that if you provide me with the one linked to your GitHub-id.

Smit-create:
I think the test failures are due to #633, `pytest` also throws a bunch of warnings which I fixed in #643.

crondonm:
Hi @Smit-create,

The Github-Id is correct. Thank you very much.

Also, my first version included the phrase:

"This code was adapted by Carlos Rondón-Moreno from Schmitt-Grohé and  Uribe's code for MATLAB"

Given the many contributions from @jstac  and the team, could you please change it to:

"This code was adapted from Schmitt-Grohé and  Uribe's original MATLAB code"

Thanks again,

Carlos

jstac:
Many thanks @Smit-create, great work.

@crondonm , @oyamad I propose that we return a `MarkovChain` object here, as we do for `tauchen` and `rouwenhorst`, in line with [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  @crondonm , this means that we should probably drop returning `Xvec`, the simulation data.  Since a VAR can easily be simulated by [LSS](https://quanteconpy.readthedocs.io/en/latest/tools/lss.html), we do not loose much.  

In addition, if we return the Markov chain as a `MarkovChain` object, we can use its `simulate` method to generate time series.  The `simulate` method is fast and efficient.

> Given the many contributions from @jstac and the team, could you please change it to...

Thanks @crondonm .  In the open source world it's typical to omit credits from the file.  But this will always be your PR that got merged into QuantEcon!

crondonm:
Hi @jstac and team,

I was looking at the new code and found that the Transition Matrix it gets is way smaller than with Matlab. 

For the following calibration, 

Θ = np.array([[0.72137370, -0.1323976], [0.0028990, 0.9705325]])
Σ = np.array([[0.00116490, -0.0000131], [-0.0000131, 0.0001328]])
N = np.array([21, 11])

discrete_var.py gets a 75x75 transition matrix.
tpm.py (the original code I submitted) gets a 218x218 matrix
tpm.m (Original Matlab file) gets a 219x219 

The matrix sizes are expected to change with randomness but not by that much. Any idea what could be happening?

Thanks, 

Carlos.

jstac:
Thanks for reporting @crondonm .  I tested and also get a smaller matrix.  I tried different seeds and simulation lengths without much change so it seems to be a bug in the code.  I should be able to take another look over the weekend, hopefully fix the bug and finalize this PR.

quantecon/markov/approximation.py
```diff
@@ -212,3 +214,200 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
...
+
+    if return_sim:
+        return Pi, S, Xvec
+    return Pi, S
+
+
+@njit
+def _run_sim(A, C, Pi, Xvec, S, sim_length, burn_in, seed):
+    m = len(A)
+    np.random.seed(seed)
```
oyamad, line 395: I do not know what the whole code does, but I do not think this is a good strategy.

Maybe better to follow the pattern used throughout the package (see e.g. [this](https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/markov/core.py#L471)) to have the outer function `discrete_var` accept a `random_state`, and
* generate random values within `discrete_var` by `random_state.randn(m, sim_length)` (or something appropriate),
* store them in `Xvec` (after multiplying by `C`?), and
* modify `Xvec` in place in `_run_sim`?

quantecon/markov/approximation.py
```diff
@@ -212,3 +214,200 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
             z = x[j] - rho * x[i]
             P[i, j] = (std_norm_cdf((z + half_step) / sigma) -
                        std_norm_cdf((z - half_step) / sigma))
+
+
+def cartesian_product(arrays, row_major=True):
```
oyamad, line 219: Does this do something different from [`quantecon.cartesian`](https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/gridtools.py#L13)?

quantecon/markov/approximation.py
oyamad: Sorry, I haven't read https://github.com/QuantEcon/QuantEcon.py/pull/640#issuecomment-1295750336. Yes agreed, let an existing object do a simulation.

oyamad: Sorry, I haven't read https://github.com/QuantEcon/QuantEcon.py/pull/640#issuecomment-1295750336. Yes agreed, let an existing object do a simulation.

quantecon/markov/approximation.py
```diff
@@ -212,3 +214,200 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
...
+    sigma_vector = np.sqrt(np.diagonal(Sigma))    # Stationary std dev
+    upper_bounds = std_devs * sigma_vector
+
+    # Build the individual grids along each dimension
+    V = []
+    for i in range(m):
+        b = np.linspace(-upper_bounds[i], upper_bounds[i], grid_sizes[i])
+        V.append(b)
+
+    S = cartesian_product(V, row_major=row_major)
```
oyamad, line 366: Does this part do something different from [mlinspace](https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/gridtools.py#L54)?

jstac:
@Smit-create and @shlff, can either of you finish of this PR?

We need to 

- [ ] drop the option of returning `Xvec` and remove discussion of it from the docs
- [ ] in `_run_sim`, the data for `Xvec` does not need to be stored
- [ ] drop the option to return column major ordered data but add instructions in the docs that tell the user how to convert `Pi` and `S` to this format (check instructions with existing code before deleting option to return column major).
- [ ] investigate whether `quantecon.mlinspace` can be used as discussed by @oyamad 
- [ ] make `discrete_var` return a `MarkovChain` object, similar to `tauchen`.

shlff:
Thanks @jstac and @crondonm . 

Hi @crondonm , I've tried to replicate the issue you mentioned in the [comment above](https://github.com/QuantEcon/QuantEcon.py/pull/640#issuecomment-1301109652) with anaconda installed and the code being run in a Python environment (Python 3.10.4).

However, I couldn't get a similar result as you got (i.e., a shrinking size for the matrix) by implementing the following steps in my terminal:
```
(base) ➜  ~ cd Documents/GitHub/qe/QuantEcon.py
(base) ➜  QuantEcon.py git:(pr/640) git pull
Already up to date.
(base) ➜  QuantEcon.py git:(pr/640) conda activate py310
(py310) ➜  QuantEcon.py git:(pr/640) pip3 install -e.
Obtaining file:///Users/shuhu/Documents/GitHub/qe/QuantEcon.py
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Requirement already satisfied: scipy>=1.0.0 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from quantecon==0.5.3) (1.9.1)
Requirement already satisfied: sympy in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from quantecon==0.5.3) (1.10.1)
Requirement already satisfied: numba in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from quantecon==0.5.3) (0.55.1)
Requirement already satisfied: requests in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from quantecon==0.5.3) (2.28.1)
Requirement already satisfied: numpy in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from quantecon==0.5.3) (1.21.5)
Requirement already satisfied: llvmlite<0.39,>=0.38.0rc1 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from numba->quantecon==0.5.3) (0.38.0)
Requirement already satisfied: setuptools in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from numba->quantecon==0.5.3) (63.4.1)
Requirement already satisfied: charset-normalizer<3,>=2 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from requests->quantecon==0.5.3) (2.0.4)
Requirement already satisfied: urllib3<1.27,>=1.21.1 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from requests->quantecon==0.5.3) (1.26.11)
Requirement already satisfied: idna<4,>=2.5 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from requests->quantecon==0.5.3) (3.3)
Requirement already satisfied: certifi>=2017.4.17 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages (from requests->quantecon==0.5.3) (2022.9.24)
Requirement already satisfied: mpmath>=0.19 in /Users/shuhu/opt/anaconda3/envs/py310/lib/python3.10/site-packages/mpmath-1.2.1-py3.10.egg (from sympy->quantecon==0.5.3) (1.2.1)
Building wheels for collected packages: quantecon
  Building editable for quantecon (pyproject.toml) ... done
  Created wheel for quantecon: filename=quantecon-0.5.3-py3-none-any.whl size=3629 sha256=1d96531cb7e0a3ea2527e32f6c73639112e36ff204741e1be9a646782a194560
  Stored in directory: /private/var/folders/1q/0w60kv813lgc9w3t7x8dgbj00000gn/T/pip-ephem-wheel-cache-fps_m5tg/wheels/ed/76/88/f96b8e4509fead5f42be8f88b92bfb7ac69a5b7b511a93a942
Successfully built quantecon
Installing collected packages: quantecon
  Attempting uninstall: quantecon
    Found existing installation: quantecon 0.5.3
    Uninstalling quantecon-0.5.3:
      Successfully uninstalled quantecon-0.5.3
Successfully installed quantecon-0.5.3
(py310) ➜  QuantEcon.py git:(pr/640) which ipython
/Users/shuhu/opt/anaconda3/envs/py310/bin/ipython
(py310) ➜  QuantEcon.py git:(pr/640) ipython
Python 3.10.4 (main, Mar 31 2022, 03:38:35) [Clang 12.0.0 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: from quantecon.markov import discrete_var

In [3]: T, burn_in = 1_000_000, 100_000

In [4]: Theta = np.array([[0.72137370, -0.1323976], [0.0028990, 0.9705325]])

In [5]: Sigma = np.array([[0.00116490, -0.0000131], [-0.0000131, 0.0001328]])

In [6]: N = np.array([21, 11])

In [7]: Pi, S = discrete_var(Theta, Sigma, N, sim_length=T, burn_in=burn_in, row_major=False)

In [8]: print(len(Pi), len(S))
219 219

In [9]: Pi_r, S_r = discrete_var(Theta, Sigma, N, sim_length=T, burn_in=burn_in, row_major=True)

In [10]: print(len(Pi_r), len(S_r))
219 219

In [11]: 
```

As you can see the sizes of matrices are still 219 by 219.

I've discussed this with @jstac who later got the same result as mine with these steps. 

One possibility could be that probably you and @jstac were using an old commit. Looking forward to your comments.

shlff:
Thanks @jstac and @oyamad . Let ``V`` be a list of two NumPy arrays with lengths of m, n, respectively. 

I've investigated the function [quantecon.gridtools.cartesian(V, order="C")](https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/gridtools.py#L13) can be used equivalently as [quantecon.markov.approximation.cartesian_product(V, row_major=True)](https://github.com/crondonm/QuantEcon.py/blob/2e7526d554534e0f28b125eca1dea4cd26669ee3/quantecon/markov/approximation.py#L219), but ``quantecon.gridtools.cartesian(V, order="F")`` cannot be used equivalently as ``quantecon.markov.approximation.cartesian_product(V, row_major=False)``.

The following code illustrates:
```
In [1]: import numpy as np

In [2]: from quantecon.markov.approximation import cartesian_product

In [3]: from quantecon.gridtools import cartesian

In [4]: m, n = 2, 6

In [5]: np.random.seed(1)

In [6]: V = [np.random.rand(m), np.random.rand(n)]

In [7]: print(V)
[array([0.417022  , 0.72032449]), array([1.14374817e-04, 3.02332573e-01, 1.46755891e-01, 9.23385948e-02,
       1.86260211e-01, 3.45560727e-01])]

In [8]: S = cartesian_product(V, row_major=True)

In [9]: print(S)
[[4.17022005e-01 1.14374817e-04]
 [4.17022005e-01 3.02332573e-01]
 [4.17022005e-01 1.46755891e-01]
 [4.17022005e-01 9.23385948e-02]
 [4.17022005e-01 1.86260211e-01]
 [4.17022005e-01 3.45560727e-01]
 [7.20324493e-01 1.14374817e-04]
 [7.20324493e-01 3.02332573e-01]
 [7.20324493e-01 1.46755891e-01]
 [7.20324493e-01 9.23385948e-02]
 [7.20324493e-01 1.86260211e-01]
 [7.20324493e-01 3.45560727e-01]]

In [10]: S_q = cartesian(V)

In [11]: print(S_q)
[[4.17022005e-01 1.14374817e-04]
 [4.17022005e-01 3.02332573e-01]
 [4.17022005e-01 1.46755891e-01]
 [4.17022005e-01 9.23385948e-02]
 [4.17022005e-01 1.86260211e-01]
 [4.17022005e-01 3.45560727e-01]
 [7.20324493e-01 1.14374817e-04]
 [7.20324493e-01 3.02332573e-01]
 [7.20324493e-01 1.46755891e-01]
 [7.20324493e-01 9.23385948e-02]
 [7.20324493e-01 1.86260211e-01]
 [7.20324493e-01 3.45560727e-01]]

In [12]: S == S_q # check equivalence of row_major
Out[12]: 
array([[ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True]])

In [13]: S_r = cartesian_product(V, row_major=False)

In [14]: print(S_r)
[[4.17022005e-01 1.14374817e-04]
 [7.20324493e-01 1.14374817e-04]
 [4.17022005e-01 9.23385948e-02]
 [7.20324493e-01 9.23385948e-02]
 [4.17022005e-01 1.46755891e-01]
 [7.20324493e-01 1.46755891e-01]
 [4.17022005e-01 1.86260211e-01]
 [7.20324493e-01 1.86260211e-01]
 [4.17022005e-01 3.02332573e-01]
 [7.20324493e-01 3.02332573e-01]
 [4.17022005e-01 3.45560727e-01]
 [7.20324493e-01 3.45560727e-01]]

In [15]: S_rq = cartesian(V, order="F")

In [16]: print(S_rq)
[[4.17022005e-01 1.14374817e-04]
 [7.20324493e-01 1.14374817e-04]
 [4.17022005e-01 3.02332573e-01]
 [7.20324493e-01 3.02332573e-01]
 [4.17022005e-01 1.46755891e-01]
 [7.20324493e-01 1.46755891e-01]
 [4.17022005e-01 9.23385948e-02]
 [7.20324493e-01 9.23385948e-02]
 [4.17022005e-01 1.86260211e-01]
 [7.20324493e-01 1.86260211e-01]
 [4.17022005e-01 3.45560727e-01]
 [7.20324493e-01 3.45560727e-01]]

In [17]: S_r == S_rq
Out[17]: 
array([[ True,  True],
       [ True,  True],
       [ True, False],
       [ True, False],
       [ True,  True],
       [ True,  True],
       [ True, False],
       [ True, False],
       [ True, False],
       [ True, False],
       [ True,  True],
       [ True,  True]])
```
Do you think we should 
- modify the function ``quantecon.gridtools.cartesian()`` to be aligned with ``quantecon.markov.approximation.cartesian_product(V, row_major=False)`` or
- just keep and continue to use ``quantecon.markov.approximation.cartesian_product``?

quantecon/markov/approximation.py
```diff
@@ -299,21 +298,12 @@ def discrete_var(A,
...
-    S : array_like(float)
-        An array where element (i,j) of S is the discretized
-        value of the j-th element of x_t in state i. Reducing S to its
-        unique values yields the grid values.
-    Xvec : array_like(float)
-        A matrix of size m x sim_length containing the simulated time
-        series of the m discretized states.
+    mc : MarkovChain
+        An instance of the MarkovChain class that stores the transition
+        matrix and state values returned by the discretization method
```
jstac, line 306: Hi @shlff , can you please put back information about Pi and S?  The user needs to understand what `S[i,j]` represents.

jstac:
@shlff , thanks for your comments.  I've requested that we "drop the option to return column major ordered data but add instructions in the docs that tell the user how to convert Pi and S to this format (check instructions with existing code before deleting option to return column major)."  Hence, since we are only using row major, it seems that we can switch to `quantecon.markov.approximation.cartesian_product`.

Smit-create:
> Hence, since we are only using row major, it seems that we can switch to `quantecon.markov.approximation.cartesian_product`.

Thanks @shlff and @jstac. @jstac Do you mean dropping the `row_major` argument from `dicrete_var` as well as removing `quantecon.markov.approximation.cartesian_product` and instead re-use `quantecon.gridtools.cartesian`?

jstac:
> Do you mean dropping the row_major argument from dicrete_var as well as removing quantecon.markov.approximation.cartesian_product and instead re-use quantecon.gridtools.cartesian?

Yes, that's right.  If someone can come up with instructions for reordering `Pi` into column major and put them in the docs that would be great too.  

BTW, let's change `Pi` to `P`, which is more consistent with the rest of the libraries Markov code.

Smit-create:
TODO:
- [ ] Add instructions for reordering into column major

@jstac, @shlff, Please review and edit the above TODO list if anything needs to be changed.

quantecon/markov/approximation.py
```diff
@@ -212,3 +214,188 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
...
+
+    if grid_sizes is None:
+        # Set the size of every grid to default_grid_size
+        grid_sizes = np.full(m, default_grid_size)
+
+    n = grid_sizes.prod()  # Size of the discretized state
+
+    # Compute stationary variance-covariance matrix of AR process and use
+    # it to obtain grid bounds.
+    Sigma = qme.solve_discrete_lyapunov(A, Omega)
```
oyamad, line 318: Shouldn't it be
```suggestion
    Sigma = qme.solve_discrete_lyapunov(A, Omega @ Omega.T)
```
?

oyamad:
I had a quick look at the code and the paper. As I understand it, the whole procedure can be structured as follows?

```python
# In the actual implementation, follow the `random_state` approach
v = np.random.multivariate_normal(np.zeros(m), Omega, size=sim_length-1)

# Need to add burn_in option to quantecon.simulate_linear_model
X_vectors = quantecon.simulate_linear_model(A, x0, v, ts_length=sim_length)

X_indices = _find_closest_state_index(tuple_of_grids, X_vectors)

P = _estimate_P(X_indices, num_states)
```

For implementation of `_find_closest_state_index`, the following should be more efficient than the linear search over `S` in the current code, given that each dimension of `S` is sorted:
```python
def _find_closest_state_index(tuple_of_grids, X_vectors):
    # m = length of `vector` and `tuple_of_grids`
    indices_for_each_dimension = np.empty(m, dtype=np.int_)
    for t in range(sim_length):
        for i in range(m):
            idx = np.searchsorted(tuple_of_grids[i], vector[i])
            # Then determine which of `idx-1` or `idx` is closer
            # and store it in `indices_for_each_dimension[i]`.

        # Determine the index in the cartesian product,
        # depending on how the state vectors are ordered.
        X_indices[t] = _idx_cartesian(indices_for_each_dimension, nums_grids)

    return X_indices
```

shlff:
> TODO:
> 
> * [ ]  Add instructions for reordering into column major
> 
> @jstac, @shlff, Please review and edit the above TODO list if anything needs to be changed.

Thanks @Smit-create . Nice work!

Hi @jstac I've reviewed @oyamad's [this comment](https://github.com/QuantEcon/QuantEcon.py/pull/640#discussion_r1012447782) again and find the following code in the function ``discrete_var``:
```
from quantecon.gridtools import cartesian
V = []
for i in range(m):
      b = np.linspace(-upper_bounds[i], upper_bounds[i], grid_sizes[i])
      V.append(b)
S = cartesian(V)
```
can be equivalently replaced by
```
from quantecon.gridtools import mlinspace
S = mlinspace(-upper_bounds, upper_bounds, grid_sizes)
```
Thus the [above commit](https://github.com/QuantEcon/QuantEcon.py/pull/640/commits/740cf739b884124afd182cf8f792fb7d00ac2801) adds this modification.

The following code illustrates:
```
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: from quantecon.gridtools import cartesian

In [3]: from quantecon import matrix_eqn as qme

In [4]: from quantecon.gridtools import mlinspace

In [5]: A = np.array([[0.72137370, -0.1323976], [0.0028990, 0.9705325]])
   ...: Omega = np.array([[0.00116490, -0.0000131], [-0.0000131, 0.0001328]])
   ...: grid_sizes = np.array([21, 11])
   ...: std_devs = np.sqrt(10)

In [6]: m = len(A)   # The number of dimensions of the original state x_t
   ...: default_grid_size = 10
   ...: 
   ...: if grid_sizes is None:
   ...:     # Set the size of every grid to default_grid_size
   ...:     grid_sizes = np.full(m, default_grid_size)
   ...: 
   ...: n = grid_sizes.prod()  # Size of the discretized state
   ...: 
   ...: # Compute stationary variance-covariance matrix of AR process and use
   ...: # it to obtain grid bounds.
   ...: Sigma = qme.solve_discrete_lyapunov(A, Omega)
   ...: sigma_vector = np.sqrt(np.diagonal(Sigma))    # Stationary std dev
   ...: upper_bounds = std_devs * sigma_vector

In [7]: # Original code: Build the individual grids along each dimension
   ...: V = []
   ...: for i in range(m):
   ...:     b = np.linspace(-upper_bounds[i], upper_bounds[i], grid_sizes[i])
   ...:     V.append(b)
   ...: S_r = cartesian(V)

In [8]: # Use QuantEcon's mlinspace
   ...: S_rq = mlinspace(-upper_bounds, upper_bounds, grid_sizes)

In [9]: S_r == S_rq
Out[9]: 
array([[ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True],
       [ True,  True]])

```

shlff:
> 

Hi @jstac and @Smit-create , I figured out a possible method of converting the state values S from row-major to column-major ordering **ex post**.

Let ``S_r`` be the row-major state values as given and ``S_c`` be the column-major state values. The following python code does the conversion job:
```
n = 1                                                                                              # choose column index
cols = len(set((S_r[:, n])))                                                           # count the number of columns
S_c = np.concatenate([S_r[i::cols] for i in range(cols)])          # rearrange the order
```
In the following test I verify the result ``S_c`` from the above code with that generated from [quantecon.gridtools.mlinspace](https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/gridtools.py#L54) with option ``order="F"``:
```
In [1]: import numpy as np
   ...: from quantecon import matrix_eqn as qme
   ...: from quantecon.gridtools import mlinspace

In [2]: A = np.array([[0.72137370, -0.1323976], [0.0028990, 0.9705325]])
   ...: Omega = np.array([[0.00116490, -0.0000131], [-0.0000131, 0.0001328]])
   ...: grid_sizes = np.array([21, 11])
   ...: std_devs = np.sqrt(10)
   ...: seed=1234
   ...: sim_length=1_000_000
   ...: burn_in=100_000

In [3]: m = len(A)   # The number of dimensions of the original state x_t
   ...: default_grid_size = 10
   ...: 
   ...: if grid_sizes is None:
   ...:     # Set the size of every grid to default_grid_size
   ...:     grid_sizes = np.full(m, default_grid_size)
   ...: 
   ...: n = grid_sizes.prod()  # Size of the discretized state
   ...: 
   ...: # Compute stationary variance-covariance matrix of AR process and use
   ...: # it to obtain grid bounds.
   ...: Sigma = qme.solve_discrete_lyapunov(A, Omega)
   ...: sigma_vector = np.sqrt(np.diagonal(Sigma))    # Stationary std dev
   ...: upper_bounds = std_devs * sigma_vector

In [4]: # Build the individual grids along each dimension
   ...: S_r = mlinspace(-upper_bounds, upper_bounds, grid_sizes, order="C")  # row major ordering
   ...: S_c = mlinspace(-upper_bounds, upper_bounds, grid_sizes, order="F")  # column major ordering

In [5]: n = 1 # column index
   ...: cols = len(set((S_r[:, n])))
   ...: S_c_test = np.concatenate([S_r[i::cols] for i in range(cols)])

In [6]: S_c_test == S_c
Out[6]: 
array([[ True,  True],
       [ True,  True],
       ...
       (**same as above**)])

```

crondonm:
Hi @jstac and team,

Thanks a lot for all your work on this issue. 

Do you have an estimate for merging this PR with the main branch?

We are working on new python codes, and it would be great to use discrete_var for those instead of my clunky-old routine. 

Thanks again,

Carlos

jstac:
Hi @crondonm , my apologies --- this is waiting on me and I've been flat out trying to finish the DP text with Tom Sargent.  My brain is deep in the proofs and cannot think about anything else.  I'm just wrapping them up now and I hope to have this PR complete over the w/e or Monday at the latest.

jstac:
@crondonm Sorry for slow progress.  I'm getting back to this now and I'll finish it up over the next day or two.  I separated out part of the problem in https://github.com/QuantEcon/QuantEcon.py/pull/658

jstac:
Hi @crondonm @oyamad @shlff @Smit-create , looking to finish this up, and in the spirit of suggestions by @oyamad , I suggest the following:

1. Complete and merge #658 #660 
2. Add a function `discretize_markov_process` that takes an `arbitrary` series $X = (X_t)_{t=0}^{T-1}$ in $\mathbb R^n$ plus a set of grid points in each dimension and converts it to a `MarkovChain` by first applying discretization onto the grid via `cartesian_nearest_index` and then estimation of the Markov chain via `estimate_mc_discrete`.
3. Add a function `discretize_gaussian_var` which plays the role that @crondonm originally suggested but uses the method proposed by @oyamad : take in parameters and then do as follows:

```python
# In the actual implementation, follow the `random_state` approach
v = np.random.multivariate_normal(np.zeros(m), Omega, size=sim_length-1)
X_vectors = quantecon.simulate_linear_model(A, x0, v, ts_length=sim_length)
mc = discretize_markov_process(X_vectors, grids)
```

I think we don't need burn-in, since ergodic theory tells us that the whole time series is useful.  But if we want to go the extra step, we can generate the Gaussian VAR by first computing the stationary distribution and then starting the sample from that distribution.

oyamad:
A few comments:

(1)
I feel a function called `discretize_markov_process` should take an object that represents a continuous Markov process (for example, an instance of a class, say, `AR1` which holds `A` and `Omega` as attributes), rather than a series of finitely many observations. Call it `fit_discrete_mc` or something else?

jstac:
> Call it fit_discrete_mc or something else

Yes, that's better.

oyamad:
(2)
I didn't get the point of deviating from the paper's formulation `x_t = A x_{t-1} + Omega u_t`, where `u` can be any `r`-dimensional random variable with mean `O` and standard deviation `I`. As clearly noted here https://github.com/QuantEcon/QuantEcon.py/blob/740cf739b884124afd182cf8f792fb7d00ac2801/quantecon/markov/approximation.py#L281-L286 normality is not important.

At the moment it is fine to stick to normal, but in the future we can extend the code to accept any distribution, say from `scipy.stats`, for `u`. It would look like:

```py
def discrete_var(A, Omega, ..., rv=None, random_state=None):
    random_state = check_random_state(random_state)
    m, r = Omega.shape

    if rv is None:
        u = random_state.multivariate_normal(
            np.zeros(r), np.identity(r), size=sim_length-1
        )
    else:
        u = rv.rvs(size=sim_length-1, random_state=random_state)

    v = Omega @ u.T
    X_vectors = quantecon.simulate_linear_model(A, x0, v, ts_length=sim_length)
    ...
```

oyamad:
(3)
The function name should be something that shows that it is based on a simulation approach. (There is a [function](https://quantecon.github.io/QuantEcon.jl/latest/api/QuantEcon.html#QuantEcon.discrete_var) with the same name `discrete_var` but using a different approach in QuantEcon.jl.) I don't have a good idea though (`discretize_var_by_monte_carlo` would be too long...).

Another possibility is to add a `method` argument: `discretize_var(..., method='monte carlo')`.

oyamad:
Let me elaborate on my comment (2) above.

`rv` (if not None) is supposed to be an instance of a [multivariate random variable](https://docs.scipy.org/doc/scipy/reference/stats.html#multivariate-distributions) from `scipy.stats`. All classes there have a `rvs` method which has a `size` argument, but not necessarily has a `cov` argument. The assumption is that it is user's responsibility in creating such an object to adjust the parameters so that the random variable has mean 0 and std dev 1, and then the code internally computes `Omega @ u` (where `u` is the output of `rv.rvs(...)`). So it is important to take `Omega` as in the paper, rather than `Omega @ Omega.T`.

jstac:
Let me try to summarize the above discussion and suggest what I think we need.

First, we already have a function for estimating a finite Markov chain from *discrete state data* in https://github.com/QuantEcon/QuantEcon.py/pull/658

What we now need is

1. function `fit_discrete_mc`  that takes an arbitrary series $(X_t)_{t=0}^{T-1}$ in $\mathbb R^n$ plus a set of grid points in each dimension and converts it to a `MarkovChain` by first applying discretization onto the grid via `cartesian_nearest_index` and then estimation of the Markov chain via #658 
2. a function that produces a finite Markov chain from a VAR model.  

For (2), I think `discretize_var` is fine as a name and we don't need to mention Monte Carlo in the name.  I agree that we can keep the word `gaussian` out of it but I propose that, for now, we restrict the function to the Gaussian case, finish it and merge it.  (I know that @crondonm is hoping to use it soon and I would like to as well.)

Then, we open an issue on extending `discretize_var` to the non-Gaussian case.

oyamad:
A remaining choice is about the formulation of VAR:

1. The current code:
   `x_t = A x_{t-1} + u_t` where `u` has mean `O` and cov `Omega`.
2. The paper:
   `x_t = A x_{t-1} + Omega u_t` where `u` has mean `O` and cov `I`.

For extendability (if we use the `rvs` method of the objects from `scipy.stats`), I think we should follow the paper's formulation (and it is also consistent with [LinearStateSpace](https://quanteconpy.readthedocs.io/en/latest/tools/lss.html#quantecon.lss.LinearStateSpace)).

oyamad:
In the future, we might introduce a class `VAR`:

```py
class VAR:
    def __init__(self, A, Omega, rv=None):
        ...

    def simulate(self, x0, sim_length, random_state=None):
        # Call `simulate_linear_model`

    def discretize(self, grid_sizes, std_devs, sim_length, random_state=None):
        # Call `discretize_var`
```

jstac:
> For extendability (if we use the rvs method of the objects from scipy.stats), I think we should follow the paper's formulation

I fully agree.

jstac:
> In the future, we might introduce a class `VAR`:
> 
> ```python
> class VAR:
>     def __init__(self, A, Omega, rv=None):
>         ...
> 
>     def simulate(self, x0, sim_length, random_state=None):
>         # Call `simulate_linear_model`
> 
>     def discretize(self, grid_sizes, std_devs, sim_length, random_state=None):
>         # Call `discretize_var`
> ```

Agreed.  Some people might prefer `x_t = A x_{t-1} + b + Omega u_t`, for added convenience.

crondonm:
Hi @jstac and team,

Hopefully, you are having a wonderful holiday season. I just wanted to follow up on this issue. Is it close to being finished? 

If you have time, can you explain to me how to use it after all the changes? I lost track of the original codes after all the modifications.

jstac:
@crondonm Apologies again for the delay --- trying to do too many things.

I'm on holidays with my family right now but to push this forward I've opened issue #678 .   

There's not much left to do with this issue except pull all the parts together.

oyamad:
It should be straightforward now that every component has been prepared in `quantecon`: here 745c0d447a95277aadbbc66c50a6dc9656bafe36 is how I would implement using functions from `quantecon`.

jstac:
Thanks for the reminder @crondonm .  This is now essentially done in https://github.com/QuantEcon/QuantEcon.py/commit/745c0d447a95277aadbbc66c50a6dc9656bafe36

@oyamad That looks perfect to me.  I propose that we go ahead with it.  

At the same time, I suggest opening an issue that recommends adjusting the code to call `fit_discrete_mc` from #678 once it is merged.

oyamad:
For discussion: in 745c0d447a95277aadbbc66c50a6dc9656bafe36,
* The process is formulated as `x_t = A x_{t-1} + C u_t`, where the matrix for the noise term is denoted as `C` to be consistent with [`LinearStateSpace`](https://quanteconpy.readthedocs.io/en/latest/tools/lss.html#quantecon.lss.LinearStateSpace).
* `burn_in` has been dropped. But this may affect the size of the returned `mc`.

What remain are:
* [ ] Adjust the docstring
* [ ] Adjust the tests

oyamad:
Now f4ebd6f89b949a3c917204c58fc2ede5d7a781a6 in branch [VAR-approximation-2](https://github.com/QuantEcon/QuantEcon.py/compare/VAR-approximation-2) uses `fit_discrete_mc` (#681).

jstac:
Many thanks @oyamad .  Just to clarify,  outstanding tasks are as follows?

> 
>     * [ ]  Adjust the docstring
>     * [ ]  Adjust the tests

In terms of adjusting the docstring, the issue is this one?

> * The process is formulated as `x_t = A x_{t-1} + C u_t`, where the matrix for the noise term is denoted as `C` to be consistent with [`LinearStateSpace`](https://quanteconpy.readthedocs.io/en/latest/tools/lss.html#quantecon.lss.LinearStateSpace).

CC @HumphreyYang @Smit-create

oyamad:
Yes, that's one thing.

List of adjustments needed:
* [ ] Formulation `x_t = A x_{t-1} + C u_t`
* Arguments
  - [ ] `burn_in` removed
  - [ ] `rv`added
  - [ ] `order` added
  - [ ] `random_state` added
* [ ] `Notes` section has become irrelevant
* [ ] Update `Examples` section

HumphreyYang:
Hi @oyamad 

> For discussion: in [745c0d4](https://github.com/QuantEcon/QuantEcon.py/commit/745c0d447a95277aadbbc66c50a6dc9656bafe36),
> 
> * The process is formulated as `x_t = A x_{t-1} + C u_t`, where the matrix for the noise term is denoted as `C` to be consistent with [`LinearStateSpace`](https://quanteconpy.readthedocs.io/en/latest/tools/lss.html#quantecon.lss.LinearStateSpace).

I am changing the documentation, but I did not find a detailed discussion on the use `C` instead of `Omega` given `C = sp.linalg.sqrtm(Omega)` in the chat above. I found it slightly anti-intuitive since $\Omega  u_t$ feels more in line with [Downward Nominal Wage Rigidity, Currency Pegs, and Involuntary Unemployment](https://www.journals.uchicago.edu/doi/10.1086/688175) in our example, as $\Omega$ is scaling the $u_t$ term which should be drawn from distributions with mean 0 and std 1 for this technique to work. Could you please kindly let me know the rationale for the change so that I can have them recorded in the documentation :) ?


> * `burn_in` has been dropped. But this may affect the size of the returned `mc`.

I have used `C = sp.linalg.sqrtm(Omega)` to adjust the test. The length does not change and the values are aligned with the previous implementation for `state_values`. However, `P` is different, this may be due to the change in seed and burn-in since we are using RandomState now:

```
>       assert_allclose(mc.P, self.P_out, atol = 1e-6, err_msg = f"P = {mc.P} \n should be {self.P_out}")
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-06
E       P = [[1.61290323e-02 1.12903226e-01 0.00000000e+00 3.70967742e-01
E         5.00000000e-01 0.00000000e+00]
E        [1.00964548e-04 8.51048124e-01 3.82857566e-02 4.03858192e-04
E         1.10111936e-01 4.93604457e-05]
E        [0.00000000e+00 3.02295449e-01 6.97266822e-01 0.00000000e+00
E         3.85201268e-04 5.25274456e-05]
E        [3.60600761e-05 4.86811027e-04 0.00000000e+00 6.97473992e-01
E         3.02003137e-01 0.00000000e+00]
E        [3.17039037e-05 1.11090478e-01 4.55177474e-04 3.75374219e-02
E         8.50778784e-01 1.06434534e-04]
E        [0.00000000e+00 4.45945946e-01 3.37837838e-01 0.00000000e+00
E         1.89189189e-01 2.70270270e-02]] 
E        should be [[8.06451613e-02 1.93548387e-01 0.00000000e+00 3.54838710e-01
E         3.70967742e-01 0.00000000e+00]
E        [8.39514352e-05 8.50806955e-01 3.79097454e-02 4.96901738e-04
E         1.10647992e-01 5.44549850e-05]
E        [0.00000000e+00 3.01394785e-01 6.97927443e-01 0.00000000e+00
E         5.70755895e-04 1.07016730e-04]
E        [1.03646634e-04 5.52782048e-04 0.00000000e+00 6.99407487e-01
E         2.99936085e-01 0.00000000e+00]
E        [3.14483775e-05 1.09581871e-01 4.53755161e-04 3.85467256e-02
E         8.51289608e-01 9.65914451e-05]
E        [0.00000000e+00 3.63636364e-01 3.37662338e-01 0.00000000e+00
E         2.46753247e-01 5.19480519e-02]]
E       Mismatched elements: 27 / 36 (75%)
E       Max absolute difference: 0.12903226
E       Max relative difference: 0.8
```

oyamad:
Thanks @HumphreyYang

* Sorry, I didn't notice the documentation entry https://quanteconpy.readthedocs.io/en/latest/tools/lss.html#quantecon.lss.LinearStateSpace is empty. (I thought I saw something there...) @Smit-create Do you see why it is empty?
  So refer to the source code https://github.com/QuantEcon/QuantEcon.py/blob/15aa789198f28969c2e7aea005720168a244188d/quantecon/_lss.py#L67-L81 or the corresponding lecture: https://python.quantecon.org/linear_models.html. A VAR process can be considered as a part of a `LinearStateSpace` which is denoted by `x_{t+1} = A x_t + C w_{t+1}` in the above (we could use `w` in place of `u`).
* It is a general question, what should be tested against code that involves randomness. I know what to test in this case. For example, that two calls with a same seed return the exactly same outcome?

HumphreyYang:
Many thanks for the information about the lecture @oyamad. I have pushed a new commit [here](https://github.com/HumphreyYang/QuantEcon.py/commit/bb918643f5d9e681ff63795b8886da82f8eb4227) due to lack of permission. Since all the literature we referred to used the variance-covariance matrix, I kept Omega in the example, which indicates to readers how to use this function with a variance-covariance matrix. Please kindly let me know if anything needs to be changed. 

> * It is a general question, what should be tested against code that involves randomness. I know what to test in this case. For example, that two calls with a same seed return the exactly same outcome?

I had seeds set up when generating the result, but the issue is slightly subtle: it is caused by the precision of the decimal I had in the test case before, and the small difference is exaggerated across states. 

Many thanks in advance.

crondonm:
Dear all,

Given the recent commit, is this ready to merge with the main branch? 

Thanks,

Carlos

jstac:
Many thanks @HumphreyYang for pushing this forward.  It looks good to me so I'll hand over to @oyamad or @mmcky .

jstac:
Hi @oyamad and @mmcky , I know @crondonm  is hoping to use this for research soon.  Please let me know if there's something I can do to help expedite.

quantecon/markov/approximation.py
```diff
@@ -212,3 +215,154 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
             z = x[j] - rho * x[i]
             P[i, j] = (std_norm_cdf((z + half_step) / sigma) -
                        std_norm_cdf((z - half_step) / sigma))
+
+
+def discrete_var(A,
```
mmcky, line 220: Should this function be named `disrete_VAR` given `var` is an acronym for Vector Auto Regression (VAR). `PEP8` has a blanket rule for lower case function names so may not be a great idea. `discrete_var` just reminds me of `discrete variable`.

mmcky:
> In the future, we might introduce a class `VAR`:
> 
> ```python
> class VAR:
>     def __init__(self, A, Omega, rv=None):
>         ...
> 
>     def simulate(self, x0, sim_length, random_state=None):
>         # Call `simulate_linear_model`
> 
>     def discretize(self, grid_sizes, std_devs, sim_length, random_state=None):
>         # Call `discretize_var`
> ```

I really like this idea @oyamad

I am happy to merge this once the conflicts are resolved but I think moving to a `VAR` class would be ideal.

mmcky:
> I have pushed a new commit [here](https://github.com/HumphreyYang/QuantEcon.py/commit/bb918643f5d9e681ff63795b8886da82f8eb4227) due to lack of permission.

@HumphreyYang I have given you access to update this PR directly. 

Do we need to update this PR from the `main` branch to incorporate?

mmcky:
thanks @crondonm for this contribution. 

@oyamad @HumphreyYang if you have time would you mind updating the check boxes to indicate which have been completed. That would be great.

HumphreyYang:
Hi @mmcky, 

I have merged the two branches, there should be no conflict at this point now. I will update the function name now.

mmcky:
Thanks @HumphreyYang 

> I will update the function name now.

Please don't update this yet. This is for discussion. I would mainly be interested in thoughts around naming the function. It is currently compliant with PEP8 and we need a very good reason to deviate from it.

jstac:
I can definitely see @mmcky 's point but my preference is to stick with PEP8

oyamad:
Editing the docstring has almost been done in branch [`VAR-approximation-2`](https://github.com/QuantEcon/QuantEcon.py/compare/main...VAR-approximation-2).

One question about the contribution of Schmitt-Grohé and Uribe's note. I am totally ignorant of the literature, but I guess what is non-trivial in this approach is how to choose the discrete set of points for approximation. Is the following description appropriate?
https://github.com/QuantEcon/QuantEcon.py/blob/03e1a8ddcedcfeabe1df9ee31c651942404a9d23/quantecon/markov/approximation.py#L278-L279

oyamad:
For tests, we need to add a test with `order='F'` and another with a `scipy.stats` multivariate distribution (`multivariate_normal` would be fine) for the `rv` argument and a non-square matrix for the `C` argument.

A question to @HumphreyYang: where are these "Expected outputs" https://github.com/QuantEcon/QuantEcon.py/blob/03e1a8ddcedcfeabe1df9ee31c651942404a9d23/quantecon/markov/tests/test_approximation.py#L123-L142 coming from?

oyamad:
Another question: `A` having eigenvalues smaller than one (in absolute values) is a sufficient condition (for `solve_discrete_lyapunov` to return a correct solution), but is it also necessary? Is this claim https://github.com/QuantEcon/QuantEcon.py/blob/03e1a8ddcedcfeabe1df9ee31c651942404a9d23/quantecon/markov/approximation.py#L285 too strong? Would it better to say more weakly "Its eigenvalues are assumed to have moduli bounded by unity", as in in the docstring of  `solve_discrete_lyapunov` https://github.com/QuantEcon/QuantEcon.py/blob/67b9a41a936cb39704e6eabd1c05118c3f030efa/quantecon/_matrix_eqn.py#L43-L46 ?

HumphreyYang:
> For tests, we need to add a test with `order='F'` and another with a `scipy.stats` multivariate distribution (`multivariate_normal` would be fine) for the `rv` argument and a non-square matrix for the `C` argument.
> 
> A question to @HumphreyYang: where are these "Expected outputs"
> 
> https://github.com/QuantEcon/QuantEcon.py/blob/03e1a8ddcedcfeabe1df9ee31c651942404a9d23/quantecon/markov/tests/test_approximation.py#L123-L142
> 
> coming from?

Hi @oyamad, 

Thanks for asking. The results are generated with the assumption that the previous code is correct. I modified the previous code so that it takes the same random state as our new implementation. After the modification, under the same random state seed, the results generated by the two implementations are the same for `S_out` and the error margin is within 2e-05 for `P_out`:

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 2 / 36 (5.56%)
E       Max absolute difference: 2.2438475e-06
E       Max relative difference: 2.03758815e-05
E        x: array([[1.612903e-02, 1.129032e-01, 0.000000e+00, 3.709677e-01,
E               5.000000e-01, 0.000000e+00],
E              [1.009645e-04, 8.510504e-01, 3.828576e-02, 4.038582e-04,...
E        y: array([[1.612903e-02, 1.129032e-01, 0.000000e+00, 3.709677e-01,
E               5.000000e-01, 0.000000e+00],
E              [1.009645e-04, 8.510481e-01, 3.828576e-02, 4.038582e-04,...

test_approximation.py:156: AssertionError
```

Given the magnitude of the error and only two elements are classified as mismatches, I assume the error is caused by imprecise computation in the transformations.

Therefore, the expected output is generated predicated on the two assumptions: 
1. the previous test cases are correct;
2. the marginal error is not caused by incorrect code.

crondonm:
Given the rest of the description. It seems to me that the last sentence is redundant:
 
https://github.com/QuantEcon/QuantEcon.py/blob/03e1a8ddcedcfeabe1df9ee31c651942404a9d23/quantecon/markov/approximation.py#L278-L279

mmcky:
@oyamad the test added here is coming from the file `tpm_test.ipynb` that was included in the original `zip` file at the top of this PR. I am happy to accept that as the source of truth for the test -- as it aligns with @crondonm submission. Do you agree?

jstac:
> @oyamad the test added here is coming from the file `tpm_test.ipynb` that was included in the original `zip` file at the top of this PR. I am happy to accept that as the source of truth for the test -- as it aligns with @crondonm submission. Do you agree?

It's not easy to find other tests for this function.

oyamad:
@mmcky Yes.

Remaining are:
* [ ] Add a test with with `order='F'`.
* [ ] Add a test with a `scipy.stats` multivariate distribution (`multivariate_normal` would be fine) for the `rv` argument.
* [ ] Add a test with a non-square matrix for the `C` argument.
* Wording in the docstring:
  * [ ] https://github.com/QuantEcon/QuantEcon.py/pull/640#issuecomment-1401375652
  * [ ] https://github.com/QuantEcon/QuantEcon.py/pull/640#issuecomment-1401386389

HumphreyYang:
Hi @mmcky,

CI on the windows platform has some issues with `pip`. Could you please have a quick look when you have time?

Many thanks in advance.

HumphreyYang:
Hi @oyamad,

Many thanks for editing the docstring. It is a wonderful exemplar for me to follow in the future. 

I have added the three tests and fixed a small bug in the code while testing. For the record, tests with `order='F'` and a `scipy.stats` multivariate distribution are relatively straightforward. However, a test for a non-square `C` is tricky, given random samples of `u` will change when the size `r` is changed under the same seed. My approach is to hold the first two rows of `u` unchanged as before and add a new row at the end and check if it returns the same solution as before. 

```python
u = random_state.standard_normal(size=(sim_length-1, 2))
u =  np.insert(u, 2, u[:, 0], axis=1)
```

The test result is the same as the previous `P_out`, which indicates the change in `P` results from the changes in random`u`. Therefore, I conclude that the code is correct and add the result as a test case. 

Could you please kindly review these changes when you are available?

Many thanks in advance.

quantecon/markov/approximation.py
```diff
@@ -332,7 +334,7 @@ def discrete_var(A,
     if rv is None:
         u = random_state.standard_normal(size=(sim_length-1, r))
     else:
-        u = rv.rvs(size=sim_length-1, random_state=random_state)
+        u = rv.rvs(size=(sim_length-1, r), random_state=random_state)
```
oyamad, line 381: My assumption was that `rv` is a "frozen" object (if it comes from `scipy.stats`).

```py
r = 3
rv = multivariate_normal(cov=np.identity(r))
rv.rvs(size=5)
```

```
array([[ 0.6345146 ,  1.24005431,  0.49748714],
       [ 0.70736992, -0.98265265,  1.62881673],
       [-1.46650587, -1.30473345, -0.16520163],
       [-1.51102623, -0.10933109, -0.94159965],
       [ 0.29775017, -0.35714403,  0.28893599]])
```

```py
type(rv)
```

```
scipy.stats._multivariate.multivariate_normal_frozen
```

I wasn't aware that the classes had an `rvs` method. But for some classes, the mean and cov are hard to control?

Maybe we should add an example.

HumphreyYang: Many thanks @oyamad for letting me know.

I have adjusted the code and the test to fit into the convention.

I directly used your example to show how to use the `rv` parameter as I think it would be slightly redundant if we have another example.

Please kindly let me know what you think about the current changes, and if it is needed to have another example.

oyamad: Thanks, but I guess the case with `rv` shouldn't be the main example. (Passing `scipy.stats.multivariate_normal` is redundant.) What would be a "primary" example for the noise term other than normal?

jstac: I don't think there's an obvious primary example aside from normal.  But how https://en.wikipedia.org/wiki/Multivariate_t-distribution for an example?

HumphreyYang: Many thanks @oyamad and @jstac for your suggestions. 

I just updated the docstring to include an example using multivariate t-distribution. 

Please kindly let me know if it needs any changes.

quantecon/markov/approximation.py
```diff
@@ -368,6 +366,39 @@ def discrete_var(A,
            [ 0.15422567, -0.03232648],
            [ 0.15422567, -0.03232648],
            [ 0.19278209, -0.03232648]])
+
+    The simulation below uses the same parameters with :math:`{u_t}`
+    drawn from a multivariate t-distribution
+
+    >>> mc = discrete_var(A, C, grid_sizes, 
+    ...               rv=scipy.stats.multivariate_t(shape=np.identity(2)),
```
oyamad, line 374: Thanks @HumphreyYang, but it is the covariance matrix that must be the identity matrix.

HumphreyYang: Many thanks @oyamad. Sorry I forgot the constraint on the covariance matrix.

Do you think we should add these lines below into the docstring as well, in addition to my update to emphasize our constraint on the distribution?

```python
>>> df = 100
>>> Sigma = np.diag(np.tile((df-2)/df, 2))
>>> t_dis = scipy.stats.multivariate_t(shape=Sigma, df=df).rvs(10_000_000)
>>> np.round(np.cov(t_dis.T), 2)
array([[1., 0.],
       [0., 1.]])
```

oyamad:
> Do you think we should add these lines below into the docstring as well, in addition to my update to emphasize our constraint on the distribution?

No, I think the update in https://github.com/QuantEcon/QuantEcon.py/pull/640/commits/a85a6cd626082bc1c36e7f42f050139e592ece36 is just fine. Thanks!

oyamad:
@HumphreyYang Will you cherry-pick the minor chnages in 496942dff0d9c670d382fff8183c94588c707055? Thanks!

quantecon/markov/approximation.py
```diff
@@ -242,3 +245,198 @@ def _fill_tauchen(x, P, n, rho, sigma, half_step):
...
+
+            `P` : ndarray(float, ndim=2)
+                A 2-dim array containing the transition probability
+                matrix over the discretized states.
+
+            `state_values` : ndarray(float, ndim=2)
+                A 2-dim array containing the state vectors (of dimension
+                m) as rows, which are ordered according to the `order`
+                option.
+
```
HumphreyYang, line 329: Many thanks @oyamad. I think they are in the current branch after your merge. Please correct me if I am wrong.

oyamad (approved):
Tests all green. I think this is ready to merge.

jstac:
Many thanks @oyamad, @crondonm, @HumphreyYang and all other contributers.  @mmcky , I think this is ready to merge.

mmcky:
thanks everyone for your input on this. LGTM.

PR #448, opened 2018-12-09
Travis: Add macOS to the build matrix

rht:
This fixes #359.

PR #645, opened 2022-11-02
CI: Fix time errors on windows and numpy warnings

### LINKED ISSUE #647: CI: NumPy warnings
```
quantecon/tests/test_lqnash.py::TestLQNash::test_nnash
  /home/runner/work/QuantEcon.py/QuantEcon.py/quantecon/tests/test_lqnash.py:96: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    f1_ml = np.asarray(np.matrix("""\

quantecon/tests/test_lqnash.py::TestLQNash::test_nnash
  /home/runner/work/QuantEcon.py/QuantEcon.py/quantecon/tests/test_lqnash.py:100: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    f2_ml = np.asarray(np.matrix("""\
```

Fix: https://github.com/QuantEcon/QuantEcon.py/pull/645/commits/9c62de59b6c4d3f4a9a33e63478c87bd01e2723a

Smit-create:
#633 
Fixes #647


Smit-create:
I think `pytest` is slower on `windows` and so the results differ much on it. We can set a higher `rtol` for `windows` and lower for the other two.

oyamad (commented):
I vote for setting a higher `rtol' for all platforms (not skipping on particular platforms).

quantecon/util/tests/test_timing.py
```diff
@@ -50,10 +55,15 @@ def test_function_two_arg(n, a):
             loop_timer(5, test_function_one_arg, self.h, digits=10)
         test_two_arg = \
             loop_timer(5, test_function_two_arg, [self.h, 1], digits=10)
+
+        tol = 0.01
```
oyamad, line 59: Maybe better to use a relative tol, like `assert_allclose(actual, desired, rtol=rtol)`, rather than a hard-coded absolute tol?

Smit-create:
> Can we remove these parts?
> 
> https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/util/tests/test_timing.py#L18-L20
> 
> https://github.com/QuantEcon/QuantEcon.py/blob/89a70127b1321444b8f22ba262c7401f4b08e9c4/quantecon/util/tests/test_timing.py#L39-L41

That issue with osx is still not fixed. I tried enabling that and found these results: https://github.com/QuantEcon/QuantEcon.py/actions/runs/3374887986/jobs/5600990154:
```
----------------------------- Captured stdout call -----------------------------
TAC: Elapsed: 0:00:0.11
TAC: Elapsed: 0:00:0.19
TOC: Elapsed: 0:00:0.48
```
While the expected values are `0.1`, `0.1`, and `0.3`. See the relevant discussion here: https://github.com/QuantEcon/QuantEcon.py/pull/448.

Locally, this passes for me on MAC M1 but not sure why that fails on the CI.

oyamad:
Just raise `rtol`, say to `0.3`, so that the tests pass? I think it is enough to test that the code itself runs without an error. We do not know what value to "expect" for `time.sleep(self.h)` + overhead time.

oyamad:
Then let `rtol = 2` for example?

oyamad (approved):
@Smit-create Great, thanks!

IanMichaelHarper:
> I think `pytest` is slower on `windows` and so the results differ much on it. We can set a higher `rtol` for `windows` and lower for the other two. [prjemian/punx#167](https://github.com/prjemian/punx/issues/167)

Is there a reason it's slower and can it be fixed?

PR #649, opened 2022-11-03
[MAINT] TST: Rename `tearDown` to `teardown_method` (pytest fixes)

Smit-create:
docs: [teardown_method](https://docs.pytest.org/en/6.2.x/xunit_setup.html#method-and-function-level-setup-teardown)


oyamad (approved):
@Smit-create Thanks, looks good to me.

PR #654, opened 2022-11-23
ENH: check_random_state: Accept `np.random.Generator`

oyamad:
Numpy's recommended procedure to generate random numbers now seems to be to use `np.random.Generator` instead of `np.random.RandomState`: https://numpy.org/doc/stable/reference/random/index.html

This PR is to update `check_random_state` to accept `np.random.Generator`, following the corresponding code in scipy:
https://github.com/scipy/scipy/blob/b4a970bf4eba1065ea31f8d0988cd4ec523a07d4/scipy/_lib/_util.py#L174-L203

There are discussions on this at scipy and scikit-learn:
* scipy/scipy#13778
* scikit-learn/scikit-learn#16988
* scikit-learn/scikit-learn#22271
* scikit-learn/scikit-learn#23962

The issue is that some methods are present for `np.random.RandomState` but not for `np.random.Generator`: see the table in [here](https://numpy.org/doc/stable/reference/random/index.html#quick-start). Thus, for example, `random_sample()` and `rand()` have to be replaced with `random()`, and `randn()` with `standard_normal()` and so on.
These can be done as commits added to this PR or as independent PRs.

TODOs:
* [x] Replace `np.random.RandomState` specific methods to `np.random.Generator` compatible methods
  - [x] `random_sample()` -> `random()`
  - [x] `rand()` -> `random()`, but `rand(d0, d1, ..., dn)` -> `random((d0, d1, ..., dn))` (or `random(size=(d0, d1, ..., dn))`)
  - [x] `randn()` -> `standard_normal()`, but `randn(d0, d1, ..., dn)` -> `standard_normal((d0, d1, ..., dn))` (or `standard_normal(size=(d0, d1, ..., dn))`)
  - [x] `randint()` -> `util.rng_integers()`
* [x] Update docstrings
* [x] Add tests

### DIFF
--- a/quantecon/util/random.py
+++ b/quantecon/util/random.py
@@ -5,28 +5,36 @@ Utilities to Support Random State Infrastructure
 import numpy as np
 import numbers
 
-#-Random States-#
+
+# Random States
 
 def check_random_state(seed):
     """
-    Check the random state of a given seed.
+    Turn `seed` into a `np.random.RandomState` instance.
 
-    If seed is None, return the RandomState singleton used by np.random.
-    If seed is an int, return a new RandomState instance seeded with seed.
-    If seed is already a RandomState instance, return it.
+    Parameters
+    ----------
+    seed : None, int, `np.random.RandomState`, or `np.random.Generator`
+        If None, the `np.random.RandomState` singleton is returned. If
+        `seed` is an int, a new ``RandomState`` instance seeded with
+        `seed` is returned. If `seed` is already a `RandomState` or
+        `Generator` instance, then that instance is returned.
 
-    Otherwise raise ValueError.
+    Returns
+    -------
+    `np.random.RandomState` or `np.random.Generator`
+        Random number generator.
 
     Notes
     -----
-    This code was sourced from scikit-learn.
+    This code was originally sourced from scikit-learn.
 
     """
     if seed is None or seed is np.random:
         return np.random.mtrand._rand
     if isinstance(seed, (numbers.Integral, np.integer)):
         return np.random.RandomState(seed)
-    if isinstance(seed, np.random.RandomState):
+    if isinstance(seed, (np.random.RandomState, np.random.Generator)):
         return seed
     raise ValueError('%r cannot be used to seed a numpy.random.RandomState'
                      ' instance' % seed)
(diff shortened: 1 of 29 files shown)


oyamad:
There is incompatibility between [`RandomState.randint`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.randint.html) and [`Generator.integers`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.integers.html) (i.e., there is no method with a common name).

Use [`util.rng_integers`](https://github.com/QuantEcon/QuantEcon.py/blob/730582d79934652f26456862b162b5e40f300927/quantecon/util/random.py#L9), which is just an alias of [`scipy._lib._util.rng_integers`](https://github.com/scipy/scipy/blob/b4a970bf4eba1065ea31f8d0988cd4ec523a07d4/scipy/_lib/_util.py#L449-L510).

oyamad:
As demonstrated by #655, some calls can be replaced independent of this PR, without any breaking of the code (except for `randint`; see above).

quantecon/util/random.py
```diff
@@ -41,3 +41,48 @@ def check_random_state(seed):
...
+        If True, sample from the interval [low, high] instead of the default
+        [low, high) Defaults to False.
+    Returns
+    -------
+    out: int or ndarray of ints
+        size-shaped array of random integers from the appropriate distribution,
+        or a single such random int if size not provided.
+    """
+    return scipy_rng_integers(gen, low, high=high, size=size,
+            dtype=dtype, endpoint=endpoint)
```
oyamad, line 88: I would say this is redundant. Isn't just importing `scipy._lib._util.rng_integers` directly enough?

oyamad:
I propose to update the docstrings as follows (to keep them short):

```py
def some_function(main_arguments, random_state=None):
    """
    Parameters
    ----------
    ...

    random_state : int or np.random.RandomState/Generator, optional
        Random seed (integer) or np.random.RandomState or Generator
        instance to set the initial state of the random number generator
        for reproducibility. If None, a randomly initialized RandomState
        is used.

    """
    ...
```

```py
class SomeClass:
    def some_method(main_arguments, random_state=None):
        """
        Parameters
        ----------
        ...

        random_state : int or np.random.RandomState/Generator, optional
            Random seed (integer) or np.random.RandomState or Generator
            instance to set the initial state of the random number
            generator for reproducibility. If None, a randomly
            initialized RandomState is used.

        """
        ...
```

oyamad:
Let me merge this. Thanks for your help @Smit-create.

PR #656, opened 2022-11-24
[Refactor]: Make random methods compatible to RandomGenerator

Smit-create:


oyamad:
@Smit-create Will you keep `randint` as is in this PR (restrict only to `rand`/`random_sample` and `randn`)?

Changes of `randint` should be made as commits added to the branch [`np-random-generator`](https://github.com/QuantEcon/QuantEcon.py/tree/np-random-generator). (Do you have a write access?)

oyamad:
Sorry, `rand(d0, d1, ..., dn)` has to be replaced with `random((d0, d1, ..., dn))`
(or `random(size=(d0, d1, ..., dn))`).

Smit-create:
> (Do you have a write access?)

Yes, I have access. I will push `randint` changes in that PR.

oyamad (approved):
Looks good to me!

Issue #663, opened 2022-12-06
Unify QuantEcon.py and QuantEcon.jl APIs?


Smit-create:
One of the APIs that differ in `julia` and `python` versions is: `tauchen`. 
Julia version is: [Doc1](http://quantecon.github.io/QuantEcon.jl/latest/api/QuantEcon.html#QuantEcon.tauchen-Union{Tuple{T3},%20Tuple{T2},%20Tuple{T1},%20Tuple{Integer,%20T1,%20T2},%20Tuple{Integer,%20T1,%20T2,%20Any},%20Tuple{Integer,%20T1,%20T2,%20Any,%20T3}}%20where%20{T1%3C:Real,%20T2%3C:Real,%20T3%3C:Real})
Python version is: [Doc2](https://github.com/QuantEcon/QuantEcon.py/blob/6ffbe570008979e22b3156dd77745990b53ac74d/quantecon/markov/approximation.py#L138)

We can try to unify such APIs that differ in `julia` and `python`(like name of function arguments, position of parameters, etc.) which will help the users in migrating from one language to the other easily.

jstac:
Yes, that would be great @Smit-create .  The `tauchen` one has always bothered me.  This one might need unifying as well:

http://quantecon.github.io/QuantEcon.jl/latest/api/QuantEcon.html#QuantEcon.rouwenhorst

The only issue is that these are breaking changes, so we would need to coordinate changes on the lectures and any other downstream code (e.g., the DP textbook).

oyamad:
See also #363.

Smit-create:
One another difference:
Python: `markov_chain_object.P`
Julia: `markov_chain_object.p`

marcdelabarrera:
Can you make the UserWarning to appear only once per session?

PR #664, opened 2022-12-06
Unify `tauchen` and `rouwenhorst` API

### LINKED ISSUE #363: Unify and improve `tauchen` and `rouwenhorst`
The arguments to these two functions have different order, even though they do something very similar.

I think it would be worth breaking backwards compatibility to fix this.  We just need to take care to update the lectures accordingly.  

Also, `ybar` is missing from `tauchen` and it would be good to have it.

See http://quanteconpy.readthedocs.io/en/latest/markov/approximation.html

I propose that we change the signature of `tauchen` to `tauchen(n, ybar, sigma, rho, m=3)`.

Smit-create:
Fixes #363 
#663

### DIFF
--- a/quantecon/markov/approximation.py
+++ b/quantecon/markov/approximation.py
@@ -135,30 +135,31 @@ def rouwenhorst(n, ybar, sigma, rho):
     return MarkovChain(theta, bar)
 
 
-def tauchen(rho, sigma_u, b=0., m=3, n=7):
+def tauchen(n, rho, sigma, mu=0., n_std=3):
     r"""
     Computes a Markov chain associated with a discretized version of
     the linear Gaussian AR(1) process
 
     .. math::
 
-        y_{t+1} = b + \rho y_t + u_{t+1}
+        y_t = \mu + \rho y_{t-1} + \epsilon_t
 
-    using Tauchen's method. Here :math:`{u_t}` is an i.i.d. Gaussian process
+    using Tauchen's method. Here :math:`{\epsilon_t}` is an i.i.d. Gaussian process
     with zero mean.
 
     Parameters
     ----------
-    b : scalar(float)
-        The constant term of {y_t}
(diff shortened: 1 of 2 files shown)


oyamad:
Maybe add a note after this?

https://github.com/QuantEcon/QuantEcon.py/blob/e1c08e0e740c6e370ca9100901802e80d5eed93b/quantecon/markov/approximation.py#L25-L55

Something like "where `p = q = (1 + rho) / 2`."

oyamad (approved):
It looks good to me!

jstac:
Nice work @Smit-create !  

This looks good to me too but please coordinate with @mmcky to make sure the lectures are suitably modified.

Smit-create:
Thanks for the review. Once this is merged, I will quickly start fixing the lecture codes.

mmcky:
> Thanks for the review. Once this is merged, I will quickly start fixing the lecture codes.

@Smit-create the best way to go here is to install this branch version on a PR marked `do not merge` and you can use that for testing. You just need to update the `environment.yml` file to use the branch version using `git+https`. I will try and set that up for you later this evening if not sure of the syntax.

quantecon/markov/approximation.py
```diff
@@ -12,11 +12,11 @@
 from numba import njit
 
 
-def rouwenhorst(n, ybar, sigma, rho):
+def rouwenhorst(n, rho, sigma, mu=0.):
```
mmcky, line 15: @Smit-create while it is difficult in this change to preserve backward compatibility, can you add a `deprecation` notice to the doctoring at the bottom.

Smit-create: Done, thanks!

mmcky:
@Smit-create I will merge this into `master` but we will need lecture PR's tested prior to releasing `v0.6.0`

Issue #671, opened 2022-12-18
Migrate Commit History?


mmcky:
When cleaning up the `API` the commit history is not longer super clear from #601 

For example `arma.py` was migrated to `_arma.py` (with `arma.py` retained to preserve backward compatibility). 

Can the commit history of `arma.py` be concatenated to `_arma.py`?

Smit-create:
I tried to save the commit history to the new file using `git mv`. I see that the reason it got lost is due to merge conflicts and when I resolved those conflicts manually, it somehow got messed up. I will try to look into this.

Smit-create:
One option is that we can revert that PR and then I'll come up with a new PR fixing that in one go such that we don't have any merge conflicts.

Smit-create:
I checked by merging #674 into a copy of `main` branch in my fork and found that the commit history is not overridden, i.e., it uses the same old commits: https://github.com/Smit-create/QuantEcon.py/tree/check_671_merge

Another solution that I found is this:
1. Revert PR #601 and merge it in the `main`.
2. Merge #674 into the `main` and you can see the results as expected. This is how it would look: https://github.com/Smit-create/QuantEcon.py/commits/check_671_5/quantecon/_arma.py

oyamad:
Resolved by #682

PR #601, opened 2022-02-11
MAINT: Distinguish between private and public namespaces

Smit-create:
Related to #350 

I am trying to work on #350 by using private and public namespace changes. Since `ecdf.py` is not meant for public namespace i.e we want users to use `from quantecon import ECDF` instead of `from quantecon.ecdf import ECDF`, we can do the following to resolve this:
1. Rename `ecdf.py` to `_ecdf.py` using `git mv` (This preserves the commit history of `ecdf.py` and transfers it to `_ecdf.py`)
2. Create a new file `ecdf.py` and add warnings (This is required at the current stage to avoid breaking the code and can be removed safely in future versions)

The current PR works in the following way:
```python
>>> from quantecon.ecdf import ECDF  # works but throws a warning
<stdin>:1: DeprecationWarning: Please use `ECDF` from the `quantecon` namespace, the `quantecon.ecdf` namespace is deprecated.
>>> from quantecon import ECDF # works fine
```
We can remove the `ecdf.py` file in future versions and then we will just have `_ecdf.py` as a private namespace

cc @mmcky @jstac

PR #674, opened 2022-12-20
Fix Commit history of API changes

Smit-create:
Fixes #671

PR #675, opened 2022-12-22
Revert "MAINT: Distinguish between private and public namespaces (#601)"

Smit-create:
https://github.com/QuantEcon/QuantEcon.py/issues/671#issuecomment-1362408285

cc @mmcky @oyamad


Smit-create:
This will preserve the history back to the original one. Once this is merge, we can go ahead with #674

oyamad (approved):

PR #680, opened 2022-12-31
Fix APIs and preserve commit history

### LINKED ISSUE #671: Migrate Commit History?

Smit-create:
Fixes #671 
#601


Smit-create:
Closing in favor of #682

Issue #678, opened 2022-12-30
Add function fit_discrete_mc


jstac:
As discussed in https://github.com/QuantEcon/QuantEcon.py/pull/640, we should add a function called `fit_discrete_mc` that takes  a time series $X_0, \ldots, X_{T-1}$ taking values in $\mathbb R^n$ (each $X_t$ is a vector in continuous state space $\mathbb R^n$) and fits a finite Markov chain to it using the following steps:

1. Set up a one-dimensional grid in each dimension of $\mathbb R^n$ (e.g., if $n=2$ then set up a grid on the $x$ axis and the $y$ axis)
2. Take the cartesian product of these 1D grids to get a set of grid points in $\mathbb R^n$
3. Produce a new time series $\hat X_0, \ldots, \hat X_{T-1}$ by mapping each $X_t$ to the nearest grid point in $\mathbb R^n$.
4. Fit a finite Markov chain to the new time series and return as a `MarkovChain` object.

This function just combines existing tools, since

#. for step 3, we can use `cartesian_nearest_index`
#. for step 4, we can use `estimate_mc` (see https://github.com/QuantEcon/QuantEcon.py/pull/658)

@oyamad Could you please confirm that you agree?
@Smit-create @HumphreyYang might you be willing to collaborate on this issue?

oyamad:
One thing I am concerned with is that it is less general than what the general name `fit_discrete_mc` would suggest, in that the state space of the finite MC will be just a Cartesian product. I would like to have an implementation that would warrant future extendability. Suppose that the function will be called with `fit_discrete_mc(X, S)`.

* For now, as `S` it accepts an array_like (typically tuple) of array_like's of 1D grid points.
* In a future extension, add a `cartesian` option (with default `True`), where with `cartesian=False`, `S` will be a 2D array_like of n-dimensional points (which will be the state space of `mc`).

If this works, we can proceed for now as you propose.

jstac:
Thanks for your comments @oyamad .

The extension you suggest sounds like a good idea.  My understanding is that someone can proceed now in line with my suggestion and your bullet point one. We can then open as issue to add the enhancement that you suggested.

Smit-create:
Thanks @jstac and @oyamad ,
So are we trying to implement something like this?
```py
def fit_discrete_mc(X, S):
    X_new = cartesian_nearest_index(X, S)
    return estimate_mc(X_new)
```

jstac:
[...]

PR #681, opened 2023-01-02
Add fit_discrete_mc

Smit-create:
#678

### DIFF
--- a/quantecon/markov/estimate.py
+++ b/quantecon/markov/estimate.py
@@ -1,7 +1,7 @@
 import numpy as np
 from numba import njit
 from .core import MarkovChain
-from ..gridtools import cartesian_nearest_index, cartesian
+from .._gridtools import cartesian_nearest_index, cartesian
 
 
 def estimate_mc(X):
@@ -56,7 +56,7 @@ def _count_transition_frequencies(index_series, trans_counter):
 
 
 def fit_discrete_mc(X, grids, order='C'):
-    """
+    r"""
     Function that takes an arbitrary time series :math: `(X_t)_{t=0}^{T-1}` in
     :math: `\mathbb R^n` plus a set of grid points in each dimension and converts
     it to a MarkovChain by first applying discretization onto the grid
--- a/quantecon/markov/estimate.py
+++ b/quantecon/markov/estimate.py
@@ -55,7 +55,7 @@ def _count_transition_frequencies(index_series, trans_counter):
     return trans_counter
 
 
-def fit_discrete_mc(X, grid, order='C'):
+def fit_discrete_mc(X, grids, order='C'):
     """
     Function that takes an arbitrary time series :math: `(X_t)_{t=0}^{T-1}` in
     :math: `\mathbb R^n` plus a set of grid points in each dimension and converts
@@ -66,10 +66,10 @@ def fit_discrete_mc(X, grid, order='C'):
     ----------
 
     X: array_like(ndim=2)
-        Time-series such that the t-th column being :math:`x_t`.
-        It should be of the shape n x t.
+        Time-series such that the t-th row being :math:`x_t`.
+        It should be of the shape T x n, where n is the number of dimensions.
 
-    grid: array_like(array_like(ndim=1))
+    grids: array_like(array_like(ndim=1))
         Array of sorted arrays. Set of grid points in each dimension
 
     Returns
@@ -79,9 +79,9 @@ def fit_discrete_mc(X, grid, order='C'):
         An instance of the MarkovChain class constructed after discretization
         onto the grid.
     """
-    X_indices = cartesian_nearest_index(X.T, grid, order=order)
+    X_indices = cartesian_nearest_index(X, grids, order=order)
     mc = estimate_mc(X_indices)
     # Assign the visited states in the cartesian product as the state values
-    prod = cartesian(grid, order=order)
+    prod = cartesian(grids, order=order)
     mc.state_values = prod[mc.state_values]
     return mc
--- a/quantecon/markov/estimate.py
+++ b/quantecon/markov/estimate.py
@@ -1,6 +1,7 @@
 import numpy as np
 from numba import njit
 from .core import MarkovChain
+from ..gridtools import cartesian_nearest_index, cartesian
 
 
 def estimate_mc(X):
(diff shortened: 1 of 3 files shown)


oyamad (commented):
Thanks @Smit-create

* [ ] This part is missing: https://github.com/QuantEcon/QuantEcon.py/blob/745c0d447a95277aadbbc66c50a6dc9656bafe36/quantecon/markov/approximation.py#L371-L373
* [ ] Add tests
* [ ] `Examples` section will be helpful

There are a few points open for discussion:
* What shape should `X` be of?
* Is the name `S` of the second argument appropriate?
* Is this for approximation or estimation (which of `approximation.py` or `estimate.py` to include this code in)?

jstac:
Thanks for the review @oyamad .

I think this should be in estimate.py because there is no specified model to approximate.  I suggest a better name for S is grids.

I am away from my computer so I won't give further feedback.  @Smit-create , it would be great if you could coordinate with @oyamad to finish this PR.

Smit-create:
> There are a few points open for discussion:
> 
> * What shape should `X` be of?
> * Is the name `S` of the second argument appropriate?
> * Is this for approximation or estimation (which of `approximation.py` or `estimate.py` to include this code in)?

Thanks, @oyamad. I have addressed these questions and added the missing part. If the current diff looks good, I can start working on Examples and tests.

oyamad:
* Regarding the shape of `X`, I would prefer to follow the convention in `estimate_mc(X)` and `cartesian_nearest_index(X, ...)` where `X` is supposed to be a `T x n` array with `X[t]` containing the `t`-th observation. What do you think?
* For the second argument, I would follow the suggestion from @jstac (plural `grids`).

Smit-create:
> * where `X` is supposed to be a `T x n` array

I kept it as `n x T` because the output of `simulate_linear_model` is `n x T`, moreover, one can easily think of this as horizontally stacked individual vectors of shape `n x 1`.

> I would follow the suggestion from

Sure will change that.

jstac:
> I kept it as `n x T` because the output of `simulate_linear_model` is `n x T`

Both @oyamad and @Smit-create 's opinions are sensible on this point.

Another option --- combining these opinions but with a breaking change --- is to convert `linear_state_space` so that its output is $T \times n$.

This is more work but I personally prefer it, so that we have consistency across all of these functions.

oyamad (commented):
When `n=1`, it is supposed to pass for example `X = [[0.1], [0.6], [1.2], [0.2]]` and `grids = ([0., 0.5, 1.],)`. Is there any clever way to detect this case and allow passing `X = [0.1, 0.6, 1.2, 0.2]` and `grids = [0., 0.5, 1.]`?

Smit-create:
> Is there any clever way to detect this case

Thanks for the review. I am not sure if we should support this as we are clearly expecting the input shape of `T X n` which is independent of `n`. IMHO, it is good to keep it unchanged regarding this.

quantecon/markov/estimate.py
```diff
@@ -52,3 +53,50 @@ def _count_transition_frequencies(index_series, trans_counter):
...
+    Examples
+    --------
+
+    >>> grids = (np.arange(3), np.arange(2))
+    >>> X = [(-0.1, 1.2), (2, 0), (0.6, 0.4), (1.0, 0.1)]
+    >>> mc = fit_discrete_mc(X, grids)
+    >>> mc.state_values
+    array([[0, 1],
+        [1, 0],
+        [2, 0]])
```
oyamad, line 84: Mind the indentation:
```suggestion
    array([[0, 1],
           [1, 0],
           [2, 0]])
```

quantecon/markov/estimate.py
```diff
@@ -52,3 +53,50 @@ def _count_transition_frequencies(index_series, trans_counter):
...
+    >>> mc.P
+    array([[0., 0., 1.],
+        [0., 1., 0.],
+        [0., 1., 0.]])
```
oyamad, line 88:
```suggestion
    array([[0., 0., 1.],
           [0., 1., 0.],
           [0., 1., 0.]])
```

quantecon/markov/estimate.py
```diff
@@ -52,3 +53,50 @@ def _count_transition_frequencies(index_series, trans_counter):
...
+    Function that takes an arbitrary time series :math: `(X_t)_{t=0}^{T-1}` in
+    :math: `\mathbb R^n` plus a set of grid points in each dimension and converts
+    it to a MarkovChain by first applying discretization onto the grid
+    and then estimation of the Markov chain.
+
+    Parameters
+    ----------
+
+    X: array_like(ndim=2)
+        Time-series such that the t-th row being :math:`x_t`.
```
oyamad, line 69:
```suggestion
        Time-series such that the t-th row is :math:`x_t`.
```
or
```suggestion
        Time-series with the t-th row being :math:`x_t`.
```

quantecon/markov/estimate.py
```diff
@@ -52,3 +53,50 @@ def _count_transition_frequencies(index_series, trans_counter):
...
+        It should be of the shape T x n, where n is the number of dimensions.
+
+    grids: array_like(array_like(ndim=1))
+        Array of sorted arrays. Set of grid points in each dimension
```
oyamad, line 73:
```suggestion
        Array of `n` sorted arrays. Set of grid points in each dimension
```

quantecon/markov/tests/test_estimate.py
```diff
@@ -51,3 +51,34 @@ def test_mult_dim_state(self):
...
+                  (4.4, 4.0), (0.6, 0.4), (1.0, 0.1)]
+
+    def test_order_f(self):
+        mc = fit_discrete_mc(self.X, self.grids, order='F')
+        S_expected = np.array([[1, 0], [2, 0], [0, 1], [2, 3], [3, 4]])
+        P_expected = np.array([[1., 0., 0., 0., 0.],
+                            [0., 0., 0., 1., 0.],
+                            [0., 1., 0., 0., 0.],
+                            [0., 0., 0., 0., 1.],
+                            [1., 0., 0., 0., 0.]])
```
oyamad, line 69: Indentation

quantecon/markov/tests/test_estimate.py
```diff
@@ -51,3 +51,34 @@ def test_mult_dim_state(self):
...
+        assert_allclose(mc.P, P_expected)
+
+    def test_order_c(self):
+        mc = fit_discrete_mc(self.X, self.grids, order='C')
+        S_expected = np.array([[0, 1], [1, 0], [2, 0], [2, 3], [3, 4]])
+        P_expected = np.array([[0., 0., 1., 0., 0.],
+                            [0., 1., 0., 0., 0.],
+                            [0., 0., 0., 1., 0.],
+                            [0., 0., 0., 0., 1.],
+                            [0., 1., 0., 0., 0.]])
```
oyamad, line 81: Indentation

quantecon/markov/estimate.py
Smit-create: Thanks, resolved.

oyamad (approved):
@Smit-create Thanks!

jstac:
Many thanks @Smit-create and @oyamad

PR #687, opened 2023-01-06
Migrate `np.dot` to `@` Operator

### LINKED ISSUE #589: Migrate `np.dot` to use python @
It would be nice to migrate `np.dot` to use the inbuilt `python` @ operator for matrix multiplication.

Some thought is required as `np.dot` doesn't behave in the exact same way as `@`

https://stackoverflow.com/questions/34142485/difference-between-numpy-dot-and-python-3-5-matrix-multiplication

HumphreyYang:
This PR resolves #589.

### DIFF
--- a/quantecon/_lqnash.py
+++ b/quantecon/_lqnash.py
@@ -125,7 +125,7 @@ def nnash(A, B1, B2, R1, R2, Q1, Q2, S1, S2, W1, W2, M1, M2,
 
         # break up the computation of F1, F2
         F1_left = v1 - (H1 @ B2 + G1.dot(M1.T)) @ (H2 @ B1 + G2.dot(M2.T))
-        F1_right = H1 @ A +G1.dot(W1.T) - (H1 @ B2 + G1.dot(M1.T)) @ \
+        F1_right = H1 @ A + G1.dot(W1.T) - (H1 @ B2 + G1.dot(M1.T)) @ \
             (H2 @ A + G2.dot(W2.T))
         F1 = solve(F1_left, F1_right)
         F2 = H2 @ A + G2.dot(W2.T) - (H2 @ B1 + G2.dot(M2.T)) @ F1
--- a/quantecon/_lqnash.py
+++ b/quantecon/_lqnash.py
@@ -118,28 +118,27 @@ def nnash(A, B1, B2, R1, R2, Q1, Q2, S1, S2, W1, W2, M1, M2,
         F10 = F1
         F20 = F2
 
-        G2 = solve(np.dot(B2.T, P2.dot(B2))+Q2, v2)
-        G1 = solve(np.dot(B1.T, P1.dot(B1))+Q1, v1)
-        H2 = np.dot(G2, B2.T.dot(P2))
-        H1 = np.dot(G1, B1.T.dot(P1))
+        G2 = solve(B2.T @ P2 @ B2 + Q2, v2)
+        G1 = solve(B1.T @ P1 @ B1 + Q1, v1)
+        H2 = G2 @ B2.T @ P2
+        H1 = G1 @ B1.T @ P1
 
         # break up the computation of F1, F2
-        F1_left = v1 - np.dot(H1.dot(B2)+G1.dot(M1.T),
-                           H2.dot(B1)+G2.dot(M2.T))
-        F1_right = H1.dot(A)+G1.dot(W1.T) - np.dot(H1.dot(B2)+G1.dot(M1.T),
-                                                H2.dot(A)+G2.dot(W2.T))
+        F1_left = v1 - (H1 @ B2 + G1.dot(M1.T)) @ (H2 @ B1 + G2.dot(M2.T))
+        F1_right = H1 @ A +G1.dot(W1.T) - (H1 @ B2 + G1.dot(M1.T)) @ \
+            (H2 @ A + G2.dot(W2.T))
         F1 = solve(F1_left, F1_right)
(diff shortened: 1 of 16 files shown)


jstac:
Thanks @HumphreyYang , much appreciated.  Perhaps @mmcky will review and merge when he finishes his break.

quantecon/_lqnash.py
```diff
@@ -118,28 +118,27 @@ def nnash(A, B1, B2, R1, R2, Q1, Q2, S1, S2, W1, W2, M1, M2,
...
+        H1 = G1 @ B1.T @ P1
 
         # break up the computation of F1, F2
-        F1_left = v1 - np.dot(H1.dot(B2)+G1.dot(M1.T),
-                           H2.dot(B1)+G2.dot(M2.T))
-        F1_right = H1.dot(A)+G1.dot(W1.T) - np.dot(H1.dot(B2)+G1.dot(M1.T),
-                                                H2.dot(A)+G2.dot(W2.T))
+        F1_left = v1 - (H1 @ B2 + G1.dot(M1.T)) @ (H2 @ B1 + G2.dot(M2.T))
+        F1_right = H1 @ A + G1.dot(W1.T) - (H1 @ B2 + G1.dot(M1.T)) @ \
+            (H2 @ A + G2.dot(W2.T))
```
Smit-create, line 129: Here are some `.dot`. Are these kept intentionally?

HumphreyYang: > I left only a few np.dot in cases involving scalars. I noticed during the process that [the documentation](https://quanteconpy.readthedocs.io/en/latest/tools/lqnash.html) for parameter A of nnash function in _lqnash.py seems to indicate that A can be scalar. However, based on the setup of the formula, it seems to me that A is unlikely to be a scalar, given the two-player setup here. Please kindly correct me if I am wrong.

Many thanks @Smit-create. Yes, as noted before, these are cases where scalars are allowed, and `@` is not capable of multiplying matrices with scalars.

HumphreyYang:
Hi @mmcky, 

Could you please kindly review this PR when you have time? Sorry for mentioning you directly since I haven't been granted access that allows me to request reviews in this repo.

HumphreyYang:
>I left only a few np.dot in cases involving scalars. I noticed during the process that [the documentation](https://quanteconpy.readthedocs.io/en/latest/tools/lqnash.html) for parameter A of `nnash` function in _lqnash.py seems to indicate that A can be scalar.

Here is a short update on this PR. I updated the scalar terms to at least 2d to match the matrices in `_lqcontrol.py` so that `np.dot` can be transformed into `@`. I believe the same can be done for `_lqnash.py`, but before that, confirmations on the input restrictions of `nnash` function will be very useful as a guide.

mmcky:
thanks @HumphreyYang -- sorry this has taken me a long time. 

I think this change looks good. 

My only thoughts are that it is unfortunate to have to mix `.dot` and `@` due to the scalar issues. 

Just checking when you say `scalar` do you mean multiply every element in a matrix by a scalar? 

Would `3 * A` work for the scalar multiplication case?

HumphreyYang:
Hi @mmcky,

Many thanks for your suggestion!

> Just checking when you say `scalar` do you mean multiply every element in a matrix by a scalar? 

The tricky case I am referring to is where a variable can be both `scalar` or `array`. In that case, `*` will not capture the `array` case. For example, in `lqnash`:

```
    W1 : scalar(float) or array_like(float)
        As above, size (n, k_1)
    W2 : scalar(float) or array_like(float)
        As above, size (n, k_2)
    M1 : scalar(float) or array_like(float)
        As above, size (k_2, k_1)
    M2 : scalar(float) or array_like(float)
        As above, size (k_1, k_2)
```

But I found a way to enforce at least 2d and overcome this issue (i.e. if it is a scalar `1`, then we put it into `[[1]]`).

I will implement this first and ping you to see if it looks good to you.

Many thanks in advance.

quantecon/game_theory/normal_form_game.py
```diff
@@ -258,7 +258,7 @@ def reduce_last_player(payoff_array, action):
             if isinstance(action, numbers.Integral):  # pure action
                 return payoff_array.take(action, axis=-1)
             else:  # mixed action
-                return payoff_array.dot(action)
+                return payoff_array @ action
```
oyamad, line 261: I am not sure whether there would be no side effect from this change, when the dimension of `payoff_array` > 2.

oyamad (commented):
I would say, unless you are perfectly sure, `@` should be restricted to the cases where `A` in `A @ b` or `A @ B` has dimension 2.

I added a comment on one part of the change, but it is just one example.

quantecon/game_theory/normal_form_game.py
HumphreyYang: Many thanks @oyamad,

I think this is a really nice point as the equivalence is only established in 2-D arrays. Although all the tests are passed at the moment, it would be great to be cautious. In cases where `N-D.dot(1-D)`, the behaviour might also be unexpected.

I can write some tests on this while trying edge cases if you think it is a good idea.

Many thanks in advance.

mmcky:
thanks @HumphreyYang appreciate your work on this. It looks like #787 covers more cases -- do you think we should close this and invest in #787?

HumphreyYang:
> thanks @HumphreyYang appreciate your work on this. It looks like #787 covers more cases -- do you think we should close this and invest in #787?

Sounds great! This is historic and should be sent to a museum.

PR #692, opened 2023-01-12
Update __Iss.py -  Remove broken link

bensonarafat:
Removed broken link

### DIFF
--- a/quantecon/_lss.py
+++ b/quantecon/_lss.py
@@ -1,11 +1,6 @@
 """
 Computes quantities associated with the Gaussian linear state space model.
 
-References
-----------
-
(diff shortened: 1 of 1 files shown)


oyamad (approved):
@bensonarafat Thanks!

@mmcky I think the link on this part can be removed: the comments on the head of a file no longer shows up in the documentation (see for example https://quanteconpy.readthedocs.io/en/latest/tools/lss.html). If we want to provide a link to a lecture, it should be in the docstring of a class/function (`LinearStateSpace` in this case).

mmcky:
thanks @bensonarafat for your PR. It is greatly appreciated. 

Thanks @oyamad

PR #703, opened 2023-05-18
CI: Add testing against Python 3.11

oyamad:
And drop 3.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,10 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.9, '3.10']
-        exclude:
-          - os: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -17,10 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.9, '3.10']
-        exclude:
-          - os: windows-latest
-            python-version: 3.8
+        python-version: [3.9, '3.10', 3.11]
```
mmcky, line 20: @oyamad does the `3.11` require quotes like `3.10`?

mmcky (approved):

.github/workflows/ci.yml
oyamad: I believe only `3.10` requires quotes.

PR #706, opened 2023-05-23
FIX: DOC: Remove `html4_writer = True`

oyamad:
Fix the build error "HTML 4 is no longer supported by Sphinx": https://readthedocs.org/projects/quanteconpy/builds/20742271/


mmcky:
thanks @oyamad

PR #724, opened 2024-03-04
ENH: add njit interpolation functions

mmcky:
This PR adds the following interpolation functions

- `interp` that accepts x as `float` or `np.array`

and the following internal functions:

- `_interpf` that accepts x as `float`
- `_interpa` that accepts x as `np.array`

### DIFF
--- a/quantecon/__init__.py
+++ b/quantecon/__init__.py
@@ -33,7 +33,7 @@ from ._gridtools import (
 )
 from ._inequality import lorenz_curve, gini_coefficient, shorrocks_index, \
     rank_size
-from ._interpolation import interp, interpf, interpa
+from ._interpolation import interp
 from ._kalman import Kalman
 from ._lae import LAE
 from ._arma import ARMA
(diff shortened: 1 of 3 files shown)


oyamad:
`np.interp` is available through `Numba`: see https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html.
Any advantage over that here?

mmcky:
> `np.interp` is available through `Numba`: see https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html. Any advantage over that here?

thanks @oyamad probably not -- i didn't realise it was supported. 

As we have decided to revert the order of arguments to match the numpy order we should use the numba built-in.

mmcky:
cc @jstac

jstac:
Thanks @mmcky heading to the hospital now.

mmcky:
> Thanks @mmcky heading to the hospital now.

Hope all goes well. Thanks @jstac

jstac:
Thanks @oyamad .

Sorry @mmcky , my bad.  perhaps we can test if swapping for the numpy version degrades performance or causes any issues in one or two lectures that are currently using jitted functions plus the version from `interpolation`.  If all is good we can close this PR and swap every 1d interp to the numpy version.

Issue #723, opened 2024-03-03
MAINT: Deprecations and Test Warnings


mmcky:
The following deprecation notices and warnings are issued during test runs that should be fixed. @Smit-create would you be able to look into these?

```
quantecon/markov/tests/test_core.py::test_raises_non_homogeneous_state_values
  /Users/mmcky/work/quantecon/QuantEcon.py/quantecon/markov/core.py:230: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    values = np.asarray(values)

quantecon/markov/tests/test_ddp.py::TestFiniteHorizon::test_backward_induction
quantecon/markov/tests/test_ddp.py::test_ddp_beta_1_not_implemented_error
  /Users/mmcky/work/quantecon/QuantEcon.py/quantecon/markov/ddp.py:430: UserWarning: infinite horizon solution methods are disabled with beta=1
    warnings.warn(msg)

quantecon/tests/test_graph_tools.py::test_raises_non_homogeneous_node_labels
  /Users/mmcky/work/quantecon/QuantEcon.py/quantecon/_graph_tools.py:140: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    values = np.asarray(values)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

oyamad:
* The 2nd one is as intended. Maybe define a custom warning rather than the generic `UserWarning`?
* 1st and 3rd should be addressed.

kp992:
Thanks, @oyamad @mmcky. I can look into this.

mmcky:
thanks @kp992

PR #726, opened 2024-03-07
Fix deprecation warnings coming from `np.asarray`

kp992:
Fix deprecation warnings coming from `np.asarray`.
Related to #723

### DIFF
--- a/quantecon/_graph_tools.py
+++ b/quantecon/_graph_tools.py
@@ -137,7 +137,7 @@ class DiGraph:
         if values is None:
             self._node_labels = None
         else:
-            values = np.asarray(values)
+            values = np.asanyarray(values)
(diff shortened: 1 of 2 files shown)


oyamad:
@kp992 Thanks, but this doesn't fix the issue: https://github.com/QuantEcon/QuantEcon.py/actions/runs/8183575000/job/22376624643?pr=726#step:8:76

As the message says, create an `ndarray` with `dtype=object` here https://github.com/QuantEcon/QuantEcon.py/blob/d7c2f5261301ff159c32dd79300e500a442856bc/quantecon/markov/tests/test_core.py#L587-L590 and here https://github.com/QuantEcon/QuantEcon.py/blob/d7c2f5261301ff159c32dd79300e500a442856bc/quantecon/tests/test_graph_tools.py#L291-L294

PR #733, opened 2024-08-13
CI: Add Python 3.12, drop 3.9

oyamad:


oyamad:
* For `python -m ensurepip --upgrade`, refer to https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean.
* But it did not work for windows-latest python-version 3.12, so excluded.
* The `ModuleNotFoundError: No module named 'urllib3.packages.six.moves'` error was fixed by requiring `urllib3>=2`.

oyamad:
Let me merge this.

PR #741, opened 2024-11-05
Multiplayer Polymatrix Game Solver (new for QuantEcon)

GautamsGitHub:
### Multiplayer Polymatrix Game Solver
The Linear Complementary Problem solving step requires some modifications from the one available in QuantEcon's LCP Lemke. I have added this solver. It is based on [Howson's paper ](https://www.jstor.org/stable/2634798). The only other implementation I have seen of it is in C in [GameTracer](http://dags.stanford.edu/Games/gametracer.html).

**Polymatrix Games**
I have added a class for dealing with polymatrix games. This was part of a larger project of mine with many utility functions on games and my own class for normal form games but I have trimmed most of this away from contribution to QuantEcon.

**gam files**
Gametracer's gam files are a good way of storing games in files because they can be generated by [GAMUT](http://gamut.stanford.edu/). I add a function to load a QuantEcon game from a gam file.


oyamad:
@GautamsGitHub Thank you so much for your contribution! Let me play around with it and then I will comment.

oyamad (commented):
One quick comment: Could you modify the docstrings in accordance with [this page](https://quantecon.org/wiki-py-docstrings/)?

For example, `Args:` -> `Parameters` with `----------`.

oyamad (commented):
For the errors in the tests, you may cherry-pick this 82e4eba; and also this 61d102b.

quantecon/game_theory/howson_lcp.py
```diff
@@ -0,0 +1,124 @@
...
+                retro = True
+                break
+            elif leaving_var < n:
+                pivcol = leaving_var + n
+            else:
+                pivcol = leaving_var - n
+
+    combined_solution = _get_solution(tableau, basis, z)
+
+    eq_strategies = [
```
oyamad, line 116: Use "actions" rather than "strategies"? Maybe change `eq_strategies` to `eq_actions` or `eq` or `NE`?

quantecon/game_theory/howson_lcp.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
+from .polymatrix_game import PolymatrixGame
+from ..optimize.pivoting import _pivoting, _lex_min_ratio_test
+from ..optimize.lcp_lemke import _get_solution
+
+
+def polym_lcp_solver(polym: PolymatrixGame):
+    """
+    Finds the Nash Equilbrium of a polymatrix game
```
oyamad, line 9:
```suggestion
    Finds a Nash Equilbrium of a polymatrix game
```

oyamad (changes requested):
The code looks very good!
Some requests:
* In `polym_lcp_solver`, it would be better to provide a `max_iter` option.
* Would it make sense to return a `NashResult` when a `full_output` option is set `True`, which is an "object containing information about the computation"?
* Please expand the docstring of `PolymatrixGame` by adding
  - the definition of polymatrix game;
  - an example showing how to construct an instance;
  - description of how a `NormalFormGame` input is "approximated".
* Please expand the docstrig of `polym_lcp_solver` adding an example showing how to use it.

GautamsGitHub:
I have been doing some more testing and I need to fix many things. I had a different setup for reading gam files etc. earlier. This version does not work properly yet.

oyamad:
For conversion to `NormalFormGame`, would Numpy broadcasting work?

```python
def get_player(polym, player_idx):
    N = polym.N
    opps = tuple(range(player_idx+1, N)) + tuple(range(player_idx))
    if N == 2:
        return Player(polym.polymatrix[(player_idx, opps[0])])
     
    newaxes = np.full((N-1, N-1), np.newaxis)
    np.fill_diagonal(newaxes, slice(None))
    payoff_array = sum([
        polym.polymatrix[(player_idx, opps[i])][:, *newaxes[i]]
        for i in range(N-1)
    ])
    return Player(payoff_array)


nfg = NormalFormGame([get_player(polym, i) for i in range(polym.N)])
```

oyamad:
@GautamsGitHub What problems do you have? Let me know if I can help.

GautamsGitHub:
> @GautamsGitHub What problems do you have? Let me know if I can help.

I wasn't too sure on the formatting of the documentation and type hints. The format should now be consistent with most of the rest of the project. I think the type hints make things clear. Next, I will add some doctest examples.

quantecon/game_theory/howson_lcp.py
GautamsGitHub: I have gone with NE

quantecon/game_theory/howson_lcp.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+
+            _pivoting(tableau, pivcol, pivrow)
+            basis[pivrow], leaving_var = pivcol, basis[pivrow]
+
+            if leaving_var == finishing_x or leaving_var == finishing_y:
+                p += 1
+                break
+            elif leaving_var == finishing_v:
+                print("entering the backtracking case")
```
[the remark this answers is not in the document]
GautamsGitHub: Yes. Perhaps I will add a debug flag, because this backtracking case is very rare so I like to know when it occurs.

quantecon/game_theory/howson_lcp.py
```diff
@@ -0,0 +1,124 @@
...
+                tableau, pivcol, 0, argmins,
+            )
+
+            _pivoting(tableau, pivcol, pivrow)
+            basis[pivrow], leaving_var = pivcol, basis[pivrow]
+
+            if leaving_var == finishing_x or leaving_var == finishing_y:
+                p += 1
+                break
+            elif leaving_var == finishing_v:
```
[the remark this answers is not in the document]
GautamsGitHub: It gets reached.

quantecon/game_theory/polymatrix_game.py
```diff
@@ -0,0 +1,138 @@
...
+                ]
+            ]
+        )
+        for action_combination in action_combinations
+    ])
+    hh_actions = hh_actions_and_payoffs[:, :-1]
+    combined_payoffs = hh_actions_and_payoffs[:, -1]
+
+    # different ways to solve the simultaneous equations
+    if is_polymatrix:
```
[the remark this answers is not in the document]
GautamsGitHub: When a Normal Form Game is not a Polymatrix Game, it is not clear what a good approximation would mean (we don't even know if we are interested in Nash Equilibria of the game yet). In this code, I used an approach of trying to minimise some metric of the differences in payoffs from the games; it is not very careful. Perhaps I should get rid of this provision.

GautamsGitHub:
> For conversion to `NormalFormGame`, would Numpy broadcasting work?
> 
> ```python
> def get_player(polym, player_idx):
>     N = polym.N
>     opps = tuple(range(player_idx+1, N)) + tuple(range(player_idx))
>     if N == 2:
>         return Player(polym.polymatrix[(player_idx, opps[0])])
>      
>     newaxes = np.full((N-1, N-1), np.newaxis)
>     np.fill_diagonal(newaxes, slice(None))
>     payoff_array = sum([
>         polym.polymatrix[(player_idx, opps[j])][:, *newaxes[j]]
>         for j in range(N-1)
>     ])
>     return Player(payoff_array)
> 
> 
> nfg = NormalFormGame([get_player(polym, i) for i in range(polym.N)])
> ```

I am not sure what extracting a player from a game is good for. There is a to_nf function in PolymatrixGame.

oyamad:
It is a proposal of an alternative for your `to_nf` function. (A `NormalFormGame` is basically a collection of `Player`s.)
My code with Numpy broadcasting, if correct, is shorter and runs faster, I suppose. (Of course, it is your choice, whether or not to adopt it.)

quantecon/game_theory/polymatrix_game.py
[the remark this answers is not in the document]
oyamad: It is your choice. Any approach, if it is "reasonable" in some sense, is fine as long as it is clearly documented.

GautamsGitHub:
> It is a proposal of an alternative for your `to_nf` function. (A `NormalFormGame` is basically a collection of `Player`s.) My code with Numpy broadcasting, if correct, is shorter and runs faster, I suppose. (Of course, it is your choice, whether or not to adopt it.)

I have now come to understand your numpy broadcasting for this method. It is very good. I am taking out the special handling of the 2 player case because it is unnecessary.

quantecon/game_theory/polymatrix_game.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Attributes
+    ----------
+    number_of_players : int
+        Number of players.
+
+    nums_actions : Sequence[int]
+        The number of actions available to each player.
+
+    polymatrix : Mapping[ tuple[int, int], Matrix ]
```
oyamad, line 126: Is `self.polymatrix` not restricted to a `dict` with `np.array` as values?

oyamad (commented):
A few comments...

quantecon/game_theory/polymatrix_game.py
```diff
@@ -0,0 +1,243 @@
...
+        -------
+        Player
+            Player object which has the player's payoff function.
+        """
+        self.N
+        opps = tuple(range(player_idx+1, self.N)) + tuple(range(player_idx))
+        newaxes = np.full((self.N-1, self.N-1), np.newaxis)
+        np.fill_diagonal(newaxes, slice(None))
+        payoff_array = sum([
+            self.polymatrix[(player_idx, opps[j])][:, *newaxes[j]]
```
oyamad, line 213: It turned out that this expression `[:, *newaxes[j]]` is not supported in Python 3.10:

> ```
> E     File "/home/runner/work/QuantEcon.py/QuantEcon.py/quantecon/game_theory/polymatrix_game.py", line 213
> E       self.polymatrix[(player_idx, opps[j])][:, *tuple(newaxes[j])]
> E                                                 ^
> E   SyntaxError: invalid syntax
> ```

(https://github.com/QuantEcon/QuantEcon.py/actions/runs/11947385419/job/33303244882#step:7:70)

Please apply the diff 26cf5a6

quantecon/game_theory/polymatrix_game.py
```diff
@@ -0,0 +1,243 @@
...
+of their payoffs from bimatrix games against each player.
+i.e. If two opponents deviate, the change in payoff
+is the sum of the changes in payoff of each deviation.
+
+"""
+
+import numpy as np
+from itertools import product
+from collections.abc import Sequence, Mapping
+from typing import Any, TypeAlias, Self
```
oyamad, line 14: `Self` does not exist in Python 3.10...

> ```
> E   ImportError: cannot import name 'Self' from 'typing' (/home/runner/miniconda3/envs/qe/lib/python3.10/typing.py)
> ```

(https://github.com/QuantEcon/QuantEcon.py/actions/runs/11947499924/job/33303544452?pr=742#step:7:27)

oyamad:
I have written "`random_polymatrix_game`". You might cherry pick 7c550bcbf36b4805b30d255344b50b13f9ce85a6 if you think it is useful...

GautamsGitHub:
I am taking a look at the doctests. Some doctests in other modules do not pass for me but can be fixed by putting things in print. For the relative file import I'm making in the game_converters doctest, it works when i run `pytest --doctest-modules` from the game_theory directory but I am not sure if this is the common way.

quantecon/game_theory/polymatrix_game.py
GautamsGitHub: Oh well. I have commented them out.

quantecon/game_theory/polymatrix_game.py
GautamsGitHub: I am not sure on best practice with type hints regarding the numpy parts. So I made a type alias `Matrix` and left it as `any` as a placeholder but it should be a 2D numpy array (I could not find a neat way to type hint this). When I return one from `hh_player_payoffs`, I should specify it is a dict; however if someone wants to construct a polymatrix game using some other mapping, I don't think I need to stop them.

oyamad:
@GautamsGitHub Thanks, we are almost there!

Regarding the `polymatrix` option in `__init__` and the `self. polymatrix` attribute:

(1)
Under the understanding that any `Mapping` object has the method `len` implemented (I referred to [this](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes)), `number_of_players` can be inferred:

```python
n = len(polymatrix)
self.N = (math.isqrt(4*n+1) + 1) // 2
```

(2)
The current implementation of `get_player()` fails at this line: https://github.com/QuantEcon/QuantEcon.py/blob/659443de3fc9867ecb6b9680366d133492a2231b/quantecon/game_theory/polymatrix_game.py#L243
It assumes that the values of `self.polymatrix` are `np.array`.
If we allow the values to not always be `np.array`, we would have to do this:

```python
            np.asarray(self.polymatrix[(player_idx, opps[j])])[tuple(newaxes[j])]
```

(3)
I believe that the type of the values of `self.polymatrix` are determined, as `np.array`:

```python
pairs = np.where(np.arange(N)[:, np.newaxis] != np.arange(N))
self.polymatrix = {
    (i, j): np.asarray(polymatrix[(i, j)]) for i, j in zip(*pairs)
}
```

(4)
Then, `nums_actions` can also be inferred:

```python
self.nums_actions = self.polymatrix[(0, 1)].shape + \
    tuple([self.polymatrix[(0, j)].shape[-1] for j in range(2, self.N)])
```

(5)
So I would propose the following:

```python
import math

...

    def __init__(
            self,
            polymatrix: Mapping[
                tuple[int, int],
                Matrix
            ]
    ) -> None:
        n = len(polymatrix)
        self.N = (math.isqrt(4*n+1) + 1) // 2

        pairs = np.where(np.arange(self.N)[:, np.newaxis] != np.arange(self.N))
        self.polymatrix = {
            (i, j): np.asarray(polymatrix[(i, j)]) for i, j in zip(*pairs)
        }

        self.nums_actions = self.polymatrix[(0, 1)].shape + tuple(
            [self.polymatrix[(0, j)].shape[-1] for j in range(2, self.N)]
        )
```

What do you think?

oyamad:
Please register in `game_theory/__init__.py` functions you want to export; see 61d102bc96ca27b39c73db14cba4b8679a8418ed

GautamsGitHub:
2. When I say `Matrix`, I mean 2D numpy array, I just could not find a neat way to type hint it. For example I use numpy stacking in Howson LCP. Should I make `Matrix: TypeAlias = NDArray`?

GautamsGitHub:
On inferring the numbers of players and actions:
I am considering also leaving in the option of specifying the numbers. It allows creating games with holes (which are useful to represent polymatrix-like games where players are in a graph and don't all face each other), and as long as they do not expect everything (like Howson LCP) to work, it is fine. On the other hand, it would be good to have the polymatrix store everything (and if someone wants to ignore a match-up, they have to put in 0s)

oyamad:
That makes sense, allowing "holes".
Maybe something like this?

```python
def __init__(self, polymatrix, nums_actions=None):
    if nums_actions is None:
        # infer nums_actions from polymatrix
    else:
        ...
        self.polymatrix = {}
        for i, j in zip(*pairs):
            try:
                self.polymatrix[(i, j)] = np.asarray(polymatrix[(i, j)])
            except KeyError:
                self.polymatrix[(i, j)] = \
                    np.zeros((nums_actions[i], nums_actions[j]))
```

oyamad:
2. I have no experience with type hints, and actually, we don't do type annotation in this library. At least for the docstrings, just follow the "convention" in this library? For example:

   ```
   Attributes
   ----------
   N : scalar(int)
       The number of players.

   nums_actions : tuple(int)
       Tuple of the numbers of actions, one for each player.

   polymatrix : dict[tuple(int), ndarray(float, ndim=2)]
       Dictionary containing the payoff matrices of pairwise interations,
       where `polymatrix[(i, j)]` is a 2D ndarray of player `i`'s payoffs
       from the interaction with player `j`.
   ```

GautamsGitHub:
Let me know if I have missed some things (also I will add some more features, like saving to gam file, and more tests).
Tests and doctests on the new parts are working for me.
In 2. does "failing" mean there is some test it is failing (that can be run now to check it is fixed) or just it was wrong?

GautamsGitHub:
So far I have avoided making changes to any of the existing material. But I think the function to compare Normal Form Games (which I currently have in `test_polymatrix_game.py`) is useful. Do you have a recommendation on this.
Functions called `isclose` are common and are similar to this function so I think I will just name it that.

`
    @classmethod
    def isclose(
            nf1: NormalFormGame,
            nf2: NormalFormGame,
            atol: float = 1e-4
    ) -> bool:
        if nf1.N != nf2.N:
            return False
        for player in range(nf1.N):
            if nf1.nums_actions[player] != nf2.nums_actions[player]:
                return False
        for player in range(nf1.N):
            if not np.allclose(
                nf1.players[player].payoff_array,
                nf2.players[player].payoff_array,
                atol=atol
            ):
                return False
        return True
`

oyamad:
@GautamsGitHub Thanks!

First of all, I propose to discuss "saving to gam file" and "`isclose`" (which will be useful!) in other independent PRs. Let us focus in this PR on `PolymatrixGame` and `polym_lcp_solver`.

By "failing" in 2. I meant it failed with the following input:

```python
# Dictionary with Python lists as values
polymatrix = {
    (0, 1): [[1, -1], [-1, 1]],
    (1, 0): [[-1, 1], [1, -1]]
}
polymg = PolymatrixGame(2, (2, 2), polymatrix)
polymg.to_nfg()
```

With your new changes, `PolymatrixGame(polymatrix)` fails at https://github.com/QuantEcon/QuantEcon.py/blob/91e2930ff3a69f6f1271d9ee299be9d1b613e2e6/quantecon/game_theory/polymatrix_game.py#L219-L220

### On the `polymatrix` input and the `polymatirx` attribute

My opinion is:
1. The `polymatrix` input should allow flexibility: for example, it should accept a dict with Python lists (not ndarrays) as values (for convenience).
2. The type of the `polymatirx` attribute should be fixed: say, to `dict[tuple(int), ndarray(float, ndim=2)]`. The docstring should clearly document it, so as for the user not to be "surprised".

If we follow this, the values of the input have to converted by e.g. `np.asarray`. What do you think?

GautamsGitHub:
For the documentation: the 2 dimensional numpy array currently sits behind a type alias of Matrix so the docstring just says `polymatrix : Mapping[ tuple[int, int], Matrix ]`
I suppose removing the alias would make it clearer.

For the argument to the constructor, asarray is a good idea.

quantecon/game_theory/polymatrix_game.py
```diff
@@ -0,0 +1,317 @@
...
+    Attributes
+    ----------
+    N : int
+        Number of players.
+
+    nums_actions : Sequence[int]
+        The number of actions available to each player.
+
+    polymatrix : Mapping[ tuple[int, int], Matrix ]
+        Maps each pair of player numbers to a matrix.
```
oyamad, line 159:
```suggestion
    Attributes
    ----------
    N : scalar(int)
        Number of players.

    nums_actions : tuple(int)
        The number of actions available to each player.

    polymatrix : dict[tuple(int), ndarray(float, ndim=2)]
        Maps each pair of player numbers to a matrix.
```

(And other places)

oyamad (changes requested):
(Hopefully) last request:
If you agree, regarding the type descriptions in the docstrings, would you follow the convention in the library, like this?
https://github.com/QuantEcon/QuantEcon.py/blob/b47b26d55c4f89301cab769b1703f5c6dfff7a4d/quantecon/game_theory/normal_form_game.py#L551-L563

quantecon/game_theory/polymatrix_game.py
GautamsGitHub: Refreshed and saw this after committing.
The convention I have gone for throughout (except for the `ndim=2` part) is like `tuple[int]` instead of `tuple(int)` because this is what VS code generates automatically for the numpy style docs. However, I think it is alright if sometimes integers are documented as `int` and other times as `scalar(int)`.

GautamsGitHub: I made the Sequence instead of concrete type on returns and attributes a few times though. Thanks.

GautamsGitHub:
I think the current balance between making PolymatrixGame indestructible and simplicity is alright. Trying to construct using different sized rows still reaches one of these
`
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (2, 2) + inhomogeneous part.
` 
but protecting against it would make the function very long and unclear.

oyamad (approved):
I think this can be merged. (The test failure is not because of the changes by this PR.)

@GautamsGitHub Great job!
If you would like, you can add other features in new PRs!

GautamsGitHub:
I will try to squash some of the commits where I remember flake8 moments after committing.

oyamad:
Either way is fine. (I can squash into one commit upon merge.) If you want to squash yourself, please rebase on the `main` branch.

GautamsGitHub:
So I think I just diverged when I squashed the commits, then went git pull to see what would happen etc ...
Anyway, I think I will leave it at this.

oyamad:
@GautamsGitHub OK, thanks again!
Merging.

PR #743, opened 2024-11-10
CI: Install libmamba from conda-forge

oyamad:
Resolve "`Error while loading conda entry point: conda-libmamba-solver (libarchive.so.19: cannot open shared object file: No such file or directory)`"

Refer to conda/conda-libmamba-solver#283


mmcky:
thanks @oyamad

PR #744, opened 2024-11-27
FIX: Remove test_solves_huge_7_player_polymatrix

oyamad:
Remove `test_solves_huge_7_player_polymatrix` which causes `ModuleNotFoundError: No module named 'numpy._core'` on `main` with Python 3.10:

```
_____________________ test_solves_huge_7_player_polymatrix _____________________

    def test_solves_huge_7_player_polymatrix():
        filename = "huge_7_player_polymatrix.pkl"
        with open(os.path.join(data_dir, filename), 'rb') as file:
>           polymg: PolymatrixGame = pickle.load(file)
E           ModuleNotFoundError: No module named 'numpy._core'

quantecon/game_theory/tests/test_howson_lcp.py:377: ModuleNotFoundError
```
(https://github.com/QuantEcon/QuantEcon.py/actions/runs/12042134410/job/33575263897#step:7:78)

(This doesn't happen on my environment with 3.12 though.)

@GautamsGitHub I do not know what is happening here. I remove this part to prevent test failure on the `main` branch. Will you open a new PR to see what causes this, if you want to keep this "pickle test"?


GautamsGitHub:
This is unfortunate. But I have a solution, I will save a stacking of the polymatrix using numpy itself instead of pickle. The stacking, which I had as a method of PolymatrixGame in a different version, can also be convenient because it can be applied to a concatenated mixed action and give a concatenated payoffs for responses.

 ```python
def to_one_big_matrix(self, costs=True, low_avoider=2.0):
        """
        Puts all of the matrices of the Polymatrix Game into
        One Big Matrix where the submatrices on the diagonal
        are zero and the submatrices at (i, j) are the
        component payoffs to player i in the head to head game
        between players i and j.

        Args:
            costs (bool, optional):
                Should the payoffs be turned into costs.
            low_avoider (float, optional): 
                A number to keep the entries positive.
                Defaults to 2.0.

        Returns:
            _type_: One Big Matrix.
        """
        positive_cost_maker = self.range_of_payoffs()[1] + low_avoider
        sign = -1 if costs else 1
        M = np.vstack([
            np.hstack([
                np.zeros((self.actions[player], self.actions[player])) if p2 == player
                else (positive_cost_maker + sign * self.polymatrix[(player, p2)])
                for p2 in range(self.players)
            ])
            for player in range(self.players)
        ])
        return M
```

PR #746, opened 2024-11-27
CI: Install numba=0.61.0rc1, Test with Python 3.13

oyamad:


mmcky:
thanks for managing these tests @oyamad.

oyamad:
Python 3.13 is not installed... (https://github.com/QuantEcon/QuantEcon.py/actions/runs/12044135624/job/33580662520)
Maybe not supported yet?

mmcky:
@oyamad it installs `python=3.13` but then `conda` uses `python=3.12.3`

Not sure why as it should be available https://conda-forge.org/blog/2024/09/26/python-313/

I suspect its coming from a package in `environment_np2.yml`. Perhaps we could try adding

```
python=3.13
```

to the environment to force the version (if compatible).

oyamad:
Using `conda-incubator/setup-miniconda` it now successfully works with Python 3.13!

PR #749, opened 2024-12-08
Add __repr__ for PolymatrixGame

oyamad:
Similar to that of `NormalFormGame`.

```
>>> g = gt.random_game((2, 3, 4))
>>> g
<2x3x4 3-player NormalFormGame of dtype float64>
>>> polymg = gt.random_polymatrix_game((2, 3, 4))
>>> polymg
<2x3x4 3-player PolymatrixGame>
```

@GautamsGitHub By the way, shall we enforce homogeneous dtype across players in `PolymatrixGame`, as in `NormalFormGame`?
https://github.com/QuantEcon/QuantEcon.py/blob/b733469cae922e7956b23d782054329d464129f7/quantecon/game_theory/normal_form_game.py#L566-L591

### DIFF
--- a/quantecon/game_theory/polymatrix_game.py
+++ b/quantecon/game_theory/polymatrix_game.py
@@ -39,7 +39,7 @@ from collections.abc import Sequence, Mapping, Iterable
 # from typing import TypeAlias, Self
 from numpy.typing import NDArray
 
-from .normal_form_game import NormalFormGame, Player
+from .normal_form_game import NormalFormGame, Player, _nums_actions2string
 
 
 def hh_payoff_player(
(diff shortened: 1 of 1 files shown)


GautamsGitHub:
This sounds good. It is helpful because we can combine player's payoffs into vectors without worrying about there being different dtypes. I also think raising an error is better than casting everything to some datatype.

oyamad:
Sorry, I realized that the dtype is set `float` whatever the dtypes of the input... https://github.com/QuantEcon/QuantEcon.py/blob/66a4451614c895baa0e6ee30345564f38add59c2/quantecon/game_theory/polymatrix_game.py#L220-L223
(`matrix_builder` is created as an nparray of `float`.)

PR #764, opened 2025-02-12
RFC: support_enumeration: Reduce intermediate array creation

oyamad:
Same as QuantEcon/GameTheory.jl#184


mmcky:
thanks @oyamad

PR #766, opened 2025-05-18
RFC: game_theory: Use numba.typed.Dict in `_vertex_enumeration_gen`

oyamad:
I let ChatGPT read the existing code of [`vertex_enumeration.py`](https://github.com/QuantEcon/QuantEcon.py/blob/main/quantecon/game_theory/vertex_enumeration.py) and got the following suggestion:
* The current [`_vertex_enumeration_gen`](https://github.com/QuantEcon/QuantEcon.py/blob/23a8a9bfd5719ca358d7b9118cd19fad7dc2fe12/quantecon/game_theory/vertex_enumeration.py#L88-L123) has complexity `O(V_0 * V_1)`, where `V_i` is the number of vertices of the best response polytope of player `i`.
* One should use Dict in place of the inner loop, which will reduce the complexity to `O(V_0 + V_1)`.

Benchmarks with random n x n games showed significant speedups, and larger speedups for larger games: about 1.3x faster for n=12, 2x faster for n=14, 15, and 4x faster for n=17. (And no additional overheads for small size games.)

### DIFF
--- a/quantecon/game_theory/vertex_enumeration.py
+++ b/quantecon/game_theory/vertex_enumeration.py
@@ -11,7 +11,8 @@ Tardos, and V. Vazirani eds., Algorithmic Game Theory, 2007.
 """
 import numpy as np
 import scipy.spatial
-from numba import jit, guvectorize
+from numba import jit, guvectorize, types
+from numba.typed import Dict
 
 
 def vertex_enumeration(g, qhull_options=None):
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @oyamad

mmcky:
thanks @oyamad I have now release `v0.8.1`

PR #837, opened 2026-06-21
Add tests for Kalman.stationary_coefficients

### LINKED ISSUE #128: New Kalman test needed
Added the method `stationary_coefficients` to the Kalman class.  Tests needed.  (TJS might have ideas on a suitable test, if necessary.  Otherwise something straightforward will do.)

hinayuki64:
## Summary
- Add `TestKalmanStationaryCoefficients` to cover `Kalman.stationary_coefficients`
- Verify MA and VAR coefficient recursion against a reference implementation
- Exercise both Riccati solvers (`doubling`, `qz`) and invalid `coeff_type` handling

Closes #128

## Test plan
- [x] `pytest quantecon/tests/test_kalman.py -v`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/quantecon/tests/test_kalman.py
+++ b/quantecon/tests/test_kalman.py
@@ -29,7 +29,9 @@ def _knowing_forecasts_of_others_kalman():
     https://python-advanced.quantecon.org/knowing_forecasts_of_others.html
 
     The lecture sets H=None (no observation noise). Kalman.stationary_values
-    requires H, so a negligible diagonal H is added for numerical use only.
+    requires H, so a small diagonal H is added for numerical use only.
+    Use 1e-4 rather than smaller values to avoid singular-matrix errors in
+    scipy.linalg.inv on some platforms (e.g. Python 3.14 on CI).
     """
     beta, rho, b = 0.9, 0.8, 1.5
     sigma_v, sigma_e = 0.5, 0.6
@@ -65,7 +67,7 @@ def _knowing_forecasts_of_others_kalman():
         [1., 0., 0., 0., 1., 0.],
         [1., 0., 0., 0., 0., 0.],
     ])
-    H = np.eye(G.shape[0]) * 1e-8
+    H = np.eye(G.shape[0]) * 1e-4
 
     return Kalman(LinearStateSpace(A, C, G, H))
 
--- a/quantecon/tests/test_kalman.py
+++ b/quantecon/tests/test_kalman.py
@@ -7,6 +7,7 @@ import pytest
 from numpy.testing import assert_allclose
 from quantecon import LinearStateSpace
 from quantecon import Kalman
+from quantecon import solve_discrete_riccati
 
 
 RICCATI_METHODS = ['doubling', 'qz']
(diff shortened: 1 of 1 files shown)


oyamad:
~@hinayuki64 #838 merged, will you rebase this PR on `main`?~ No never mind, already done..

oyamad:
@hinayuki64 Thanks for the PR. I asked ChatGPT to review this: It suggested to add a test based on the QuantEcon lecture "[Knowing the Forecasts of Others](https://python-advanced.quantecon.org/knowing_forecasts_of_others.html)" which involves asymmetric matrices, in addition to the current test which is very symmetric. Would you mind working on that too?

hinayuki64:
Thank you for the suggestion. I added a test case based on the pooling-equilibrium example from ["Knowing the Forecasts of Others"](https://python-advanced.quantecon.org/knowing_forecasts_of_others.html), alongside the symmetric toy model. The lecture has no observation noise (H=None), but Kalman requires H and the model is nearly singular without it, so I used a small diagonal H (1e-4 * I) for numerical stability.

oyamad:
@hinayuki64 Sorry, I was not understanding what ChatGPT said well, but (instead of what you implemented) we may want to implement the simple process described in [37.5.3](https://python-advanced.quantecon.org/knowing_forecasts_of_others.html#two-noisy-signals), which involves asymmetric `(n, k)` (= `(1, 2)`) and nontrivial `H` (= `np.eye(2) * sigma_e`).

hinayuki64:
@oyamad OK, I switched the asymmetric test to the [section 37.5.3](https://python-advanced.quantecon.org/knowing_forecasts_of_others.html#two-noisy-signals) two-noisy-signals model (n, k) = (1, 2) with H = np.eye(2) * sigma_e, instead of the pooling-equilibrium state space I had added earlier.

quantecon/tests/test_kalman.py
```diff
@@ -3,11 +3,67 @@
...
+    """
+    Two-noisy-signals filter from section 37.5.3 of the QuantEcon lecture
+    "Knowing the Forecasts of Others":
+    https://python-advanced.quantecon.org/knowing_forecasts_of_others.html#two-noisy-signals
+
+    State: theta_{t+1} = rho theta_t + v_t
+    Observations: w_t = [1, 1]' theta_t + [e_1t, e_2t]'
+    """
+    A = np.array([[rho]])
+    C = np.array([[np.sqrt(sigma_v)]])
```
oyamad:
```suggestion
    C = np.array([[sigma_v]])
```
`_expected_stationary_coefficients` seems not independent enough to catch this regression, as it computes the answers the computed `K_infinity`.

hinayuki64:
Thanks — updated C to `[[sigma_v]]` as suggested, and the asymmetric tests now compare `stationary_coefficients` against coefficients built from the lecture's closed-form K (37.25–37.26), with a separate check that `K_infinity` matches that reference.

quantecon/tests/test_kalman.py
```diff
@@ -3,9 +3,80 @@
...
+    Stationary Kalman gain from lecture eqs. (37.25)-(37.26), independent of
+    Kalman.stationary_values.
+    """
+    p = solve_discrete_riccati(
+        np.array([[rho]]),
+        np.array([[np.sqrt(2.)]]),
+        np.array([[sigma_v ** 2]]),
+        np.array([[sigma_e ** 2]]),
+        np.zeros((1, 1)),
+    ).item()
```
oyamad: Just one comment:
`p` is the solution to a quadratic equation, so can be explicitly solved by hand (using `solve_discrete_riccati` is an overkill?).

hinayuki64: I replaced `solve_discrete_riccati` with the explicit quadratic formula for p from eq. (37.26).

quantecon/tests/test_kalman.py
```diff
@@ -84,3 +162,86 @@ def test_update_nonstationary(self):
...
+                    A, G, self.K_ref, k, j, 'var')
+                _assert_coeff_lists_equal(actual, expected)
+
+    def test_coefficients_are_not_diagonal(self):
+        """Sanity check: (n, k) = (1, 2) model yields non-diagonal MA coefficients."""
+        expected = _expected_stationary_coefficients(
+            self.kf.ss.A, self.kf.ss.G, self.K_ref, self.kf.ss.k, 1, 'ma')
+        psi_1 = expected[1]
+        assert psi_1.shape == (2, 2)
+        assert not np.allclose(psi_1, np.diag(np.diag(psi_1)))
```
oyamad: Final comment:
I think this function `test_coefficients_are_not_diagonal` can be deleted, which only tests the expected answer (rather than the output of library code).

hinayuki64: Thanks. I removed it based on your feedback.

oyamad (approved):
@hinayuki64 Thanks!
I think this is ready to merge.

mmcky:
thanks @hinayuki64 and @oyamad

PR #842, opened 2026-06-27
BUG: Fix DLE.compute_sequence under NumPy >= 2.4 (#839)

### LINKED ISSUE #839: DLE.compute_sequence fails under NumPy 2.4 (setting an array element with a sequence)
## Summary

`quantecon.DLE.compute_sequence` raises under **NumPy 2.4** when it computes the asset-price terms `R1_Price`, `R2_Price`, `R5_Price`. The same code runs fine under NumPy 2.3.x, so this is a regression triggered by NumPy tightening size-1-array → scalar conversion on assignment.

## Environment

- `quantecon` latest (installed via `pip install --upgrade quantecon`)
- `numpy==2.4.6` (ships with `anaconda=2026.06`) — **fails**
- `numpy==2.3.5` (ships with `anaconda=2025.12`) — works
- Python 3.13

## Traceback

```
----> 1 econ1.compute_sequence(x0, ts_length=100)

File .../site-packages/quantecon/_dle.py:215, in DLE.compute_sequence(self, x0, ts_length, Pay)
    213     self.R5_Price = np.empty((ts_length + 1, 1))
    214     for i in range(ts_length + 1):
--> 215         self.R1_Price[i, 0] = self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
    216             self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])

TypeError: only 0-dimensional arrays can be converted to Python scalars
The above exception was the direct cause of the following exception:
ValueError: setting an array element with a sequence.
```

## Root cause

In `compute_sequence`, the selector `e1` is a 2-D row vector:

```python
e1 = np.zeros((1, self.nc))
e1[0, 0] = 1
```

so each price expression `e1 @ self.Mc @ ... @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])` evaluates to a **size-1 array** (shape `(1,)`), not a Python scalar. Assigning that into the scalar slot `self.R1_Price[i, 0]` used to work because NumPy silently converted a size-1 array to a scalar, but NumPy 2.4 now raises (`only 0-dimensional arrays can be converted to Python scalars` → `setting an array element with a sequence`).

## Suggested fix

Coerce each price term to a scalar with `.item()` (works regardless of `e1`'s shape):

```python
for i in range(ts_length + 1):
    self.R1_Price[i, 0] = (self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
        self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])).item()
    self.R2_Price[i, 0] = (self.beta**2 * (e1 @ self.Mc @
        np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
    self.R5_Price[i, 0] = (self.beta**5 * (e1 @ self.Mc @
        np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
```

Alternatively, define `e1` as 1-D (`e1 = np.zeros(self.nc); e1[0] = 1`) so the matmul chain reduces to a 0-d scalar.

## Downstream impact

This breaks 5 lectures in `lecture-python-advanced.myst` under `anaconda=2026.06` (numpy 2.4.6): `cattle_cycles`, `growth_in_dles`, `irfs_in_hall_model`, `lucas_asset_pricing_dles`, `permanent_income_dles`. See the validation in QuantEcon/lecture-python-advanced.myst#346. It affects any `DLE` whose dimensions make the price expression size-1 (lectures that don't reach the price path, e.g. `gorman_heterogeneous_households`, `hs_invertibility_example`, `rosen_schooling_model`, are unaffected).

mmcky:
## Summary

Fixes #839. `DLE.compute_sequence` raises under **NumPy 2.4** when it computes the asset-price terms `R1_Price`, `R2_Price`, `R5_Price`. The same code runs under NumPy 2.3.x, so this is a 2.3 → 2.4 regression.

## Root cause

Each price expression evaluates to a **size-1 array**, not a Python scalar: `self.beta` is a `(1, 1)` array and `e1` is a row vector, so the matmul chain reduces to shape `(1,)` or `(1, 1)`. Assigning such an array into a scalar slot like `self.R1_Price[i, 0]` used to work because NumPy silently converted a size-1 `ndim > 0` array to a scalar. That conversion was deprecated in NumPy 1.25 and became a hard error in NumPy 2.4 (listed under *Expired deprecations* in the [2.4.0 release notes](https://numpy.org/doc/2.4/release/2.4.0-notes.html)), so the assignment now raises `ValueError: setting an array element with a sequence` (underlying `TypeError: only 0-dimensional arrays can be converted to Python scalars`).

## Fix

Coerce each price expression to a scalar with `.item()` before assignment. This works regardless of the intermediate shape and reproduces the pre-2.4 values exactly.

Note: redefining `e1` as 1-D (suggested in the issue discussion) is **not sufficient on its own** — because `self.beta` is itself a `(1, 1)` array, `self.beta * e1 @ …` re-introduces a 2-D shape and the result stays `ndim > 0`. I verified this empirically: under both the current 2-D `e1` and the proposed 1-D `e1`, `R1` is shape `(1,)` and `R2`/`R5` are shape `(1, 1)`. `.item()` addresses the conversion at the point of assignment and is robust to all of these shapes.

The same size-1 assignment pattern also exists in the optional `Pay` branch (`Pay_Price` / `Pay_Gross`), which is only reached when a `Pay` array is passed. It is fixed here too, so `compute_sequence` is robust regardless of arguments.

## Verification

The validated MWE from the issue (Hall's permanent-income economy) now runs, and the `.item()` path reproduces the pre-regression values exactly:

| numpy | before this PR | after this PR |
| --- | --- | --- |
| 2.3.5 | ✅ runs — `R1_Price[:3] = [0.95238095, …]` | ✅ runs — identical values |
| 2.4.6 | ❌ raises | ✅ runs — identical values |

## Tests

`compute_sequence` previously had **no test coverage**. This PR adds two regression tests to `TestDLE`:

- `test_compute_sequence` — exercises the default (risk-free price) path, checks shapes and finiteness, and anchors `R1/R2/R5_Price[0, 0]` to the closed form `β**J` at `t = 0`.
- `test_compute_sequence_with_pay` — exercises the `Pay` branch (`Pay_Price` / `Pay_Gross`).

All 10 tests in `quantecon/tests/test_dle.py` pass locally (NumPy 2.3.5, Python 3.13).

## Downstream impact

This unblocks 5 lectures in `lecture-python-advanced.myst` under `anaconda=2026.06` (numpy 2.4.6): `cattle_cycles`, `growth_in_dles`, `irfs_in_hall_model`, `lucas_asset_pricing_dles`, `permanent_income_dles`.

### DIFF
--- a/quantecon/_dle.py
+++ b/quantecon/_dle.py
@@ -211,17 +211,22 @@ class DLE(object):
         self.R1_Price = np.empty((ts_length + 1, 1))
         self.R2_Price = np.empty((ts_length + 1, 1))
         self.R5_Price = np.empty((ts_length + 1, 1))
+        # Hoist the loop-invariant terms out of the loop: ``e1 @ self.Mc`` and
+        # the matrix powers of ``self.A0`` do not depend on ``i``.
+        eMc = e1 @ self.Mc
+        A0_1 = np.linalg.matrix_power(self.A0, 1)
+        A0_2 = np.linalg.matrix_power(self.A0, 2)
+        A0_5 = np.linalg.matrix_power(self.A0, 5)
         # Each price expression evaluates to a size-1 array (``self.beta`` is a
-        # (1, 1) array and ``e1`` is a row vector). Extract the scalar with
+        # (1, 1) array and ``eMc`` is a row vector). Extract the scalar with
         # ``.item()`` before assigning: NumPy >= 2.4 raises instead of silently
         # converting an ``ndim > 0`` array to a scalar on assignment. See #839.
         for i in range(ts_length + 1):
-            self.R1_Price[i, 0] = (self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
-                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])).item()
-            self.R2_Price[i, 0] = (self.beta**2 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
-            self.R5_Price[i, 0] = (self.beta**5 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
+            xi = xp[:, i]
+            denom = (eMc @ xi).item()
+            self.R1_Price[i, 0] = (self.beta * eMc @ A0_1 @ xi).item() / denom
+            self.R2_Price[i, 0] = (self.beta**2 * (eMc @ A0_2 @ xi)).item() / denom
+            self.R5_Price[i, 0] = (self.beta**5 * (eMc @ A0_5 @ xi)).item() / denom
 
         # === Gross rates of return on 1-period risk-free bonds === #
         self.R1_Gross = 1 / self.R1_Price
--- a/quantecon/_dle.py
+++ b/quantecon/_dle.py
@@ -211,13 +211,17 @@ class DLE(object):
         self.R1_Price = np.empty((ts_length + 1, 1))
         self.R2_Price = np.empty((ts_length + 1, 1))
         self.R5_Price = np.empty((ts_length + 1, 1))
+        # Each price expression evaluates to a size-1 array (``self.beta`` is a
+        # (1, 1) array and ``e1`` is a row vector). Extract the scalar with
+        # ``.item()`` before assigning: NumPy >= 2.4 raises instead of silently
+        # converting an ``ndim > 0`` array to a scalar on assignment. See #839.
         for i in range(ts_length + 1):
-            self.R1_Price[i, 0] = self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
-                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])
-            self.R2_Price[i, 0] = self.beta**2 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])
-            self.R5_Price[i, 0] = self.beta**5 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])
+            self.R1_Price[i, 0] = (self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
+                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])).item()
+            self.R2_Price[i, 0] = (self.beta**2 * (e1 @ self.Mc @
+                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
+            self.R5_Price[i, 0] = (self.beta**5 * (e1 @ self.Mc @
+                np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
 
         # === Gross rates of return on 1-period risk-free bonds === #
         self.R1_Gross = 1 / self.R1_Price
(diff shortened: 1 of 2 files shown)


quantecon/_dle.py
```diff
@@ -211,13 +211,17 @@ def compute_sequence(self, x0, ts_length=None, Pay=None):
...
-            self.R1_Price[i, 0] = self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
-                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])
-            self.R2_Price[i, 0] = self.beta**2 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])
-            self.R5_Price[i, 0] = self.beta**5 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])
+            self.R1_Price[i, 0] = (self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
+                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])).item()
+            self.R2_Price[i, 0] = (self.beta**2 * (e1 @ self.Mc @
+                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
```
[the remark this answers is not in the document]
mmcky: Good catch — fixed in 4737381. Hoisted `e1 @ self.Mc` and the matrix powers `matrix_power(self.A0, k)` out of the loop (all loop-invariant), and now compute the denominator `e1 @ self.Mc @ xp[:, i]` once per `i` instead of three times. Verified the results are bit-identical to the previous code on a fixed trajectory.

mmcky:
A note on commit 4737381 (the loop-invariant hoist from Copilot's suggestion): the main driver here is **readability**, not performance. Pulling `e1 @ self.Mc` and the `matrix_power(self.A0, k)` terms out of the loop, and computing the denominator once per `i`, makes it explicit that those quantities don't depend on the iteration index. The actual speedup is negligible in practice — the matrices are tiny and the loop is short, and this loop is a small fraction of `compute_sequence`'s total cost (dominated by the LQ solve, the simulation, and the Lyapunov solve). The change is bit-identical to the previous code (verified on a fixed trajectory), so it carries no numerical risk; the #839 fix itself lives in 68ba04f.

quantecon/_dle.py
```diff
@@ -211,17 +211,22 @@ def compute_sequence(self, x0, ts_length=None, Pay=None):
...
         for i in range(ts_length + 1):
-            self.R1_Price[i, 0] = (self.beta * e1 @ self.Mc @ np.linalg.matrix_power(
-                self.A0, 1) @ xp[:, i] / (e1 @ self.Mc @ xp[:, i])).item()
-            self.R2_Price[i, 0] = (self.beta**2 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 2) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
-            self.R5_Price[i, 0] = (self.beta**5 * (e1 @ self.Mc @
-                np.linalg.matrix_power(self.A0, 5) @ xp[:, i]) / (e1 @ self.Mc @ xp[:, i])).item()
+            xi = xp[:, i]
+            denom = (eMc @ xi).item()
+            self.R1_Price[i, 0] = (self.beta * eMc @ A0_1 @ xi).item() / denom
```
oyamad: (Pointed out by ChatGPT)
`(self.beta * eMc @ A0_1 @ xi).item() / denom` results in a different outcome from previously when `denom == 0`: The behavior doesn't change if it is `(self.beta * eMc @ A0_1 @ xi / denom).item()`.

(Python scalar divided by 0 versus NumPy scalar divided by 0)

oyamad (commented):
* My opinion is that conceptually scalar objects (such as `beta`) should be implemented as scalars; conceptually 1-d vectors should be implemented as 1-d arrays. The whole code in this file is very "matrix-oriented", which is very different from the other part of this library, and to me is very confusing. But this would require a global full re-write of the whole code.
* So I understand the preference for keeping this PR local, only applying a minimal "workaround" that removes the error.

quantecon/_dle.py
mmcky: Good catch (and thanks to ChatGPT) — fixed in 37daf9a. I moved `.item()` outside the division so the division is performed by NumPy rather than on two extracted Python floats. This preserves the pre-2.4 behaviour: a zero denominator now yields `inf`/`nan` with a `RuntimeWarning` rather than raising `ZeroDivisionError`. `denom` is also kept as a NumPy array rather than being extracted with `.item()` first, so these three lines now match the `Pay` branch, which already divides inside `.item()`. The 10 tests in `test_dle.py` still pass and values are unchanged for non-zero denominators.

oyamad (approved):
@mmcky Thanks!

mmcky:
thanks @oyamad -- once this goes green I will merge.

PR #846, opened 2026-06-28
TEST: De-flake timing tests with lower-bound assertions

mmcky:
## Summary

Fixes the flaky timing tests that have been intermittently reddening CI on green code (twice on `main`, once on #845). This is the second of the two root causes found while investigating #842's CI failure — the first (Coveralls 422 race + `fail-fast` cascade) is addressed in #845.

## Root cause

The whole `test_timing.py` suite verifies the timing utilities (`tic/tac/toc`, `loop_timer`, `Timer`, `timeit`) by sleeping for a known duration and asserting the *measured* time stays within an **upper** bound:

> `assert_allclose(measured, expected_sleep, atol=0.05, rtol=2)`

That is the wrong invariant for wall-clock timing. `time.sleep(t)` only guarantees the process is suspended for *at least* `t` seconds; a busy or descheduled CI runner can make the measured time arbitrarily larger. The tightest case, `test_timeit_lambda_function` (expected ≈ 0.05 s, ceiling ≈ 0.20 s), is the one that keeps tripping — e.g. measured `0.207 s` vs ceiling `0.20 s` on a loaded macOS runner.

## Fix

Replace the ~14 upper-bound assertions with a lower-bound check, `assert_at_least(measured, expected)`, which asserts the utilities **do not under-report** elapsed time (`measured >= expected * 0.9`). Scheduler overshoot — the actual source of the flakiness — is no longer treated as a failure, while a genuinely broken timer (returning zero / far too little) is still caught.

| Kept as-is | Why |
| --- | --- |
| `assert result['minimum'] <= result['average'] <= result['maximum']` | A relationship invariant, not wall-clock dependent. |
| `assert_(average_time >= average_of_best)` | Same — relationship between aggregates. |
| `test_timeit_single_run` exact-equality asserts | Compare fields to each other, not to absolute time. |

The now-unused `assert_allclose` import and the dead `rtol`/`atol` locals are removed (keeps the CI `flake8 --select F401` check green).

## Verification

- `flake8 --select F401,F405,E231 quantecon/util/tests/test_timing.py` — clean.
- `pytest quantecon/util/tests/test_timing.py` — **27 passed**, run repeatedly with stable results.

## Relationship to the other PRs

- #845 stops a single flake from cancelling the whole matrix (`fail-fast: false`) and removes the Coveralls race; this PR removes the flake itself. Together they should make CI reliably green.
- #842 (the DLE / NumPy 2.4 fix that started this) is functionally green; it just needs #845's hardened workflow once that merges.

### DIFF
--- a/quantecon/util/tests/test_timing.py
+++ b/quantecon/util/tests/test_timing.py
@@ -26,6 +26,28 @@ def assert_at_least(measured, expected):
         f"({TIMING_LOWER_BOUND:g} * expected {expected:.4f}s)")
 
 
+# Above this multiple of the expected sleep, a measurement can only be a unit
+# bug (``elapsed`` stored in ms/µs is ~1000x/1e6x larger), not scheduler
+# overshoot (a few x at most), so the ceiling separates the two without flaking.
+TIMING_UNIT_CEILING = 100
+
+
+def assert_in_seconds(measured, expected):
+    """Assert ``measured`` is in seconds (~``expected``), not a larger unit.
+
+    Lower bound: the timer did not under-report. Upper bound: a ceiling no real
+    ``expected``-second sleep can reach, but which a unit leak into
+    ``Timer.elapsed`` (ms/µs) always trips -- the invariant the unit tests
+    guard.
+    """
+    lower_bound = expected * TIMING_LOWER_BOUND
+    upper_bound = expected * TIMING_UNIT_CEILING
+    assert lower_bound <= measured < upper_bound, (
+        f"measured {measured:.4f}s is outside the seconds range "
+        f"[{lower_bound:.4f}, {upper_bound:.4f}) around expected {expected:.4f}s "
+        f"(did the display unit leak into elapsed?)")
+
+
 class TestTicTacToc:
     def setup_method(self):
         self.h = 0.1
@@ -99,16 +121,16 @@ class TestTimer:
         with Timer(verbose=False) as timer_sec:
             time.sleep(self.sleep_time)
         expected_sec = self.sleep_time
-        assert_at_least(timer_sec.elapsed, expected_sec)
+        assert_in_seconds(timer_sec.elapsed, expected_sec)
         
         # Timer always stores elapsed time in seconds regardless of display unit
         with Timer(unit="milliseconds", verbose=False) as timer_ms:
             time.sleep(self.sleep_time)
-        assert_at_least(timer_ms.elapsed, expected_sec)
+        assert_in_seconds(timer_ms.elapsed, expected_sec)
         
         with Timer(unit="microseconds", verbose=False) as timer_us:
             time.sleep(self.sleep_time)
-        assert_at_least(timer_us.elapsed, expected_sec)
+        assert_in_seconds(timer_us.elapsed, expected_sec)
         
     def test_invalid_unit(self):
         """Test that invalid units raise ValueError."""
@@ -248,9 +270,9 @@ class TestTimer:
         
         # All results should be in seconds regardless of display unit
         for run_time in result_ms['elapsed']:
-            assert_at_least(run_time, self.sleep_time)
+            assert_in_seconds(run_time, self.sleep_time)
         for run_time in result_us['elapsed']:
-            assert_at_least(run_time, self.sleep_time)
+            assert_in_seconds(run_time, self.sleep_time)
 
     def test_timeit_stats_only(self):
         """Test timeit with stats_only option."""
--- a/quantecon/util/tests/test_timing.py
+++ b/quantecon/util/tests/test_timing.py
@@ -20,8 +20,10 @@ TIMING_LOWER_BOUND = 0.9
 
 def assert_at_least(measured, expected):
     """Assert ``measured`` seconds is not meaningfully below ``expected``."""
-    assert measured >= expected * TIMING_LOWER_BOUND, (
-        f"measured {measured}s is well below the expected ~{expected}s sleep")
+    lower_bound = expected * TIMING_LOWER_BOUND
+    assert measured >= lower_bound, (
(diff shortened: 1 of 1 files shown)


quantecon/util/tests/test_timing.py
```diff
@@ -4,11 +4,26 @@
...
+# bound (that the timing utilities do not under-report the elapsed time) rather
+# than an upper bound on wall-clock duration, which is inherently flaky on
+# shared CI runners. See the discussion in GH #845.
+TIMING_LOWER_BOUND = 0.9
+
+
+def assert_at_least(measured, expected):
+    """Assert ``measured`` seconds is not meaningfully below ``expected``."""
+    assert measured >= expected * TIMING_LOWER_BOUND, (
+        f"measured {measured}s is well below the expected ~{expected}s sleep")
```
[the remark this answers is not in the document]
mmcky: Fixed in 809e24c — the failure message now reports the computed lower bound and the factor, e.g. `measured 0.0200s is below the lower bound 0.0450s (0.9 * expected 0.0500s)`, with floats formatted to 4 dp.

mmcky:
## Note for reviewers — what this PR changes and a fidelity check

This started as a pure de-flake but, after review, also **strengthens** the suite. Summary of the three commits and the reasoning, plus an assertion-by-assertion fidelity audit so it's clear nothing was silently weakened.

### What changed

| Commit | Change |
| --- | --- |
| `7d9e6b6` | Replace the wall-clock **upper-bound** assertions (`assert_allclose(measured, expected, atol=0.05, rtol=2)`) with lower-bound checks. `time.sleep(t)` only guarantees *at least* `t`; a busy CI runner can oversleep arbitrarily, so an upper bound on measured duration is flaky by construction (it tripped on `test_timeit_lambda_function` 3× recently). |
| `809e24c` | Make the `assert_at_least` failure message report the computed bound and factor, floats to 4 dp (Copilot suggestion). |
| `9af5760` | Restore a **bounded** check (`assert_in_seconds`) for the two unit tests, which exist to verify `Timer.elapsed` stays in seconds regardless of the display `unit` — a pure lower bound would miss a unit-leak regression. |

### The two helpers

| Helper | Used by | Asserts |
| --- | --- | --- |
| `assert_at_least(x, e)` | generic timing tests (tic/tac/toc, `loop_timer`, `Timer.elapsed`, `timeit` runs/average) | `x >= 0.9·e` — the timer does not **under**-report |
| `assert_in_seconds(x, e)` | `test_timer_units`, `test_timeit_different_units` | `0.9·e <= x < 100·e` — `elapsed` is in **seconds**, not ms (×1000) / µs (×1e6) |

The `×100` ceiling sits ~25× above the worst overshoot actually observed on CI and 10× below a millisecond leak — so it catches a unit regression with no realistic flake risk.

### Fidelity audit — no loss, net gain

The originals used `rtol=2`, so `assert_allclose(x, e, atol=0.05, rtol=2)` resolves to `x ∈ [−(e+0.05), 3e+0.05]`. The lower edge is negative, i.e. the originals were **effectively upper-bound-only** with no meaningful lower bound.

- **Newly caught** (the originals silently passed these): a timer returning `0` / not accumulating — `assert_allclose(0, 0.1, …)` passes but `0 >= 0.09` fails; and `toc()` returning the last interval (~0.1) instead of the cumulative total (~0.3) — now fails the `>= 0.27` lower bound.
- **Retained**: the unit-invariance guard — the one place an upper bound genuinely mattered — still trips on a ms/µs leak, and now also has a lower bound.
- **Unchanged**: every relationship/structural/contract assertion — `average_time >= average_of_best`, `minimum <= average <= maximum`, the single-run equalities, all `is not None` / `len == N` / `isinstance` checks, the `ValueError`/validation tests, and the whole `TestGlobalPrecision` class.
- **Only removed**: loose wall-clock upper bounds on default-unit timing, which targeted an implausible bug class and were themselves the flake source.

**Pre-existing caveat (not introduced here):** `test_timeit_lambda_function` doesn't strongly verify the `0.5` multiplier (sleeping `0.1` instead of `0.05` passes both the old and new assertion). Unchanged by this PR; noting it for completeness.

Net: the suite is no longer flaky **and** catches more realistic regressions than before.

quantecon/util/tests/test_timing.py
```diff
@@ -4,11 +4,50 @@
...
+TIMING_UNIT_CEILING = 100
+
+
+def assert_in_seconds(measured, expected):
+    """Assert ``measured`` is in seconds (~``expected``), not a larger unit.
+
+    Lower bound: the timer did not under-report. Upper bound: a ceiling no real
+    ``expected``-second sleep can reach, but which a unit leak into
+    ``Timer.elapsed`` (ms/µs) always trips -- the invariant the unit tests
+    guard.
```
oyamad: A comment from ChatGPT:

> Non-blocking nit: the comment/docstring could avoid saying “no real sleep can reach” because sleep overshoot is theoretically unbounded; “not expected under normal CI scheduling” would be more precise. Not worth blocking the PR.

oyamad (approved):
@mmcky  Thanks! Looks good to me. The comment from ChatGPT below is just optional (not blocking).

quantecon/util/tests/test_timing.py
mmcky: Thanks @oyamad (and ChatGPT) — good catch, addressed in f8c03ad.

The docstring and the `TIMING_UNIT_CEILING` comment no longer claim the ceiling is one "no real sleep can reach". They now describe overshoot as "unbounded in theory but a few x at most under normal CI scheduling" and the ceiling as one "not expected to be reached ... under normal CI scheduling", so the wording is honest about the (negligible) residual flake risk rather than implying a hard guarantee. Comment/docstring only — no behavior change, and the `100×` ceiling is unchanged. `flake8 --select F401,F405,E231` clean and 27 passed locally.

PR #849, opened 2026-07-06
CI: Fix flake8 --select so E231/F405 are actually enforced

mmcky:
## Summary

The flake8 gate in `ci.yml` has never enforced two of the three codes it names. The step is written with spaces after the commas:

```
flake8 --select F401, F405, E231 quantecon
```

The shell splits that into separate arguments, so flake8 receives `--select F401,` (just F401, with a trailing comma) followed by `F405,`, `E231`, and `quantecon` as **positional paths**. The bogus `F405,` / `E231` paths would normally raise file-not-found errors, but those aren't F401 so the `--select` filter drops them — and the step exits 0. Net effect: **only F401 has been enforced; E231 and F405 were silently skipped.**

## Fix

Use the unambiguous form so all three codes are selected against `quantecon`:

```
flake8 --select=F401,F405,E231 quantecon
```

## Pre-existing violations surfaced by turning the gate on

Enabling the gate revealed **6 pre-existing E231 violations** (0 F401, 0 F405) that had been slipping through. Since a gate fix that reddens CI is no fix, those are corrected in the same PR — all are one-character whitespace additions, no behavior change:

| File | Change |
| --- | --- |
| `_compute_fp.py`, `util/numba.py` | `{'cache':True}` → `{'cache': True}` |
| `_gridtools.py` | `range(1,n+1)` → `range(1, n+1)` |
| `tests/test_gridtools.py` | `{order:cartesian(...)}` → `{order: cartesian(...)}` |
| `markov/tests/test_approximation.py` | space after `,` in the unpack continuation and in `map(np.array, (...))` |

Codes outside the selected set (E501, W291, E128, …) remain unenforced exactly as before — this PR does not widen the gate beyond the F401/F405/E231 it was always meant to check.

## Verification

- `flake8 --select=F401,F405,E231 quantecon` — clean (exit 0).
- `pytest` on the affected modules (`test_approximation.py`, `test_gridtools.py`, `test_numba.py`) — 39 passed.

Found while reviewing the CI-hardening work in #845.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
     - name: flake8 Tests
       shell: bash -l {0}
       run: |
-        flake8 --select F401, F405, E231 quantecon
+        flake8 --select=F401,F405,E231 quantecon
     - name: Run Tests (pytest)
       shell: bash -l {0}
       run: |
--- a/quantecon/_compute_fp.py
+++ b/quantecon/_compute_fp.py
@@ -357,7 +357,7 @@ def _square_sum(a):  # pragma: no cover
     pass
 
 
-@overload(_square_sum, jit_options={'cache':True})
+@overload(_square_sum, jit_options={'cache': True})
 def _square_sum_ol(a):
     if isinstance(a, types.Number):
         return lambda a: a**2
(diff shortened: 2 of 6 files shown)


mmcky:
@oyamad this is pretty straight forward and passing so I'll merge.

PR #836, opened 2026-06-18
DOCS: Add contributing entrypoint

wenlei0603:
## Summary

- add a root `CONTRIBUTING.md` entrypoint for GitHub's contributing-guidelines link
- point contributors to the rendered Read the Docs guide and the source `docs/source/contributing.rst`

## Context

This addresses the root `CONTRIBUTING.md` suggestion from #834 without duplicating the full contributing guide in two places.

## Validation

- Confirmed `docs/source/contributing.rst` exists in the repository
- Confirmed https://quanteconpy.readthedocs.io/en/latest/contributing.html returns HTTP 200

Issue #834, opened 2026-06-10
Review developer/contributing documentation for completeness


mmcky:
## Context

The quantecon.org website is removing its developer instruction pages (see QuantEcon/website#201), since developer docs were migrated to this repo in #740. The website's Contribute buttons now point at the rendered contributing guide:

https://quanteconpy.readthedocs.io/en/latest/contributing.html

This is a request to review that the developer docs here fully cover what the website pages used to provide.

## Removed website pages and their coverage

| Removed page | Topics | Covered by `docs/source/contributing.rst`? |
|---|---|---|
| `/python-developers/` | Conda dev env, tests, documentation, **package structure**, **the QuantEcon.py API** | Partially — the *Package structure* and *API* sections have no equivalent |
| `/wiki-py-conda-dev-env/` | Step-by-step Conda development environment setup | Summarised only |
| `/wiki-py-docstrings/` | Detailed numpydoc docstring guide (~280 lines) | Unclear — worth checking nothing of value was lost |
| `/wiki-py-unit-testing/` | Unit testing guide — test naming, assertion methods, examples (~130 lines) | Unclear — worth checking nothing of value was lost |

The full text of the removed pages is retrievable from the website repo history if anything needs porting.

## Suggestion: add a root `CONTRIBUTING.md`

The repo currently has no root `CONTRIBUTING.md` (or `.github/CONTRIBUTING.md`), so GitHub shows no contributing-guidelines link on new issues/PRs. A short stub that links to the rendered readthedocs guide would be enough.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #850, opened 2026-07-07
DOC: Refresh contributing guide and fix stale docs build config

mmcky:
## Summary

Partner PR to #836 (which adds the root `CONTRIBUTING.md` entrypoint) and addresses part of #834 (review the contributing/developer docs for completeness). #836 points readers at `docs/source/contributing.rst`, so this refreshes that guide and, off the back of a fuller pass over `docs/`, fixes stale/broken items in the docs build config and repoints the API reference at the real implementation modules instead of the deprecated compatibility shims.

## What changed

**`docs/source/contributing.rst` — refresh the developer guide**

| Fix | Before → After |
| --- | --- |
| Factual error | "a preview of the docs is provided as one of the **GitHub Actions**" → built by **Read the Docs** (there is no docs-building GHA workflow — only `ci.yml`/`ci_np2.yml`, neither of which builds docs) |
| Missing commands | The guide described the workflow in prose but never showed *how*. Added concrete steps: clone + `conda env create -f environment.yml` (the repo already ships this), an editable `flit install --symlink`, running tests (`pytest`), style checks (the exact CI form `flake8 --select=F401,F405,E231 quantecon`), and building docs (`pip install -r docs/rtd-requirements.txt` then `make html`) |
| Links | `http://` → `https://` for the issue tracker and Discourse; stripped trailing whitespace |

**`docs/qe_apidoc.py` + regenerated stubs — stop documenting the deprecated shims**

The generator globbed `../quantecon/[a-z0-9]*.py` for the base-level "tools" pages, which matches the lowercase **compatibility shims** (`arma.py`, `lqcontrol.py`, …) and excludes the real implementation modules (`_arma.py`, `_lqcontrol.py`, …). So the API reference was built with e.g. `.. automodule:: quantecon.arma` — modules marked *"not meant for public use and will be removed v0.8.0"*, which carry no module docstrings and emit a `DeprecationWarning` on every attribute access during the docs build.

- New `base_module_target()` maps a base-level module to `_<name>` when `quantecon/_<name>.py` exists (21 modules); genuinely public modules (`distributions`, `quad`) are unchanged. Page names and toctrees are untouched — only the `automodule` target moves, and the implementation modules' real docstrings now render.
- The 21 committed stubs under `docs/source/tools/` are regenerated accordingly. This matters because **Read the Docs builds from the committed stubs** (it invokes Sphinx directly and never runs `qe_apidoc.py`).
- `split_index_template` is synced with the `index.rst` edits below — `make html` regenerates `index.rst` from this template, so without the sync the hand-edit would be silently reverted by the next local docs build. Regeneration now reproduces the committed files byte-for-byte (verified).
- Fixed `all_auto()` (the `single` mode), which called `gen_module()` — a function not defined anywhere in the file — and so died with `NameError`.

Removing the v0.8.0-overdue shims themselves is left for a separate discussion; this PR only stops the documentation from steering readers at them.

**`docs/source/conf.py` — fix stale/broken build config**

- Removed the dead `import pip` / `mock` block. It called `pip.main()` (removed in pip 10) and built a `MagicMock` that nothing uses — the `MOCK_MODULES` section below it is entirely commented out. It was a latent build hazard for no benefit.
- Fixed the `numydoc_show_class_members` typo (→ `numpydoc_`), which meant the setting was silently ignored. The corrected value matches numpydoc's default, so rendering is unchanged.
- Modernised the deprecated intersphinx mapping to the current form `{'python': ('https://docs.python.org/3', None)}`.

**`docs/source/index.rst`**

- Added `optimize` to the intro description — it was in the `toctree` but omitted from the prose that lists the modules. Stripped a trailing-whitespace line.

## Validation

- Regenerated the stubs by running `python qe_apidoc.py` and confirmed the diff is surgical: exactly the 21 shim-backed stubs change, and `index.rst`/`tools.rst`/other landing pages regenerate byte-identical to what is committed.
- All 63 `automodule` targets across `docs/source/` now resolve to real module files (script-checked); `conf.py` and `qe_apidoc.py` parse cleanly; RST heading underlines and trailing whitespace checked.
- I was not able to run a full local `make html` in this environment (numba/sphinx toolchain not available here), so the Read the Docs build on this PR is the authoritative rendering check.


mmcky:
@oyamad I have done a round on the docs to help get #836 merged. Are you happy with these fixes and changes?

oyamad:
`arma` changed to `_arma` in `.. automodule:: quantecon.arma` etc., the header in the doc now reads `class quantecon._arma.ARMA(...)` with `_`, which I think is not a right change.

(Text below generated by ChatGPT; "I" refers to ChatGPT...)


=== BEGIN ===

I agree with the goal of avoiding the deprecated shim modules, but I don’t think
switching the API docs to `quantecon._arma` / other private implementation
modules is the right user-facing representation.

In the rendered docs this produces headings such as:

    class quantecon._arma.ARMA(...)

`_arma` is an implementation detail, and the shim’s own deprecation message tells
users to import `ARMA` from the top-level `quantecon` namespace. The docs should
therefore document the public exported object, e.g. `quantecon.ARMA`, not the
private implementation module.

Could we instead generate shim-backed pages using `.. currentmodule:: quantecon`
plus explicit `.. autoclass:: ARMA` / `.. autofunction:: ...` directives derived
from the shim’s `__all__`? That would avoid documenting the deprecated shim while
also avoiding private `_module` names in the public API reference.

=== END ===

mmcky:
Thanks so much for catching this, @oyamad — you're completely right, and the write-up nailed the fix. 🙏

I've pushed c6019e3 which does exactly what you suggested: the shim-backed pages now use `.. currentmodule:: quantecon` plus explicit `autoclass` / `autofunction` directives derived from each shim's `__all__`. For example `docs/source/tools/arma.rst` now reads:

```rst
arma
====

.. currentmodule:: quantecon

.. autoclass:: ARMA
    :members:
    :undoc-members:
    :show-inheritance:
```

So the heading renders as `class quantecon.ARMA(...)` — the public name the deprecation message points people to — with no private `_arma` in the output and without importing the deprecated shim during the build. The real docstrings still render because the objects carry them. `qe_apidoc.py` decides class-vs-function for each `__all__` entry by parsing the implementation module with `ast`, so the generator stays import-free, and I regenerated all 21 shim-backed stubs. The two genuinely public base modules (`distributions`, `quad`) keep a plain `automodule`.

I checked it with a minimal Sphinx build: all 21 pages now render `quantecon.<Name>` headings with their docstrings, and the build is warning-clean.

One small tradeoff worth flagging: documenting the objects directly means the module-level docstring (e.g. *"Provides functions for working with and visualizing scalar ARMA processes"*) is no longer shown on these pages, and a couple of those docstrings also carry developer `TODO:` notes — so I'd rather not auto-inline them here. I've opened #852 to track adding curated module-level descriptions later, to keep this PR focused.

Thanks again for the careful review!

mmcky:
- [x] updated rtd configuration in read the docs settings

mmcky:
Heads-up on one open item before this merges: moving the base-level pages to document the public `quantecon.*` objects surfaced that **six objects in the shims' `__all__` are not actually re-exported at the top level** (`ckron`, `gridmake`, `annotate_nodes`, `num_compositions_jit`, `simulate_linear_model`, `solve_discrete_riccati_system`). They resolve neither as `quantecon.<name>` nor cleanly anywhere public, so the current commit **drops them from the reference** (the published `latest` docs still show them under the deprecated `quantecon.lss.*` / `quantecon.ce_util.*` shim namespace).

I've written this up in #853 for the team to decide (promote to `quantecon.*` / document via the impl module / omit) — it also fixes a real package bug where the shims' own deprecation message points users at `from quantecon import <name>` imports that raise `ImportError`. Flagging so we don't merge the coverage drop unnoticed; happy to implement whichever way #853 lands.

mmcky:
@oyamad I have fixed the RTD integrations so we should get reliable rebuilds on all PRs moving forward. This is ready for final review. 

Example: https://quanteconpy--850.org.readthedocs.build/en/850/tools/arma.html

Thanks for this suggestions. It's a huge improvement.

oyamad:
@mmcky Thanks.
What about files under submodules?
If you go to https://quanteconpy--850.org.readthedocs.build/en/850/markov/core.html#quantecon.markov.core.MarkovChain for example, it shows `class quantecon.markov.core.MarkovChain(P, state_values=None)` with `core`; ideally, it should be `class quantecon.markov.MarkovChain(P, state_values=None)`.

This is not blocking. We can leave it as is, or we may work on it in a separate PR.

mmcky:
Good catch @oyamad — you're right, and it's the same pattern one level down. `MarkovChain` is re-exported at `quantecon.markov` (in `markov/__init__.py`), but the subpackage pages are generated with `automodule:: quantecon.markov.core`, so the heading picks up the internal `core` module. The same applies across `game_theory.*`, `optimize.*`, `random.*`, and `util.*`.

The fix is exactly the treatment this PR applies to the base-level pages, just one level down (`.. currentmodule:: quantecon.markov` + explicit `autoclass`/`autofunction`). Two reasons I'd prefer to do it as a separate PR rather than expand this one, and you already hinted at the first:

1. It's a much bigger surface — ~30 file-pages and ~66 objects across five subpackages (game_theory alone is 16 files / 38 exports).
2. It runs into the same "object defined in an internal module but not re-exported at the package level" question as #853 (a handful in each subpackage), so it's best done once #853's policy is settled, and applied consistently everywhere.

I've captured it in #854 with the scope and the dependency on #853. Agreed it's non-blocking for this PR — thanks for the thorough review!

PR #855, opened 2026-07-08
BUG: Fix DiscreteDP input validation and iteration edge cases

### LINKED ISSUE #856: DiscreteDP: sorted SA-pair branch silently accepts out-of-range state indices
Spun out of the review of #855 for separate discussion. This is **not** a regression from that PR — it is a pre-existing gap the PR happens to surface.

## Summary

When a `DiscreteDP` is constructed in state-action-pair form with an out-of-range state index (`s_indices` containing some `s >= num_states`, where `num_states` is inferred from `Q.shape`), the two construction paths disagree. The *unsorted* path now raises, but the *sorted* path silently builds a corrupt model and even solves it, returning a bogus policy.

\#855 added an explicit `shape=(num_states, ...)` to the `coo_matrix` in the unsorted branch, which makes that branch reject an out-of-range `s_index`. The sorted branch (`_generate_a_indptr`) has no equivalent bounds check, so it silently produces a corrupt `a_indptr`.

## Reproduction (on the #855 branch)

```python
import numpy as np
from quantecon.markov.ddp import DiscreteDP

R = [1., 2., 3.]
Q = [(1., 0.), (0.5, 0.5), (0., 1.)]   # num_states = 2 (inferred from Q)
a_indices = [0, 0, 0]

# Sorted indices, state 2 is out of range -> silently builds a corrupt model
ddp = DiscreteDP(R, Q, 0.9, [0, 1, 2], a_indices)
print(ddp.a_indptr)                    # [0 1 3]  -- wrong length/values
print(ddp.solve(method='pi').sigma)    # solves and returns a bogus policy

# Unsorted indices, same out-of-range state -> raises (post #855)
DiscreteDP(R, Q, 0.9, [2, 1, 0], a_indices)
# ValueError: axis 0 index 2 exceeds matrix dimension 2
```

On `main` (pre-#855) *both* branches are silently wrong; #855 fixes the unsorted branch as a side effect of its shape change and thereby widens the asymmetry.

## Discussion / possible fix

Out-of-range indices are user error, so severity is low — but silent corruption plus a successful `.solve()` is the worst failure mode, and the two paths behaving differently is surprising.

One option is to validate up front, before branching on sorted/unsorted, e.g. reject `s_indices.min() < 0 or s_indices.max() >= num_states` (and analogously for `a_indices`) with a descriptive `ValueError`. That closes both paths symmetrically and gives a friendlier message than the raw scipy `axis 0 index ... exceeds matrix dimension ...`.

Opening for discussion on whether the up-front validation is worth adding.

oyamad:
Closes #856.

This PR fixes several input-validation and edge-case bugs in `DiscreteDP` (`quantecon/markov/ddp.py`), found while reviewing the module against its QuantEcon.jl counterpart.

**1. Duplicate state-action pairs silently corrupted the model**

In the state-action pair formulation with unsorted indices, duplicate (s, a) pairs were not detected: `coo_matrix(...).tocsr()` sums duplicate entries, and since the entries here are position pointers into `R` and `Q`, the constructor either raised an uninformative `IndexError` or, when the summed pointer happened to land in bounds, silently built a corrupted instance (`len(R) < num_sa_pairs`, rewards and transition probabilities reassigned across pairs, stale `s_indices`) that then solved without complaint. A `ValueError('duplicate state-action pair found')` is now raised, consistent with the behavior of QuantEcon.jl. Example that previously constructed (and solved) silently:

```python
s_indices = [1, 1, 0, 0]  # (1, 0) duplicated
a_indices = [0, 0, 0, 1]
R = [1., 2., 3., 4.]
Q = [(0.5, 0.5), (1., 0.), (0.25, 0.75), (0.6, 0.4)]
ddp = DiscreteDP(R, Q, 0.95, s_indices, a_indices)  # no error; wrong model
```

**2. Trailing states with no action caused out-of-bounds reads**

With sorted input indices, `_generate_a_indptr` walked `s_indices` past the end whenever the last state(s) had no action available — an unchecked out-of-bounds read under Numba's nopython mode. With unsorted input indices, the `coo_matrix` shape was inferred from the largest indices present, so `a_indptr` came out shorter than `num_states+1`, with the same consequence downstream. The kernel now bounds its walk and the COO shape is passed explicitly, so the intended informative `ValueError` ("for every state at least one action must be available") is raised in both cases. The two checks in `_check_action_feasibility` are also reordered (action availability first), since for a state with no action `s_wise_max` leaves the corresponding entry of `R_max` uninitialized.

**3. `operator_iteration` returned a tuple when `max_iter <= 0`, and `max_iter=0` crashed `policy_iteration`**

`operator_iteration` returned `(v, 0)` instead of the documented `num_iter` on its early-return path. The four solution methods now raise `ValueError` for `max_iter < 1` (previously `policy_iteration` died with `NameError`, and `value_iteration` stored the tuple in `res.num_iter`).

**4. Cosmetics**

`beta` is now validated at the top of `__init__` (previously only after the index-sorting work); typo fixes in docstrings ("optinal", "represetned", "correspnoding", "feasbile") and in a test name.

Tests are added for each fix; the full `quantecon/markov` test suite passes (90 passed).

### DIFF
--- a/quantecon/markov/ddp.py
+++ b/quantecon/markov/ddp.py
@@ -127,7 +127,7 @@ class DiscreteDP:
 
     There are two ways to represent the data for instantiating a
     `DiscreteDP` object. Let n, m, and L denote the numbers of states,
-    actions, and feasbile state-action pairs, respectively.
+    actions, and feasible state-action pairs, respectively.
 
     1. `DiscreteDP(R, Q, beta)`
 
@@ -297,6 +297,14 @@ class DiscreteDP:
 
     """
     def __init__(self, R, Q, beta, s_indices=None, a_indices=None):
+        if not (0 <= beta <= 1):
+            raise ValueError('beta must be in [0, 1]')
+        if beta == 1:
+            msg = 'infinite horizon solution methods are disabled with beta=1'
+            warnings.warn(msg)
+            self._error_msg_no_discounting = 'method invalid for beta=1'
+        self.beta = beta
+
         self._sa_pair = False
         self._sparse = False
 
@@ -349,9 +357,18 @@ class DiscreteDP:
                 self.a_indptr = a_indptr
             else:
                 # Sort indices and elements of R and Q
+                # (shape supplied explicitly, so that a_indptr has length
+                # num_states+1 even when the last states have no action)
                 sa_ptrs = sp.coo_matrix(
-                    (np.arange(self.num_sa_pairs), (s_indices, a_indices))
+                    (np.arange(self.num_sa_pairs),
+                     (self.s_indices, self.a_indices)),
+                    shape=(self.num_states, self.a_indices.max()+1)
                 ).tocsr()
+                if sa_ptrs.nnz < self.num_sa_pairs:
+                    # Duplicate entries have been summed on the conversion
+                    # to CSR, in which case the pointers would silently
+                    # pick up wrong elements of R and Q
+                    raise ValueError('duplicate state-action pair found')
                 sa_ptrs.sort_indices()
                 self.a_indices = sa_ptrs.indices
                 self.a_indptr = sa_ptrs.indptr
@@ -423,14 +440,6 @@ class DiscreteDP:
         # Check that for every state, at least one action is feasible
         self._check_action_feasibility()
 
-        if not (0 <= beta <= 1):
-            raise ValueError('beta must be in [0, 1]')
-        if beta == 1:
-            msg = 'infinite horizon solution methods are disabled with beta=1'
-            warnings.warn(msg)
-            self._error_msg_no_discounting = 'method invalid for beta=1'
-        self.beta = beta
-
         self.epsilon = 1e-3
         self.max_iter = 250
 
@@ -450,6 +459,20 @@ class DiscreteDP:
         some action available.
 
         """
+        if self._sa_pair:
+            # Check that for every state there is at least one action
+            # available; checked first, since for a state with no action
+            # s_wise_max leaves the corresponding entry of R_max
+            # uninitialized
+            diff = np.diff(self.a_indptr)
+            if (diff == 0).any():
+                # First state index such that no action is available
+                s = np.where(diff == 0)[0][0]
+                raise ValueError(
+                    'for every state at least one action must be available: '
+                    'violated for state {s}'.format(s=s)
+                )
+
         # Check that for every state, reward is finite for some action
         R_max = self.s_wise_max(self.R)
         if (R_max == -np.inf).any():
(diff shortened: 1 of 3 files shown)


mmcky (approved):
Thanks so much for this, @oyamad! 🙏 Really nice to see these `DiscreteDP` edge cases rounded up and fixed in one go — every claim was easy to check. **Approving.**

All four fixes target genuine bugs I reproduced on `main`, the fixes are correct, and each new test fails on base / passes here (full `quantecon/markov` suite: 90 passed).

Verified independently:

- **Duplicate `(s,a)`** — `nnz < num_sa_pairs` is sound and complete: no false positives (`tocsr()` sums duplicates but keeps explicit zeros, so the `arange`-pointer `0` is not dropped), and duplicates always route to the unsorted branch because `_has_sorted_sa_indices` rejects equal adjacent pairs — so the sorted branch legitimately needs no check of its own. 👍
- **Trailing no-action state** — the unbounded `_generate_a_indptr` walk was genuine numba UB (nopython disables bounds-checking; on base I reproduced it yielding a corrupt non-monotonic `a_indptr`). The bound plus the explicit `coo_matrix` shape fix both the sorted and unsorted paths, and `_generate_a_indptr` was the only kernel in `utilities.py` with that unbounded pattern.
- **`operator_iteration` / `max_iter`** — return type now matches the `num_iter : scalar(int)` docstring; the four guards are correctly placed after the `None`-default; neither internal caller relied on the old tuple.
- **`beta` move** — behavior-neutral; nothing between the old and new position reads `beta`.

Nice touch confirming parity with QuantEcon.jl — the Julia constructor raises `ArgumentError("Duplicate s-a pair found")` on the same input, so the behaviors line up.

A few optional, non-blocking thoughts — purely take-it-or-leave-it:

- `beta == 1` now emits its warning even when construction later fails structurally, since the `warn` moved above the structural checks. Harmless.
- The trailing-state test's *sorted* sub-case relies on numba UB failing on base (a bit non-deterministic); `pytest.raises(ValueError, match='at least one action')` would make it a firmer regression sentinel. A sparse-`Q` trailing case and a `max_iter=-1` value would round out coverage nicely.
- Might be worth a changelog line for the `operator_iteration` return-type change and the new `ValueError`s, since both are user-visible.

I also filed #856 for a related, out-of-scope gap I noticed while digging: the sorted branch still silently accepts out-of-range state indices, an asymmetry this PR's shape fix happens to surface (not a regression from your change). No action needed here — just opened it for discussion.

Thanks again  🎉

mmcky:
thanks @oyamad I am happy for you to merge this any time you're ready.

PR #859, opened 2026-07-10
MAINT: Make s_wise_max an ordinary method; DiscreteDP now picklable

oyamad:
`DiscreteDP.__init__` defined `s_wise_max` as a local closure and assigned it as an instance attribute (one variant per formulation). Local functions cannot be pickled, so `DiscreteDP` instances failed with `AttributeError: Can't get local object 'DiscreteDP.__init__.<locals>.s_wise_max'` under `pickle` — blocking `multiprocessing` and caching workflows.

This PR replaces the two closures with a single ordinary method dispatching on `self._sa_pair`. Call syntax and behavior are unchanged (`ddp.s_wise_max(vals, out=..., out_argmax=...)`), and the remaining instance state (`_lineq_solve` holding module-level functions, sparse `_I`) is picklable, so instances now round-trip. A pickle round-trip test over the three representations (product, sa-sparse, sa-dense) is added.

This is the Python half of the last item of the cross-implementation feature-sync plan (see QuantEcon/QuantEcon.jl#389 for the Julia half of the converters/metadata side); it follows up on #855/#857, on top of which it was developed.

### DIFF
--- a/quantecon/markov/ddp.py
+++ b/quantecon/markov/ddp.py
@@ -394,27 +394,6 @@ class DiscreteDP:
                         _s_indices[j] = i
                 self.s_indices = _s_indices
 
-            # Define state-wise maximization
-            def s_wise_max(vals, out=None, out_argmax=None):
-                """
-                Return the vector max_a vals(s, a), where vals is represented
-                by a 1-dimensional ndarray of shape (self.num_sa_pairs,).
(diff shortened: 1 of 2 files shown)


mmcky (approved):
LGTM — clean refactor, behavior unchanged, and the pickle round-trip test covers all three representations. Thanks @oyamad!

Issue #853, opened 2026-07-08
Six objects in the deprecated shims' __all__ are missing from the top-level quantecon namespace


mmcky:
## Summary

While refreshing the API-reference docs in #850, I found that **six objects are listed in the deprecated compatibility shims' `__all__` but were never re-exported into the top-level `quantecon` namespace.** As a result, `from quantecon import <name>` raises `ImportError` for these six today — even though each shim's own `DeprecationWarning` explicitly tells users to import them *"from the `quantecon` namespace."*

## The six objects

| Object | Shim (in its `__all__`) | Implementation module | Re-exported in `quantecon/__init__.py`? |
| --- | --- | --- | --- |
| `ckron` | `quantecon.ce_util` | `_ce_util` | No |
| `gridmake` | `quantecon.ce_util` | `_ce_util` | No |
| `annotate_nodes` | `quantecon.graph_tools` | `_graph_tools` | No |
| `num_compositions_jit` | `quantecon.gridtools` | `_gridtools` | No |
| `simulate_linear_model` | `quantecon.lss` | `_lss` | No |
| `solve_discrete_riccati_system` | `quantecon.matrix_eqn` | `_matrix_eqn` | No |

The curation looks deliberate — sibling objects in the same modules *are* promoted while these are not: `LinearStateSpace` but not `simulate_linear_model`; `solve_discrete_riccati` but not `solve_discrete_riccati_system`; `num_compositions` but not `num_compositions_jit`. `_ce_util` is not imported into `__init__.py` at all. There is even a `#->Propose Delete From Top Level` block in `__init__.py`, so the top-level surface has clearly been managed by hand.

## Why it matters

**1. The deprecation guidance is currently broken for these six.** Accessing one through its shim, e.g. `quantecon.lss.simulate_linear_model`, emits:

> Please use `simulate_linear_model` from the `quantecon` namespace, the `quantecon.lss` namespace is deprecated. You can use the following instead: `from quantecon import simulate_linear_model`.

…but `from quantecon import simulate_linear_model` then raises `ImportError`. A user who follows the warning hits a dead end. This is independent of the docs — it's a package-level inconsistency.

**2. It surfaced as a docs question in #850.** That PR moves the base-level API pages from documenting the deprecated shim / private module to documenting the public objects in the `quantecon` namespace. Because these six aren't in that namespace, they can't be documented as `quantecon.<name>`. For reference, how each version represents them:

| Version | Rendered heading for these six |
[...]

Issue #707, opened 2023-05-25
FIX: DOC: Fix build error with Sphinx v7.0.1


oyamad:
`jinja2.exceptions.UndefinedError: 'style' is undefined` (https://readthedocs.org/projects/quanteconpy/builds/20811367/)

Maybe readthedocs/sphinx_rtd_theme#1463?

For the moment, pin the version of `Sphinx` as < 7.0?

oyamad:
Temporary workaround by #708.
Left open as a medium run issue.

PR #864, opened 2026-07-23
DOC: repo-wide documentation consistency audit

mmcky:
## Summary

A systematic audit of the documentation surface — Sphinx sources, the apidoc generator, every docstring in the package, the repo-level prose docs, build config, and every hyperlink — together with the fixes it turned up. One open question (#853) is deliberately left untouched.

Findings were verified against the code rather than taken at face value, and the resulting diff was reviewed again for regressions before this PR.

## Highest-impact fixes

**The published docs describe the wrong package.** `docs/source/conf.py` put a *cwd-relative* path on `sys.path`. Sphinx runs from `docs/`, so it resolved to a directory that does not exist, and autodoc silently fell through to whatever `quantecon` happened to be installed in site-packages instead of the checkout. Read the Docs has therefore been publishing API docs generated from the installed wheel. The entry is now `__file__`-relative. Locally this changed the reported version in the built output from the installed `0.12.0` to the checkout's version.

**`quantecon.timings` was documented nowhere.** It is exported at top level and declares its own `__all__`, but `grep -r timings docs/` returned zero hits. The generator now emits a Timings section, and the generated pages are committed.

**A documented rebuild path was broken.** `qe_apidoc.py` never created `source/game_theory/game_generators/`, so the `make srcclean && python qe_apidoc.py` sequence in `docs/README.md` crashed with `FileNotFoundError`. `srcclean` also silently skipped `source/optimize`, so a "clean" rebuild left stale pages.

**Docstrings that contradict the code.** `robust_rule`, `robust_rule_simple` and `evaluate_F` in `_robustlq.py` each documented a return order that does not match their `return` statement. `nnash` in `_lqnash.py` had the `S1`/`S2` shapes swapped — the tests do not catch it because they use `k_1 == k_2`. Around 90 further parameter, return, default and type corrections follow, plus numpydoc structural fixes (a bare `.. deprecated::` and a missing blank line before a list were both docutils *errors*, so those admonitions never rendered).

**31 broken examples.** A `--doctest-modules` sweep found examples showing output no supported NumPy produces (pre-1.14 array spacing), NumPy 2 scalar reprs, missing imports, and two outright `NameError`s — including `lorenz()`, which does not exist. All 31 were fixed and each verified by running it. Examples that take a `random_state` now use the `rng = np.random.default_rng(1234)` pattern throughout, with every expected output regenerated by running the example.

**Prose corrected against ground truth.** The README source-install block (ran `flit install` with no `cd`, so it could not work), the BibTeX entry (carried the Zenodo DOI on a JOSS `@article`), and around ten dead or redirecting URLs — two of which, `lectures.quantecon.org` and `discourse.quantecon.org`, are NXDOMAIN.

## Please look closely at these two

Both are behaviour changes rather than documentation, and both are reasonable to push back on:

| Change | Why | Risk |
| --- | --- | --- |
| `requires-python`: `>=3.7` → `>=3.12` | Matches the classifiers, the CI matrix and RTD. Stale since the 2022 flit migration. | Narrows installability — users below 3.12 stop resolving new releases. Worth a line in the release notes. |
| Removed `quantecon/util/common_messages.py` | Referenced nowhere (`grep` hits only its own definition), and it describes a graceful Numba fallback that no longer exists, since numba is a hard dependency and `__init__.py` hard-raises. | It was importable, so this is an API removal. |

## Also included

Adds a docs job to CI. Nothing built the docs, which is precisely how the `conf.py` and `srcclean` breakage survived unnoticed. The job runs `qe_apidoc.py` and then `git diff --exit-code -- docs/source`, because RTD builds the *committed* `.rst` files and never runs the generator — regenerating without the drift check would pass green while RTD served stale pages. It builds without `-W` for now, since six warnings remain pending #853.

Unpins `sphinx<=6.2.1`, added in #708 as a temporary workaround for #707, now closed — the pin held the docs on a 2023 Sphinx whose metadata tops out at Python 3.11 while RTD builds on 3.14. Deletes the dead `docs/sphinxext/` (three vendored modules, all commented out of `conf.py`) and `MANIFEST.in` (names a nonexistent `LICENSE.txt`, and flit never reads it). The 21 deprecation shims said "will be removed v0.8.0"; v0.8.0 shipped in February 2025, so the wording is now version-agnostic — an actual removal schedule is a separate discussion.

The Sphinx unpin and the `sphinxext` deletion are the first two items of #885 from the technical-debt audit. Its third item, moving the duplicated runtime floors out of `rtd-requirements.txt`, is not part of this PR, so #885 stays open for that.

## Since the last review round (10 September 2026)

Merged `main`. Three files conflicted. `.github/copilot-instructions.md` was deleted upstream in #941 and `AGENTS.md` now carries that guidance (it makes none of the stale CI-matrix or test-count claims this branch had corrected), so the deletion stands. `quantecon/random/utilities.py` keeps the `draw` docstring from #917, which supersedes the seeded example discussed in the review thread — `draw` now takes an `rng` and its example uses the same generator pattern as the rest of the module. `normal_form_game.py` keeps the `tol` wording from #936.

The `discrete_var` example shows the simulated path again rather than its `.shape`. The output is deterministic under the seeded generator, and the same block already prints exact `state_values` from the same stream, so it introduces no new fragility.

The docs job now uses `actions/setup-python@v7`, matching the rest of `ci.yml` after Dependabot's bump on `main`.

The `ci_np2.yml` trigger fix that earlier versions of this description mentioned landed separately in #911 and is no longer part of this diff.

The index page's description of `util` now names only `Timer` and `timeit` as the user-facing timing tools, so it does not advertise `tic`/`tac`/`toc` ahead of their deprecation in #833. That PR merges cleanly onto this branch in either order.

## For the release notes

Release notes are generated from PR titles at tag time, and this PR's title would hide the removals below. Whoever cuts the next release should carry these into the notes by hand:

- `requires-python` is now `>=3.12`. Users on older interpreters stop resolving new releases.
- `quantecon.util.common_messages` is removed. It held a single string for a Numba-fallback path that no longer exists, and nothing imported it.
- `docs/sphinxext/` is removed. It held vendored, Python 2 era copies of the IPython Sphinx directives and numpydoc's `only_directives`; the docs build has used the extensions shipped with IPython and matplotlib for years, and nothing referenced the directory beyond two `sys.path` entries in `conf.py`.
- `MANIFEST.in` is removed. It named a nonexistent `LICENSE.txt`, and flit never read it.
- The docs build is no longer pinned to `sphinx<=6.2.1`.

## Deliberately not addressed

The six autodoc warnings that remain are the open question in #853 and are untouched here. See #853 for new findings on it, including why emptying `ce_util.__all__` is not a clean option.

Six `game_theory` docstrings call `np.set_printoptions(precision=4)` and never restore it. Harmless as rendered documentation, but under a package-wide doctest run it leaks into every module collected later. Left alone; tracked separately.

## Validation

- `pytest quantecon` on the merged branch — 678 passed, 2 pre-existing warnings
- `flake8 --select=F401,F405,E231 quantecon` — clean
- `sphinx-build` — succeeds, 6 warnings, all of them #853
- Doctests per module in isolation — every module touched by this PR passes, except `util/timing.py`, whose elapsed-time values are illustrative by design, and `util/notebooks.py`, which needs the network. Neither is masked with `# doctest: +SKIP`, which would render visibly in the published docs.
- `qe_apidoc.py` is idempotent and regenerates the committed `.rst` byte-for-byte on the merged branch, so the CI drift gate passes.
- An AST comparison with docstrings stripped confirms no executable code changed anywhere in `quantecon/`, apart from the shim deprecation strings and the removed module.


mmcky:
@oyamad I am working through my review of this so I have left this in draft for now. But it is looking like a good update.

quantecon/random/utilities.py
```diff
@@ -191,11 +195,14 @@ def draw(cdf, size=None):
 
     Examples
     --------
+    >>> import numpy as np
+    >>> import quantecon as qe
     >>> cdf = np.cumsum([0.4, 0.6])
-    >>> qe.random.draw(cdf)
-    1
+    >>> np.random.seed(1234)
```
oyamad, line 214: This is not a good convention.
We should extend `draw` to be able to accept a random number generator, which is now possible with recent versions of Numba.

oyamad (commented):
I found one issue to comment on while I was just browsing.

quantecon/random/utilities.py
mmcky: thanks @oyamad -- nice pickup and a good improvement. We are currently upgrading out lectures to use new `rng` patterns.

mmcky: Agreed — global seeding shouldn't be the convention in the docs. Fixed in 63d784b, and I took it a bit wider than this one line.

The examples were in three different styles: `np.random.default_rng(...)` passed as `random_state` (`game_theory.random`, `markov.approximation`), a bare int `random_state=1234` (nine sites across `bimatrix_generators`, `markov.random`, `random.utilities`), and the `np.random.seed(1234)` you flagged. They are now all on the `rng = np.random.default_rng(1234)` + `random_state=rng` pattern. Since `default_rng` seeds PCG64 through `SeedSequence` rather than the legacy Mersenne Twister, the streams are unrelated and every expected output was regenerated by running the example rather than by editing the old numbers.

`draw` is the exception, because it has no `random_state` argument and calls `np.random.random()` directly — in the pure-Python body and in the `@overload` implementation. So there is no generator to pass, and its example now checks properties of the sample rather than exact values, with a Notes section recording why:

```
>>> cdf = np.cumsum([0.4, 0.6])
>>> qe.random.draw(cdf) in (0, 1)
True
>>> qe.random.draw(cdf, 10).shape
(10,)
```

That is a placeholder for your actual suggestion rather than a substitute for it. Extending `draw` to accept a generator is a behaviour change with a jitted code path to get right, so I'd rather it not ride along in a documentation PR — happy to open a separate issue for it if you agree that's the right split.

Verified: 600 tests pass, doctests pass for all five modules that use a generator in their examples, flake8 output unchanged from baseline.

mmcky: @oyamad I need to look into the draw interface a bit more closely here.

mmcky:
@oyamad [5e26642](https://github.com/QuantEcon/QuantEcon.py/pull/864/commits/5e26642d15ce3d86b8241f8f5efe1b871b6d9d71) implements your suggestion. The actual values change due to the change in underlying algorithm being used (despite the seed changing). Is this what you had in mind?

quantecon/util/timing.py
```diff
@@ -313,14 +318,14 @@ def timeit(func, runs=1, stats_only=False, verbose=True, results=False, **timer_
     Basic usage:
     >>> def slow_function():
     ...     time.sleep(0.01)
-    >>> timeit(slow_function, runs=3)
+    >>> timeit(slow_function, runs=3, precision=2)
     Run 1: 0.01 seconds
```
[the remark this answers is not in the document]
mmcky: Partly taken, with a correction to the premise.

The `--doctest-modules` framing doesn't apply here: this repo doesn't collect doctests. `pytest.ini` contains only a `markers` section, CI runs plain `coverage run -m pytest quantecon`, and `doctest` appears nowhere in `.github/`, `CONTRIBUTING.md`, or `.readthedocs.yaml`. These examples also weren't executable to begin with — they call `time.sleep` with no `import time`, so they'd raise `NameError` before any timing mismatch mattered. So nothing was flaky or failing.

The underlying observation is still fair, though, and it caught a real regression this PR introduced: the newly added `Average time: 0.0120 seconds` at line 340 was a fabricated value — the example actually produces ~0.0103 — which is exactly the kind of thing a documentation-accuracy PR shouldn't be adding. Line 350 was a genuine inconsistency too: `timeit(add_func, runs=2)` prints, while every other example in that docstring shows its output.

Fixed in 8e72f7a, but not by switching to `verbose=False` plus assertions on the returned dict. `timeit`'s formatted printed output is the feature being documented, so an Examples section that never shows it would be worse documentation. Instead the examples stay illustrative and are now honest about it:

- an up-front note that the timing values are illustrative and vary with machine and load
- `Average time: 0.0120 seconds` replaced with a value the example actually produces
- the missing output added to the final `timeit(add_func, runs=2)` example
- `>>> import time` added so the snippets are copy-pasteable

Also applied the same treatment to the `Timer` class docstring above it, which had identical hard-coded timings (`0.0000 seconds elapsed`, `Method took 0.000123 seconds`) that this review didn't flag.

quantecon/random/utilities.py
```diff
@@ -189,13 +198,23 @@ def draw(cdf, size=None):
...
+    >>> import quantecon as qe
     >>> cdf = np.cumsum([0.4, 0.6])
-    >>> qe.random.draw(cdf)
-    1
-    >>> qe.random.draw(cdf, 10)
-    array([1, 0, 1, 0, 1, 0, 0, 0, 1, 0])
+    >>> qe.random.draw(cdf) in (0, 1)
+    True
+    >>> qe.random.draw(cdf, 10).shape
+    (10,)
```
oyamad: I am not sure if this is a good demonstration for the user. I guess passing the doctest is of secondary importance?

quantecon/random/utilities.py
```diff
@@ -189,13 +198,23 @@ def draw(cdf, size=None):
     -------
     scalar(int) or ndarray(int, ndim=1)
 
+    Notes
+    -----
+    Unlike the other functions in this module, `draw` draws from NumPy's
+    global random state and takes no `random_state` argument, so its
```
oyamad: This is not true if this function is used in a Numba jitted function. Try this one:

```python
@njit
def f(cdf):
    return qe.random.draw(cdf, 10)

cdf = np.cumsum([0.4, 0.6])
np.random.seed(1234)
f(cdf)
```

mmcky: Taken — the Notes block was wrong, though I get a different result from your snippet than I think you expected.

Run verbatim on numba 0.62.1 / numpy 2.3.5, it isn't reproducible. Three fresh processes gave `[1 1 0 0 1 1 1 1 1 0]`, `[1 0 1 0 0 0 1 1 1 1]`, `[0 1 1 0 1 0 1 0 0 1]`. `np.random.seed` called from Python seeds NumPy's global state, which Numba's RNG doesn't read — the seed call has to happen *inside* jitted code to take effect.

But that only sharpens your point, because it means the Notes block was wrong in both halves, not one. It claimed `draw` "draws from NumPy's global random state" — under jit it draws from Numba's own internal state, which is a different generator entirely — and it claimed the output can't be made reproducible, which is false on both paths:

| Path | Seeding mechanism | Result for `size=10` |
| --- | --- | --- |
| Pure Python | `np.random.seed(1234)` | `[0 1 1 1 1 0 0 1 1 1]` |
| Jitted | `np.random.seed(1234)` **inside** the jitted function | `[0 1 1 1 1 0 0 1 1 1]` |

Both stable across processes, and identical to each other — Numba's internal generator is a Mersenne Twister, stream-compatible with NumPy's legacy global state.

Rewritten in 08682a7 to say exactly that, including the inside-vs-outside distinction that tripped up the snippet.

Separately, I confirmed your original suggestion is implementable today: numba 0.62.1 accepts a `np.random.Generator` in nopython mode and reproduces the host stream exactly. That's still a signature and behaviour change with a jitted path to get right, so I'd rather it not ride along in a docs PR — opened as #916, which the new Notes block should be replaced by once it lands.

quantecon/random/utilities.py
mmcky: Agreed, and you've named the right cause — the example was shaped around the doctest rather than around the reader, and that was the wrong priority.

It only looked necessary because of the Notes block you flagged just below. Having established there that `draw` *can* be seeded, the property checks have nothing left to work around, so they're gone and the example shows the values the function actually returns:

```python
>>> cdf = np.cumsum([0.4, 0.6])
>>> np.random.seed(1234)
>>> qe.random.draw(cdf, 10)
array([0, 1, 1, 1, 1, 0, 0, 1, 1, 1])
>>> int(qe.random.draw(cdf))
0
```

On `np.random.seed` reappearing here after your earlier comment: I'd read that comment more widely than it was meant. It objected to global seeding where a generator *could* be passed instead, and I applied it to `draw` too. But `draw` takes no `random_state`, so seeding is the only reproducibility mechanism it currently has — it isn't standing in for a generator that was available. That changes with #916.

The `int()` wrapper is there because the bare NumPy 2 repr is `np.int64(0)`. The pre-PR example showed `1`, a NumPy 1 repr — the same class of staleness this PR is fixing elsewhere.

Fixed in 08682a7. Verified: 600 tests pass, the docstring passes as a doctest in repeat fresh processes, flake8 clean.

quantecon/random/utilities.py
```diff
@@ -200,21 +200,22 @@ def draw(cdf, size=None):
...
     Examples
     --------
     >>> import numpy as np
     >>> import quantecon as qe
     >>> cdf = np.cumsum([0.4, 0.6])
-    >>> qe.random.draw(cdf) in (0, 1)
-    True
-    >>> qe.random.draw(cdf, 10).shape
-    (10,)
+    >>> np.random.seed(1234)
```
oyamad, line 214: In my opinion, this is not a good convention. (We should not introduce doctest until #916 is implemented.)

quantecon/random/utilities.py
```diff
@@ -200,21 +200,22 @@ def draw(cdf, size=None):
...
+    >>> qe.random.draw(cdf, 10)
+    array([0, 1, 1, 1, 1, 0, 0, 1, 1, 1])
+    >>> int(qe.random.draw(cdf))
```
oyamad: Do we encourage the users to use `int()`? (Is `np.int64(0)` something we want to avoid?)

mmcky: I agree. I think an object repr is OK here.

quantecon/random/utilities.py
mmcky: Fair, and agreed it shouldn't become the convention — documenting the plan here so the thread has it.

The real fix is #917 (implementing #916): `draw` gains support for a passed-in `np.random.Generator`, and this example gets rewritten there to the same `rng = np.random.default_rng(1234)` pattern the rest of the module moved to in this PR — at which point the `np.random.seed` line and the Notes block below it both disappear. The seed line is a stopgap: with the current signature it's the only way the example can show real, reproducible output rather than property checks.

On the parenthetical — agreed, and it's how the PRs are sequenced: #920 (the doctest CI job) is stacked behind this PR and queued behind #917, so this example will never be enforced as a doctest in its seeded form. Intended landing order: this PR → #917 → #920.

quantecon/random/utilities.py
mmcky: Good suggestion — adopted as the PR-wide convention in a99927a. This was one of four sites wrapping scalar outputs, so all four now show the object repr: `np.int64(2)` (`cartesian_nearest_index`), `np.uint64(7)` (`_ints_arr_to_bits`), `np.float64(0.0)` (`lcp_lemke`), and `np.int64(0)` here — each output verified by running the example. It's honest about what the functions actually return, and with CI and the docs build on NumPy 2 there's no portability cost. #917 will carry the same convention into the rewritten `draw` example.

docs/source/util.rst
```diff
@@ -6,7 +6,6 @@ Utilities
 
    util/array
    util/combinatorics
-   util/common_messages
    util/compat
    util/notebooks
```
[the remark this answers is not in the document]
mmcky: Deliberate, and flagged in the PR description as one of the two behaviour changes for reviewers to look at closely. A stub isn't the right tool here: unlike the 21 deprecation shims in this PR's scope (which re-export working code from the renamed private modules), `common_messages` contained only `numba_import_fail_message` — an error string for the graceful-Numba-fallback path that no longer exists, since numba became a hard dependency and `quantecon/__init__.py` hard-raises without it. A stub would preserve an import whose only content is a stale message about a code path the package no longer has, and `grep` finds no consumers in-repo or in the docs. The removal will be called out in the next release notes (the repo's CHANGELOG defers to GitHub release notes, so that happens at tag time).

mmcky:
@oyamad I think this is in relatively good shape now. Thanks for your comments -- they were really helpful. Any other pickups? Otherwise I think this is a big improvement.

quantecon/random/utilities.py
oyamad: I still wonder why we want to include this "antipattern" even if it is temporary. Right now `draw` does not offer an option to generate reproducible outputs (in my opinion `np.random.seed(1234)` is not a choice), so the example should be as such, I suppose?

mmcky: Done in b1a6342 — merging `main` brought in the `draw` docstring from #917, so the seeded example is gone; the example now builds `rng = np.random.default_rng(1234)` and passes it as `rng=`, the same pattern as the rest of the module.

mmcky:
Author review complete. Every docstring correction was re-verified against the code, each changed module's examples run as doctests in isolation, and the apidoc drift gate, Sphinx build and full test suite are green on the merged branch. The removals are listed under "For the release notes" in the description. Ready to merge from my side; the two behaviour changes flagged above are still open for a second maintainer's view.

mmcky:
@oyamad I have done a detailed review this morning. I think this is now in a good state to merge so I will do so.

PR #904, opened 2026-07-25
FIX: fetch_nb_dependencies must not save HTTP error pages

### LINKED ISSUE #870: FIX: fetch_nb_dependencies writes HTTP error pages to disk and reports success
### Problem

`fetch_nb_dependencies` (`quantecon/util/notebooks.py`, exported at the top level) writes whatever the server returns straight to disk and reports success. When the requested remote file does not exist, GitHub's HTML error page is saved under the requested filename and the function appends `True` to the returned status list. Reproduced live during the audit:

```bash
cd "$(mktemp -d)" && python -c "from quantecon import fetch_nb_dependencies; import os; print(fetch_nb_dependencies(['does-not-exist-12345.csv'], verbose=False), os.path.getsize('does-not-exist-12345.csv'))"
```

Output: `[True] 314490` — a 314 KB HTML document saved as `does-not-exist-12345.csv`, reported as fetched. The failure surfaces later as a parse error in unrelated code.

Four distinct defects in ~30 lines:

1. No `r.raise_for_status()` — the root cause above.
2. No `timeout=` on `requests.get`, so a stalled connection blocks indefinitely.
3. `with open(fl, "wb") as fl:` rebinds the loop variable to the file handle — works today only because `fl` is not read again in that iteration.
4. `if type(files) == list:` (flake8 E721) — a tuple, the most natural alternative input, falls through to the `dict` branch and raises an opaque `AttributeError`.

The module docstring also carries an unaddressed 2016-era TODO list ("replace wget usage" — wget is no longer used), and the default target repo (`QuantEcon/QuantEcon.notebooks`) was last pushed 2023-07-15.

### Acceptance criteria

- [ ] `r.raise_for_status()` before writing, or an explicit status check that appends `False` and warns rather than writing
- [ ] A default `timeout` on the request, overridable by keyword argument
- [ ] The file handle no longer shadows the loop variable
- [ ] `isinstance(files, (list, tuple))` replaces the `type(...) ==` check
- [ ] A regression test asserting that a missing remote path returns `False` and leaves no file on disk
- [ ] The stale TODO block in the module docstring is removed or updated

Note: the 2017 remote branch `update-fetch-utility` may contain prior work here — worth a look before writing.

---

Sanjays2402:
Closes #870

`fetch_nb_dependencies` wrote whatever the server returned straight to disk and appended `True` to the status list, so requesting a file that does not exist remotely saved GitHub's HTML error page under the requested filename and reported it as fetched — surfacing later as a parse error in unrelated code.

This calls `raise_for_status()` and appends `False` on any `RequestException` instead of writing the body, adds a default request `timeout` (overridable by keyword), stops rebinding the loop variable to the open file handle, and accepts a tuple via `isinstance` rather than `type(files) == list`.

Three regression tests in `TestFetchNbDependenciesErrors` cover the three behaviours; all three fail against the original source (error page written, `timeout` rejected as an unexpected kwarg, tuple raising `AttributeError: 'tuple' object has no attribute 'keys'`) and pass with the fix. The two existing network tests are untouched.

### DIFF
--- a/quantecon/util/notebooks.py
+++ b/quantecon/util/notebooks.py
@@ -27,9 +27,12 @@ RAW = "raw"
 BRANCH = "master"
 #Hard Coded Dependencies Folder on QuantEcon.notebooks
 FOLDER = "dependencies"
+#-Default timeout (seconds) for each request-#
+TIMEOUT = 30
 
 
-def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDER, overwrite=False, verbose=True):
+def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDER, overwrite=False, verbose=True,
+                          timeout=TIMEOUT):
     """
     Retrieve raw files from QuantEcon.notebooks or other Github repo
     
@@ -45,6 +48,8 @@ def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDE
     folder      str, optional(default=FOLDER)
     overwrite   bool, optional(default=False)
     verbose     bool, optional(default=True)
+    timeout     int or float, optional(default=TIMEOUT)
+                Timeout in seconds passed to each request
 
     Examples
     --------
@@ -74,7 +79,7 @@ def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDE
     import requests
 
     #-Generate Common Data Structure-#
-    if type(files) == list:
+    if isinstance(files, (list, tuple)):
         files = {"" : files}
 
     status = []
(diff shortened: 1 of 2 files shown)


mmcky:
Thanks @Sanjays2402, and apologies for the outcome — closing this one alongside #905, which fixed the same bug and has also been closed. Neither is being closed on quality grounds.

We have decided to retire the module rather than repair it. `quantecon/util/notebooks.py` will be deprecated in 0.12.0 and removed in v1.0. `requests` is a hard dependency of QuantEcon.py solely to serve `fetch_nb_dependencies`, and the function has no remaining constituency — no usage in any of the lecture repositories, and the few callers we can find are dormant notebooks descending from two 2017-era demos. Its default remote was last updated in 2023. Any data still needed will move to QuantEcon/data-lectures, the canonical home for datasets referenced by stable URLs.

Worth saying plainly: you and the author of #905 filed within twenty-one minutes of each other, so neither of you could have seen the other's work. That collision was nothing either of you could have avoided — #870 sitting open as a high-priority bug is what drew you both to the same thirty lines on the same morning. We are closing #870 as well, so that nobody else repeats the work.

Thanks for the time you put in, and please do pick up another issue if you would like.

PR #905, opened 2026-07-25
FIX: do not write failed downloads to disk in fetch_nb_dependencies

### LINKED ISSUE #870: FIX: fetch_nb_dependencies writes HTTP error pages to disk and reports success

earfman:
Fixes #870

Addresses all six acceptance criteria from the issue.

## Behaviour change

A failed request no longer produces a file. Previously the response body was written unconditionally and `True` appended, so a missing remote path saved GitHub's HTML error page under the requested filename and reported it as fetched — the failure surfaced later as a parse error somewhere unrelated.

Reproduced against a local server returning 404 (no network dependency):

```
before:  status [True]   absent.csv written (404 HTML body on disk)
after:   status [False]  no file, UserWarning emitted
```

While verifying I found one more consequence worth flagging: with `overwrite=True`, the old code replaced a good local file with the error page and still reported success. A file containing real data became `<html>...404...</html>`. That is silent data loss, and it is also fixed here.

## Changes

- `raise_for_status()` inside a `try`; on `requests.RequestException` the code warns, appends `False`, and skips the write
- module-level `TIMEOUT = (10, 30)` (connect, read), overridable via a new `timeout=` keyword argument
- the file handle no longer shadows the loop variable (`as f` rather than `as fl`)
- `isinstance(files, (list, tuple))` replaces `type(files) == list`, so a tuple no longer falls through to the dict branch and raises `AttributeError`
- the stale 2016 TODO block is removed from the module docstring (wget is long gone), and the return value is now documented

I looked at the `update-fetch-utility` branch mentioned in the issue — it is docstring reformatting and import hoisting only, none of the four defects are addressed there, so nothing to salvage.

## Tests

New `TestFetchFailures` class, offline (a local `http.server` on an ephemeral port, `tmp_path` + `monkeypatch.chdir`, no network, no stray files): missing remote file returns `False` and writes nothing, a 200 still writes correct bytes, tuple input is accepted, and the timeout is forwarded — including an assertion that the default is a real bound rather than `None`.

The two existing tests are unchanged and still pass. Worth knowing: they fetch a live URL, so in a sandbox without access to github.com they were previously passing on a saved error page (the bug passing its own test). They pass normally against the real URL, which is what CI sees.

## Notes for review

- Callers who relied on `requests` exceptions propagating now get `False` in the status list instead. The new Returns section documents this; happy to re-raise instead if you would rather keep the exception contract.
- `raise_for_status()` only inspects the HTTP status. A host that returns 200 with an HTML page for a wrong path would still be written verbatim; a content-type check would be a separate change.
- The `(10, 30)` default is a per-read-gap timeout, not a total budget: a slow-but-steady large download still succeeds. I checked that a 120 KB transfer trickling over ~58s completes fine.

### DIFF
--- a/quantecon/util/notebooks.py
+++ b/quantecon/util/notebooks.py
@@ -11,15 +11,10 @@ when downloaded as a support File
 
 "https://github.com/QuantEcon/QuantEcon.notebooks/raw/master/dependencies/mpi/something.py" --> ./something.py
 
-TODO
-----
-1. Write Style guide for QuantEcon.notebook contributions
-2. Write an interface for Dat Server
-3. Platform Agnostic (replace wget usage)
-
 """
 
 import os
+import warnings
 
 #-Remote Structure-#
 REPO = "https://github.com/QuantEcon/QuantEcon.notebooks"
@@ -27,16 +22,19 @@ RAW = "raw"
 BRANCH = "master"
 #Hard Coded Dependencies Folder on QuantEcon.notebooks
 FOLDER = "dependencies"
+#-Default timeout (seconds) for the remote request: (connect, read)-#
+TIMEOUT = (10, 30)
 
 
-def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDER, overwrite=False, verbose=True):
+def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDER, overwrite=False,
+                          verbose=True, timeout=TIMEOUT):
     """
     Retrieve raw files from QuantEcon.notebooks or other Github repo
     
     Parameters
     ----------
-    file_list   list or dict
-                A list of files to specify a collection of filenames	
+    file_list   list, tuple or dict
+                A list or tuple of files to specify a collection of filenames
                 A dict of dir : list(files) to specify a directory
     repo        str, optional(default=REPO)
     raw 		str, optional(default=RAW)
@@ -45,6 +43,17 @@ def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDE
     folder      str, optional(default=FOLDER)
     overwrite   bool, optional(default=False)
     verbose     bool, optional(default=True)
+    timeout     float or tuple, optional(default=TIMEOUT)
+                Passed through to ``requests.get``. Either a single value or a
+                ``(connect, read)`` pair, in seconds.
+
+    Returns
+    -------
+    status      list(bool)
+                One entry per requested file: ``True`` if the file was fetched
+                and written, ``False`` if it was skipped or could not be
+                retrieved. A failed request emits a warning and leaves no file
+                on disk.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


mmcky:
Thanks @earfman, and sorry to be closing this one — it is a thorough piece of work and the diagnosis went well beyond what the issue asked for. Two findings in particular were yours rather than ours: that `overwrite=True` would replace a good local file with the error page and still report success, and that the two existing tests had been passing against a saved error page, so the bug was quietly passing its own test. Both are worth recording, and they are the reason this closure is not a judgement on the code.

We have decided to retire the module instead of repairing it. `quantecon/util/notebooks.py` will be deprecated in 0.12.0 and removed in v1.0. `requests` is a hard dependency of QuantEcon.py solely to serve `fetch_nb_dependencies`, and the function has no remaining constituency — no usage across any of the lecture repositories, and the handful of callers we can find are dormant notebooks descending from two 2017-era demos. Its default remote was last updated in 2023. Any data still needed will move to QuantEcon/data-lectures, which is the canonical home for datasets referenced by stable URLs.

Given that, merging a fix to a function scheduled for removal would spend review on code we are deleting, so we are closing this and #904 together, and closing #870 as well so that nobody else picks it up and repeats the work. That last part is on us — #870 sat open as a high-priority bug and drew two independent fixes on the same morning.

Thank you for the care you put into this, and please do consider another issue on the tracker if you would like — this one simply stopped being worth fixing.

earfman:
Thanks @mmcky, that makes sense. Retiring it beats patching something nobody calls. For transparency, this one was AI-assisted (Assisted-by: claude  opus5), and I'll put that line on anything I send from here on. I'll look through the tracker for something else.

PR #908, opened 2026-07-25
FIX: assert the sparse-Q equality check in test_ddp.py

### LINKED ISSUE #874: FIX: Sparse-Q equality check in test_ddp.py is missing its assert
### Problem

`quantecon/markov/tests/test_ddp.py:469`:

```python
assert_allclose(ddp_s.R, sparse_R)
# allcose doesn't work on sparse
np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
assert_allclose(ddp_s.beta, beta)
```

The middle line computes a boolean and discards it — the sparse transition-array equality check has never verified anything. The comment shows the intent; the `assert` went missing. (ruff flags this as B015 "pointless comparison"; the current 3-code flake8 gate cannot see it.)

### Proposed change

```python
assert np.max(np.abs(sparse_Q.toarray() - ddp_s.Q.toarray())) < 1e-15
```

or, cleaner, `assert_allclose` on densified arrays with an explicit `atol`. Confirm the assertion actually passes — if it does not, this issue just found a real DDP bug; report findings either way. Optionally fix the `# allcose` typo while there.

### Acceptance criteria

- [ ] The comparison's result is asserted; suite green
- [ ] A deliberate perturbation of `ddp_s.Q` makes the test fail (verified locally, not committed)

---

Sanjays2402:
Closes #874

The sparse-Q comparison in `test_ddp_to_sa_and_to_product` computed a boolean and threw it away, so that check has never verified anything — a wrong converted `Q` would pass silently. This adds the missing `assert` (and fixes the adjacent `allcose` typo).

The assertion passes against the current conversion code; substituting a deliberately wrong `Q` into the same expression raises `AssertionError`, confirming it now guards. `pytest quantecon/markov/tests/test_ddp.py` is green (20 passed).

### DIFF
--- a/quantecon/markov/tests/test_ddp.py
+++ b/quantecon/markov/tests/test_ddp.py
@@ -465,8 +465,8 @@ def test_ddp_to_sa_and_to_product():
     # make sure conversion worked
     for ddp_s in [ddp_sa, ddp_sa2, ddp_sa3]:
         assert_allclose(ddp_s.R, sparse_R)
-        # allcose doesn't work on sparse
-        np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
+        # allclose doesn't work on sparse
+        assert np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
         assert_allclose(ddp_s.beta, beta)
 
     # these two will have probability 0 in state 2, action 0 b/c
--- a/quantecon/markov/tests/test_ddp.py
+++ b/quantecon/markov/tests/test_ddp.py
@@ -466,7 +466,7 @@ def test_ddp_to_sa_and_to_product():
     for ddp_s in [ddp_sa, ddp_sa2, ddp_sa3]:
         assert_allclose(ddp_s.R, sparse_R)
         # allclose doesn't work on sparse
-        assert np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
+        assert_(np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15)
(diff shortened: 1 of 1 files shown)


quantecon/markov/tests/test_ddp.py
```diff
@@ -465,8 +465,8 @@ def test_ddp_to_sa_and_to_product():
     # make sure conversion worked
     for ddp_s in [ddp_sa, ddp_sa2, ddp_sa3]:
         assert_allclose(ddp_s.R, sparse_R)
-        # allcose doesn't work on sparse
-        np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
+        # allclose doesn't work on sparse
+        assert np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15
```
oyamad:
```suggestion
        assert_(np.max(np.abs((sparse_Q - ddp_s.Q))) < 1e-15)
```
to be consistent with the other part?

oyamad (changes requested):

Sanjays2402:
Applied, thanks - switched to `assert_(...)` so it matches the rest of the file (the `assert_` usages at lines 62-63, 95-96, 348-350, 366-367). `assert_` is already imported from `numpy.testing`, so no import change was needed. `pytest quantecon/markov/tests/test_ddp.py` passes: 20 passed.

oyamad (approved):
Thanks!

mmcky:
thanks @Sanjays2402 and @oyamad

PR #910, opened 2026-07-26
MAINT: Rephrase pejorative wording in test_lqnash comment

### LINKED ISSUE #901: MAINT: Replace pejorative wording in a test comment
### Problem

`quantecon/tests/test_lqnash.py:95` contains a comment using "ghetto" as a pejorative. The word carries a racial connotation and has no place in a widely-forked teaching library. The underlying concern — that hardcoded MATLAB reference values are an unsatisfying way to fix expected outputs — is legitimate and worth keeping, just phrased plainly.

A one-word fix, and a reasonable first contribution.

### Acceptance criteria

- [ ] The comment is rewritten to state the actual concern about hardcoded MATLAB reference values
- [ ] A quick grep confirms no similar wording elsewhere in the tree

---

jasperdingg:
Closes #901, rephrased comment in test_lqnash.py. Confirmed that term doesn't appear elsewhere.


jasperdingg:
Hi @mmcky, quick follow up on #910!

quantecon/tests/test_lqnash.py
```diff
@@ -92,7 +92,9 @@ def test_nnash(self):
         tfi = np.linalg.inv(tf)
         xbar = tfi.dot(aaa[:2, 2])
 
-        # Define answers from matlab. TODO: this is ghetto
+        # Define answers from matlab. TODO: hardcoding these
+        # expected values is a fragile way to pin down the outputs
+
```
quantecon-services: Nice rewording, this states the actual concern clearly. 👍

One small nit: line 97 is a blank line that contains trailing whitespace (flake8 `W293`). It won't fail CI (the workflow only selects `F401,F405,E231`), but it's worth dropping. The original had no blank line between the comment and the array, so removing it also keeps the layout as it was:

```suggestion
        # Define answers from matlab. TODO: hardcoding these
        # expected values is a fragile way to pin down the outputs
```

---

quantecon-services (commented):
Hi @jasperdingg 👋 Thanks for picking this up, and apologies for the slow turnaround on a small PR.

This does exactly what #901 asked for. The rewording keeps the underlying TODO about hardcoded MATLAB reference values while stating it plainly, and I independently confirmed the grep: the line you changed is the only occurrence of that word anywhere in the tree. The branch also merges cleanly onto current `main`, so nothing else is needed on that front.

I've left one inline nit (a whitespace-only blank line) with a suggestion you can accept in one click. Once that's in, this is good to merge.

One friendly request. QuantEcon is currently drafting an AI disclosure policy for contributions (see [QuantEcon/qeps#13](https://github.com/QuantEcon/qeps/pull/13)). It isn't in force yet, but it would help us as we shape the policy if you could add a short note to the PR description saying whether AI tools were used in preparing this PR and, if so, which tools or models (e.g. "drafted with Claude Code / Copilot / ChatGPT, then reviewed and tested by hand"). "No AI tools used" is an equally welcome answer.

The goal is transparency rather than gatekeeping. AI-assisted contributions are fine, we'd just like the provenance recorded.

---

mmcky:
thanks @jasperdingg -- I will assume this is not AI written. 

Thanks for your contribution.

PR #913, opened 2026-07-26
CI: Measure Numba JIT coverage in a single dedicated job

### LINKED ISSUE #871: INFR: Coverage measurement excludes all numba-jitted code and counts the test tree
### Problem

`.coveragerc` lists `@jit`, `@jit\(.*nopython=True`, `@njit`, `@overload` and `@guvectorize\(.*nopython=True` under `[report] exclude_lines`. The intent is sound — jit-compiled code is not traced by CPython, so its lines would otherwise read as uncovered. But coverage.py excludes the **entire decorated function** when a decorator line matches an exclusion pattern, not just the decorator line. The result is that the hottest, most algorithmically intricate code in the library is not merely uncounted — it is unmeasurable.

Measured by walking the AST and comparing against what `coverage report` attributes to each file:

| Module | AST statements | Inside `@jit` | Statements coverage sees | Reported |
| --- | --- | --- | --- | --- |
| `optimize/scalar_maximization.py` | 79 | 77 | **2** | 100% |
| `optimize/root_finding.py` | 212 | 202 | **10** | 100% |
| `optimize/nelder_mead.py` | 113 | 108 | **4** | 100% |
| `optimize/linprog_simplex.py` | 169 | 150 | **18** | 100% |
| `markov/utilities.py` | 66 | 63 | **2** | 100% |
| **Package total** | **5,954** | **1,625 (27%)** | **4,019** | — |

The entire `quantecon.optimize` subpackage — 2,396 LOC of numerical optimisation — reports 100% while measuring almost nothing. The code is not untested (it has 1,052 LOC of tests); it is *unmeasured*: no instrument can tell which branches of the numerical kernels are exercised, or detect regression in that coverage.

A second, independent defect compounds it: `[run] source = quantecon` includes `quantecon/**/tests/`, so the test files count toward the percentage. CI computes and Coveralls displays **94%**; source-only coverage is **87%**.

### Proposed fix

Run the suite twice in CI: once normally, and once with `NUMBA_DISABLE_JIT=1`, which makes jitted functions execute as ordinary Python and become fully traceable; combine with `coverage combine`. This is the standard approach for numba projects and requires no library-code changes. At the same time, narrow `source`/`omit` so the test tree is excluded from the headline figure.

Expect the reported percentage to **fall** when this lands — that is the point; the new number will be true and becomes the baseline for future coverage work.

### Acceptance criteria

- [ ] CI runs the suite a second time under `NUMBA_DISABLE_JIT=1` and combines the results
- [ ] The jit-decorator patterns are removed from `exclude_lines`, or scoped to the non-instrumented pass only
- [ ] `source`/`omit` excludes `quantecon/**/tests/` from the reported figure
- [ ] The new baseline percentage is recorded in this thread
- [ ] `quantecon/optimize/*` reports statement counts consistent with AST statement counts

---

oyamad:
Closes #871

## Problem

Two defects in coverage measurement (details in #871):

1. `.coveragerc` listed the Numba decorators (`@jit`, `@njit`, `@overload`, `@guvectorize`) under `exclude_lines`. coverage.py excludes the **entire decorated function** when a decorator line matches, so ~1,700 statements of jitted code — the numerical core of the library — were invisible to coverage. `quantecon/optimize/root_finding.py` reported 100% while measuring 10 of its 212 statements.
2. `source = quantecon` included the test tree, inflating the headline figure.

## Approach

Instead of the `NUMBA_DISABLE_JIT=1` + `coverage combine` approach proposed in the issue, this uses **Numba's native JIT coverage** (`NUMBA_JIT_COVERAGE=1`, added in 0.61, fixed to respect the coverage configuration file in 0.62). It measures the real compiled code paths — including `@overload` and `@guvectorize` targets — requires no second suite run or result combining, and doesn't risk masking JIT-only failures the way disabling JIT does.

One subtlety: Numba emits coverage at compile time (during lowering), which `cache=True` functions skip when loaded from a warm on-disk cache. Nearly every jitted function here uses `cache=True`, so the coverage job points `NUMBA_CACHE_DIR` at a throwaway directory to force compilation.

## Changes

- **`.coveragerc`** — remove the five Numba decorator patterns from `exclude_lines`; add `*/tests/*` to `omit` so the reported figure measures library code only.
- **`.github/workflows/ci.yml`** — the 3 OS × 3 Python matrix now runs plain `pytest` (production configuration, no coverage overhead). A new single `coverage` job (ubuntu, Python 3.13) runs the suite once with `NUMBA_JIT_COVERAGE=1` and uploads to Coveralls. With one coverage producer, the old per-matrix Coveralls `if:` gate (which existed to avoid the parallel-upload race) is obsolete and dropped. The job reuses the ubuntu/3.13 conda cache.
- **`environment.yml`** — pin `numba>=0.62`; on older numba the env var would not respect `.coveragerc` and the reported total would silently drop by ~5 pp.

## Results

**The reported percentage falls from 93.77% to 90.55% — by design.** The old figure both excluded ~1,700 jitted statements and counted the test tree; the new number is the true source-only baseline (5,638 statements, 533 missed; 600 tests passing, measured locally with numba 0.63.1). Please read the Coveralls delta on this PR accordingly, not as a regression.

Per the issue's acceptance criteria, `quantecon/optimize/*` now reports statement counts matching the AST counts:

| Module | Measured before | Measured now |
| --- | --- | --- |
| `optimize/root_finding.py` | 10 | 212 |
| `optimize/nelder_mead.py` | 4 | 112 |
| `optimize/scalar_maximization.py` | 2 | 79 |
| `optimize/linprog_simplex.py` | 18 | 168 |

## Known caveat: `parallel=True` / `prange`

Functions compiled with `parallel=True` get partial rather than complete line attribution: the parfor pass rewrites the `prange` loop (and hoists some setup statements) into parallel gufunc constructs *before* the lowering phase where Numba emits coverage, so the rewritten instructions lose their source-line mapping. Concretely, in `gini_coefficient` (`quantecon/_inequality.py`, the library's only `prange` function) the loop **body** is correctly reported as covered, but the `for i in prange(n):` header and one hoisted setup line show as missed despite being executed — 2 spuriously-missed lines out of 5,638, included in the baseline above. Left as-is deliberately: an `exclude_lines` pattern for `prange(` would exclude the entire loop block (coverage.py excludes the whole block under an excluded compound statement), throwing away the body measurement we just gained.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,6 +4,8 @@
 relative_files = True
 source = quantecon
 omit =
+    # Exclude the test tree from the headline figure
+    */tests/*
     */python?.?/*
     */lib-python/?.?/*.py
     */lib_pypy/_*.py
@@ -29,10 +31,3 @@ exclude_lines =
     # Don't complain if non-runnable code isn't run:
     if 0:
     if __name__ == .__main__.:
-
-    # Don't count Numba nopython=True jit functions
-    @jit
-    @jit\(.*nopython=True
-    @njit
-    @overload
-    @guvectorize\(.*nopython=True
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,13 +54,48 @@ jobs:
     - name: Run Tests (pytest)
       shell: bash -l {0}
       run: |
+        pytest quantecon
+
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v7
+      with:
+        fetch-depth: 0
+    - name: Cache conda
+      uses: actions/cache@v6
+      env:
+        CACHE_NUMBER: 0
+      with:
+        path: ~/conda_pkgs_dir
+        key:
+          ${{ runner.os }}-3.13-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
+    - uses: conda-incubator/setup-miniconda@v4
+      with:
+        auto-update-conda: true
+        miniforge-version: latest
+        environment-file: environment.yml
+        python-version: "3.13"
+        auto-activate-base: false
+        use-only-tar-bz2: true
+        activate-environment: qe
+    - name: Run Tests with coverage
+      shell: bash -l {0}
+      env:
+        # Numba emits coverage for JIT functions at compile time (lowering),
+        # which cache=True functions skip when loaded from a warm on-disk
+        # cache. Enable JIT coverage and point the cache at a throwaway dir
+        # so every jitted function is compiled -- and thus reported -- here.
+        # Requires numba >= 0.62 (coverage respects .coveragerc from there on).
+        NUMBA_JIT_COVERAGE: "1"
+        NUMBA_CACHE_DIR: ${{ runner.temp }}/numba_cache
+      run: |
         coverage run -m pytest quantecon
         coverage lcov
     - name: Coveralls
       uses: coverallsapp/github-action@v2
-      # upload from one Linux job only (avoids the Coveralls parallel-build 422
-      # race); continue-on-error so an upload hiccup can't fail CI
-      if: runner.os == 'Linux' && matrix.python-version == '3.13'
+      # continue-on-error so an upload hiccup can't fail CI
       continue-on-error: true
       with:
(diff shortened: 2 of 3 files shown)


mmcky:
thanks @oyamad much appreciated. Merging now.

PR #914, opened 2026-07-26
MAINT: Test and clarify timing precision accessors

### LINKED ISSUE #889: MAINT: Add tests for quantecon/timings and deduplicate its accessors
### Problem

`quantecon/timings/` is the only subpackage with no test file at all, yet it exposes public API — `qe.timings.float_precision()` is reachable from the top-level namespace. It is 66 lines with two public functions and a module-level mutable global. Three things to address:

1. **No tests.** `float_precision()` has a validation branch (`ValueError` for non-integer or negative input) and get/set dual behaviour, neither exercised. The module-level global means tests need to restore state — worth getting right in the first test rather than retrofitting.
2. **Two names for one read.** `float_precision()` called with no argument and `get_default_precision()` return the same value by the same mechanism. One can go, or the relationship should be documented — as it stands a reader cannot tell which is canonical.
3. **Docstring examples are doctest-shaped but never executed** (adjacent: #866 on running doctests in CI), so they are unverified prose that reads as tested.

### Acceptance criteria

- [ ] A test module covering get, set, validation errors, and state restoration between tests
- [ ] The duplicate accessor removed, or its relationship to `float_precision()` documented
- [ ] Doctest examples either run under a configured doctest collector or are reformatted so they don't claim to be tested

---

lntutor:
## Summary

- add isolated coverage for timing precision reads, updates, and validation errors
- clarify `get_default_precision()` as the read-only equivalent of `float_precision()`
- replace uncollected doctest-style prompts with regular usage guidance
- clean existing whitespace violations in the timing module

Closes #889

## Validation

- `uv run --extra testing pytest quantecon/tests/test_timings.py -q`
- `uv run --extra testing flake8 quantecon/timings quantecon/tests/test_timings.py`

I also started the full `pytest quantecon -q` suite; it ran cleanly through 47% before the local terminal session ended unexpectedly.


quantecon-services:
Hi @lntutor, thank you for picking this up and for engaging with all three parts of #889 rather than just the tests. You were the first to claim the issue, and your reading of the accessor duplication and the never-run doctest example was correct.

I am closing this PR, but not because the work was poor. The reason is a mistake in the issue itself: the audit checked for a test *directory* under `quantecon/timings/` and missed that the module is already fully covered by `TestGlobalPrecision` in `quantecon/util/tests/test_timing.py` (get, set, all three `ValueError` paths, state restoration). So the test half of this PR duplicated coverage that already existed. The `>>>`-absence assertion also would not survive #920, which is moving the project toward *running* docstring examples rather than removing them.

The docstring half was the valuable part, and I have carried it forward in #959 with credit to you: it documents `get_default_precision()` as the read-only equivalent of `float_precision()`, repairs the one example line that fails under strict doctest (keeping the `>>>` form so #920 can execute it), and picks up your whitespace and end-of-file fixes. I hope that is an acceptable outcome; please say if you would rather rework this PR to that scope yourself and I will close #959 instead.

I have also closed #889 with a note about what went wrong. If you would like another issue, the [2026-07 Technical Debt Audit milestone](https://github.com/QuantEcon/QuantEcon.py/milestone/12) has plenty, and I would be happy to review.

Thanks again, and apologies for the wasted effort on the test side.

---

Issue #903, opened 2026-07-25
Technical debt audit — July 2026 (tracking issue)


mmcky:
A technical-debt study of QuantEcon.py at v0.11.4 (commit `28d4b3b`) identified 35 actionable items, filed as the sub-issues below and grouped under the milestone **2026-07 Technical Debt Audit**. This issue is structure, not work: it carries no Type label, per QEP-2 (QuantEcon/qeps#2).

**Method.** Two independent AI-assisted audits (different models, one local, one cloud) were run against the same commit, compared, merged, and cross-validated: every disagreement between the two was re-verified against the working tree or the GitHub API before filing, and findings already covered by open issues were linked rather than re-filed. The full study reports are held locally by @mmcky.

**State of the library.** The headline is good: 600 tests pass cleanly on the current stack (NumPy 2.x, numba 0.66, Python 3.12–3.14 across three OSes), and recent maintenance is active and substantive. The debt found is overwhelmingly structural and procedural — deferred decisions, stale metadata, and quality signals that do not measure what they appear to measure — rather than a sign of neglect.

### High priority

| Issue | Theme |
| --- | --- |
| #868 | Vendor `rng_integers` — private SciPy import, breaks on any SciPy release |
| #869 | Install metadata contradicts itself (`requires-python`, floors, classifiers) |
| #870 | `fetch_nb_dependencies` writes HTTP error pages and reports success |
| #871 | Coverage excludes all numba-jitted code and counts the test tree |
| #872 | Supply-chain hardening: permissions, SHA pinning, Trusted Publishing |

### Normal priority

| Issue | Theme |
| --- | --- |
| #873 | NumPy 2 workflow: broken trigger, likely obsolete |
| #874 | Missing `assert` in `test_ddp.py` sparse-Q check |
| #875 | Contract tests for the deprecation-shim layer |
| #876 | `mc_compute_stationary`/`mc_sample_path`: warn for real or bless |
| #877 | RandomState → Generator migration policy (discuss) |
| #878 | Migrate `IVP` off legacy `scipy.integrate.ode` |
| #879 | `scipy.sparse` matrix → array migration plan |
| #880 | Slim mandatory deps: `sympy`, `requests` |
| #881 | Dead packaging config (MANIFEST.in, flit excludes, nose relics) |
| #882 | Code-hygiene quick wins |
| #883 | Review eight F841 unused locals in numerical modules |
| #884 | Ratchet the CI lint gate (follow-up to #457) |
[...]

PR #915, opened 2026-07-26
TEST: add tests for quantecon/timings (#889)

bwcummings1:
Addresses #889. The `quantecon/timings` subpackage had no tests; this adds a
`quantecon/timings/tests/` package (an `__init__.py` marker plus
`test_timings.py`, following the per-subpackage `tests/` convention used
elsewhere in the repo) with 9 tests covering the
public API:

- `float_precision()` getter (no argument returns the current precision)
- `float_precision(n)` setter, observable via both the getter and
  `get_default_precision()`
- accepts `0` and larger valid integers
- raises `ValueError` for a negative int, a float, and a string
  (parametrized)
- a failed set leaves the global precision unchanged
- `get_default_precision()` reports the live global value

An autouse fixture snapshots and restores the module global so the tests
never leak precision changes into the rest of the suite.

Coverage for `quantecon/timings/timings.py` goes from 0% to 100%
(measured with `coverage run -m pytest quantecon/timings` on this branch).
`flake8 --select=F401,F405,E231` is clean. No production code is modified.

**AI disclosure:** these tests were drafted by an AI system — poolside
Laguna S 2.1 in an orchestrated harness (orchestrator: Claude Fable 5) —
from #889's acceptance criteria and the module source. The commit message
carries model/usage detail. I reviewed and validated the tests and the
coverage delta and take responsibility for this contribution.


quantecon-services:
Hi @bwcummings1, thank you for this. It is a genuinely well-made PR: the per-subpackage `tests/` layout is the one we want, the autouse fixture that yields the snapshot is the right pattern, the zero / large-value / failed-set-leaves-state-unchanged cases are thoughtful, and the AI disclosure in both the PR body and the commit message is exactly what we are asking contributors for. CI is green across the whole matrix.

Unfortunately I have to close it, and the reason is our mistake rather than yours. When #889 was written, the audit checked for a test *directory* under `quantecon/timings/` and found none. It missed that the module is already fully covered by `TestGlobalPrecision` in `quantecon/util/tests/test_timing.py`, which exercises get, set, all three `ValueError` paths, and state restoration. That is why Coveralls reported coverage "remained the same": the 0% to 100% delta only appears when `pytest quantecon/timings` is run in isolation. Merging would leave us with the same assertions in two places.

I have closed #889 with a note explaining this, and left a suggestion on the tracking issue #903 so the remaining "no tests" findings are verified by coverage first. If you would like another issue to pick up, anything under the [2026-07 Technical Debt Audit milestone](https://github.com/QuantEcon/QuantEcon.py/milestone/12) is fair game, and I would be glad to see another PR from you.

Thanks again, and sorry for the run-around.

---

Issue #869, opened 2026-07-25
FIX: Correct install metadata: requires-python, dependency floors, classifiers


mmcky:
### Problem

`pyproject.toml` contradicts itself and reality:

- `requires-python = ">=3.7"` — Python 3.7 has been EOL since June 2023, and 3.8–3.11 are neither in CI nor in the classifiers, which list exactly 3.12 / 3.13 / 3.14.
- Floors `numpy>=1.17.0` (2019), `numba>=0.49.0` (2020), `scipy>=1.5.0` (2020) permit pip to build environments untested for years — while the code has moved on (`util/compat.py` exists for NumPy≥2 semantics; `_dle.py` carries a NumPy 2.4 workaround). The floors are also internally inconsistent with the classifiers: numba 0.49 cannot run on any Python version the classifiers name (the first numba supporting 3.12 is 0.59).

Net effect: `pip install quantecon` succeeds on configurations we do not support, and failure surfaces later as user bug reports. Wrong metadata is a shipped defect.

### Proposed change

1. Set `requires-python = ">=3.12"` to match CI and classifiers (or extend CI first if broader support is intended — either way, make the three agree).
2. Raise floors to the oldest combination actually exercised, e.g. the versions resolved by `environment.yml` on the oldest supported Python.
3. While in the file: decide `Development Status :: 4 - Beta` (12+ years in; `5 - Production/Stable` likely reflects reality) — one-line change, same edit.
4. Add a release-checklist line: metadata review on every Python-version or floor change.

### Acceptance criteria

- [ ] `requires-python`, classifiers, and the CI matrix name the same version set
- [ ] `pip install` on a now-unsupported Python is refused by the resolver (users there keep resolving to older quantecon releases — the correct outcome)
- [ ] Floors verified by at least one CI leg or a documented local run

---

_From the July 2026 technical-debt audit (AI-assisted; claims verified against `28d4b3b` on 2026-07-25)._

oaksprout:
Would you want this one done? — Fix quantecon's `pyproject.toml` so its declared Python-version floor and dependency-version floors (numpy/numba/scipy) actually match what CI tests and what the trove classifiers claim — currently it advertises support far older than anything actually exercised or working. If yes, we'll do it and open a PR; if not, say so (or just ignore this) and we won't touch it.

[...]

PR #917, opened 2026-08-01
ENH: Extend random.draw to accept a random number generator

### LINKED ISSUE #916: Extend random.draw to accept a random number generator
Raised by @oyamad in review of #864 ([comment](https://github.com/QuantEcon/QuantEcon.py/pull/864#discussion_r3638453626)): `quantecon.random.draw` should be able to accept a random number generator, which recent Numba versions make possible.

## Current behaviour

`draw` is the only function in `quantecon.random` with no `random_state` argument. It calls `np.random.random()` directly, in both the pure-Python body and the `@overload` implementation, so the generator it uses depends on where it is called from:

| Call site | Generator used | How to seed it |
| --- | --- | --- |
| Python | NumPy's legacy global state | `np.random.seed(...)` at Python level |
| Inside a jitted function | Numba's own internal state | `np.random.seed(...)` **inside** the jitted function |

These are two different generators. A `np.random.seed` call at Python level has no effect on the jitted path — a sharp edge for anyone calling `draw` from inside `@njit` code and expecting reproducibility. Both are Mersenne Twister and produce the same stream for a given seed, but they advance independently.

This leaves `draw` inconsistent with `probvec` and `sample_without_replacement`, which both accept `random_state` and work with a `np.random.Generator`.

## Feasibility

Verified on numba 0.62.1 / numpy 2.3.5 / python 3.13.9 — Numba accepts a `np.random.Generator` in nopython mode and reproduces the host stream exactly:

```python
from numba import njit
import numpy as np

@njit
def h(rng, n):
    return rng.random(n)

rng = np.random.default_rng(1234)
assert np.array_equal(h(rng, 5), np.random.default_rng(1234).random(5))  # passes
```

So the blocker that originally justified `draw`'s signature is gone.

## Scope

This is a signature and behaviour change, not a documentation fix, which is why it was split out of #864 rather than carried in it. Points worth settling before implementing:

- Whether the new parameter matches the module's existing `random_state` name or moves to a `rng` name, and whether that decision should apply module-wide rather than to `draw` alone.
- How the pure-Python and `@overload` paths stay in agreement — the overload needs its own handling for a passed generator, and the two implementations should not diverge in output for the same generator and seed.
- Whether `check_random_state` (used by the other two functions) can be reached from nopython mode, or whether the jitted path needs a narrower contract that accepts only a `Generator`.
- Backward compatibility for existing callers that rely on the current global-state behaviour, including jitted callers seeding via `np.random.seed` inside jitted code.

## Related

The docstring in `quantecon/random/utilities.py` now documents the two-generator behaviour above, added in 08682a7 on #864. That Notes block should be replaced with the new parameter once this lands.

mmcky:
> **Update (2026-08-15, edcd4c2):** the argument is named `rng` and the documented contract is `None` | `np.random.Generator`, but it is enforced only at the jitted path (compile-time `TypingError`) — the pure-Python body is a thin unchecked fallback whose behaviour for other inputs is unspecified, per review. The `requires-python` bump is reverted (deferred to #869) and the numba floor is `>=0.56.0`, the minimum this feature needs. The decision table below records the original draft rationale; commit 59cf867 (2026-08-14) introduced the `rng` naming.

Closes #916. Opened as a **draft** because the issue asked for four design points to be settled before implementing — they are answered below and are the main thing worth reviewing.

`quantecon.random.draw` was the only function in `quantecon.random` with no `random_state` argument. It called `np.random.random()` directly in both the pure-Python body and the `@overload` implementation, so the generator it used depended on where it was called from, and a `np.random.seed` at Python level had no effect on the jitted path. This adds `random_state`, bringing `draw` into line with `probvec`, `sample_without_replacement` and `DiscreteRV.draw`.

## Decisions on the four points raised in #916

| Question | Decision | Reasoning |
| --- | --- | --- |
| Parameter name: `random_state` or `rng`? Module-wide? | **`random_state`**, trailing: `draw(cdf, size=None, random_state=None)`. **Not** a module-wide rename. | `probvec`, `sample_without_replacement` and `DiscreteRV.draw` already take `random_state`; `draw` was the one hole. Moving the whole module to an `rng` name is a contract change (it would drop `RandomState` support) and deserves its own issue and deprecation cycle rather than a drive-by on one function. |
| How do the pure-Python and `@overload` paths stay in agreement? | **By test, not by construction.** The jitted sized branch keeps its hand-written `searchsorted` loop rather than mirroring the Python body's vectorised call, and `TestDraw.test_python_jitted_agree` pins the two paths together across 3 seeds × 4 sizes. | Numba does support the array form, and the two produce bit-identical output — but the loop measured roughly 2× faster at `size >= 1000` on this machine, so making the bodies textually identical would ship a silent performance regression under an "add `random_state`" release note. |
| Is `check_random_state` reachable from nopython mode? | **No — the jitted path takes a narrower contract:** `None` or a live `np.random.Generator` only. Everything else raises `numba.TypingError` at compile time, with a message that says what to pass instead. | `check_random_state` is pure Python, Numba cannot type `np.random.RandomState` at all, and it cannot construct a `Generator` from an integer seed inside nopython mode. There is no way to widen the jitted contract without silently remapping a seed to a different stream. |
| Backward compatibility | **Purely additive; no deprecation.** | `check_random_state(None)` returns `np.random.mtrand._rand` by identity and is non-consuming, so the Python `None` path is byte-identical to the old `np.random.random(...)`, not merely statistically equivalent. The jitted `None` path still uses Numba's internal state, so a jitted caller seeding with `np.random.seed` inside the jitted function is unaffected. The new parameter is trailing-with-default and there was no third positional slot, so no existing call form changes meaning. All four pre-existing `draw` tests and both existing jitted wrappers pass unmodified. |

A `Generator` passed into nopython mode reproduces the host stream bit-for-bit and its state is mutated in place, so one generator can be shared across Python and jitted calls and it stays in sync. That is now the recommended way to get reproducible draws from a call site that may be jit-compiled, and it is what the new docstring example demonstrates.

## Points I would like a decision on

**The `numba>=0.49.0` → `numba>=0.59.0` floor bump in `pyproject.toml`.** This change needs `types.NumPyRandomGeneratorType`, which numba first shipped in **0.56.0** (absent in 0.55.0). I set the floor at 0.59 rather than 0.56 because 0.59 is the first release with cp312 wheels, and the classifiers and the CI matrix already declare 3.12/3.13/3.14 — a floor below that names versions installable on no Python this package supports. Happy to drop this hunk if you would rather handle dependency metadata separately; the alternative of a `getattr` guard is worse, since it would silently degrade the feature into "every `random_state` is rejected" on an old numba rather than failing loudly. Note `requires-python = ">=3.7"` two lines above is stale by the same argument, but #864 already changes that line — see below.

**Merge order against #864.** #864 (`docs/consistency-audit`) rewrites this same docstring: it added a Notes block in 08682a7 documenting the two-generator behaviour, which #916 itself says should be replaced once this lands. That block is superseded here. Both PRs also touch `pyproject.toml` within three lines of each other. Whoever merges second gets a small, obvious conflict in `quantecon/random/utilities.py` and possibly `pyproject.toml`. My suggestion is to land #864 first and rebase this on top, since #864 is the larger diff. I left the one-line `quantecon/random/__init__.py` docstring fix (`3. draw`) out of this PR because #864 already makes exactly that change.

## Out of scope, but found along the way

The pure-Python body dispatches on `isinstance(size, int)` while the overload dispatches on `types.Integer`. These disagree for numpy integers: `qe.random.draw(cdf, np.int64(10))` silently returns **one scalar** from Python and a **10-element array** from jitted code, with no error. `size=True` is the mirror image (Python raises `TypeError`, jitted returns a scalar). Both predate this change and are untouched here — fixing them is a silent behaviour change that would muddy this PR's release note and its bisect story. Tracked separately in #918.

## Tests

Thirteen new tests in `quantecon/random/tests/test_utilities.py`, covering: Python/jitted agreement for the same Generator; reproducibility on both paths; in-place generator state mutation across a mixed chain of jitted and Python calls; that exactly one variate is consumed per draw, so a shared generator does not desynchronise; the keyword call form from jitted code; int-seed and `RandomState` on the Python path; that a seed and a `default_rng` of the same integer differ; byte-identity of the `None` path against the legacy global stream; that all three Numba spellings of "no random_state" compile (omitted arrives as Python `None`, explicit `None` and a forwarded default both arrive as `types.none`); that in-jit `np.random.seed` is still reproducible; and the three compile-time rejections.

The rejection tests assert on tokens that can only come from `draw`'s own message. A plain `'random_state' in str(e)` looks right but is vacuous — Numba echoes the caller's source line alongside the error, so the parameter name appears in the message whether or not the implementation produced it.

Each new test was checked by mutation: reverting the Generator branch to `np.random`, removing `check_random_state`, dropping either clause of the `None` predicate, swapping the two overload branches, and over-consuming a variate on any of the four Generator branches each fail at least one test.

## Verification

Run on numba 0.62.1 / numpy 2.3.5 / python 3.13.9:

- `pytest quantecon` — 613 passed
- `pytest quantecon/random/tests/test_utilities.py` — 22 passed
- `NUMBA_DISABLE_JIT=1 pytest quantecon/random/tests/test_utilities.py` — 19 passed, 3 skipped (the three compile-time rejection tests are `skipif`-guarded, since with JIT off the `@njit` wrappers run the Python body and `check_random_state` accepts the int seed)
- `flake8 --select=F401,F405,E231 quantecon` — clean
- Full `flake8` on both touched files — clean apart from the pre-existing `E305` at `utilities.py:92`
- The new `draw` doctest passes under `--doctest-modules`. Note doctests are not run in CI, and the two sibling examples in this module fail there today for a pre-existing reason (they assume an ambient `qe`); the new example is self-contained.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ keywords = [
     'economics'
 ]
 dynamic = ["description", "version"]
-requires-python = ">=3.7"
+requires-python = ">=3.12"
 dependencies = [
     'numba>=0.59.0',
     'numpy>=1.17.0',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,9 @@ keywords = [
     'economics'
 ]
 dynamic = ["description", "version"]
-requires-python = ">=3.12"
+requires-python = ">=3.7"
 dependencies = [
-    'numba>=0.59.0',
+    'numba>=0.56.0',
     'numpy>=1.17.0',
     'requests',
     'scipy>=1.5.0',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ keywords = [
 dynamic = ["description", "version"]
 requires-python = ">=3.7"
 dependencies = [
-    'numba>=0.49.0',
+    'numba>=0.59.0',
     'numpy>=1.17.0',
     'requests',
     'scipy>=1.5.0',
--- a/quantecon/random/tests/test_utilities.py
+++ b/quantecon/random/tests/test_utilities.py
@@ -1,5 +1,5 @@
 """
-Tests for util/random.py
+Tests for random/utilities.py
 
 Functions
 ---------
@@ -77,13 +77,13 @@ def draw_jitted(cdf, size=None):
 
 
 @njit
-def draw_jitted_rs(cdf, size, random_state):
-    return draw(cdf, size, random_state)
+def draw_jitted_rng(cdf, size, rng):
+    return draw(cdf, size, rng)
 
 
 @njit
-def draw_jitted_rs_kw(cdf, random_state):
-    return draw(cdf, random_state=random_state)
+def draw_jitted_rng_kw(cdf, rng):
+    return draw(cdf, rng=rng)
 
 
 @njit
@@ -92,9 +92,9 @@ def draw_jitted_explicit_none(cdf, size):
 
 
 @njit
-def draw_jitted_fwd(cdf, size=None, random_state=None):
+def draw_jitted_fwd(cdf, size=None, rng=None):
     # Forwards its own defaults, which reach the overload as types.none
-    return draw(cdf, size, random_state)
+    return draw(cdf, size, rng)
 
 
 @njit
@@ -104,9 +104,9 @@ def draw_jitted_seeded(cdf, size, seed):
 
 
 @njit
-def draw_jitted_optional_rs(cdf, size, random_state, flag):
-    rs = random_state if flag else None
-    return draw(cdf, size, rs)
+def draw_jitted_optional_rng(cdf, size, rng, flag):
+    rng_or_none = rng if flag else None
+    return draw(cdf, size, rng_or_none)
 
 
 class TestDraw:
@@ -145,7 +145,7 @@ class TestDraw:
             atol = 1e-2
             assert_allclose(pmf_computed, self.pmf, atol=atol)
 
-    # random_state: a Generator behaves the same on both paths #
+    # rng: a Generator behaves the same on both paths #
 
     def test_python_jitted_agree(self):
         # The contract that keeps the two implementations in step: for
@@ -153,20 +153,20 @@ class TestDraw:
         for seed in [0, 1234, 20260801]:
             for size in [1, 10, 1000]:
                 out_py = draw(self.cdf, size, np.random.default_rng(seed))
-                out_jit = draw_jitted_rs(self.cdf, size,
-                                         np.random.default_rng(seed))
+                out_jit = draw_jitted_rng(self.cdf, size,
+                                          np.random.default_rng(seed))
                 assert_array_equal(out_py, out_jit)
 
             # Scalar draws are compared as arrays: the Python path
             # returns np.int64 and the jitted path a Python int.
             out_py = draw(self.cdf, None, np.random.default_rng(seed))
-            out_jit = draw_jitted_rs(self.cdf, None,
-                                     np.random.default_rng(seed))
+            out_jit = draw_jitted_rng(self.cdf, None,
+                                      np.random.default_rng(seed))
             assert_array_equal(np.asarray(out_py), np.asarray(out_jit))
 
     def test_generator_reproducible(self):
         for size in [None, 10]:
-            for func in [draw, draw_jitted_rs]:
+            for func in [draw, draw_jitted_rng]:
                 out0 = func(self.cdf, size, np.random.default_rng(1234))
                 out1 = func(self.cdf, size, np.random.default_rng(1234))
                 assert_array_equal(np.asarray(out0), np.asarray(out1))
@@ -176,8 +176,8 @@ class TestDraw:
         # copied at the boundary, so that successive calls continue the
         # stream rather than restarting it.
         rng = np.random.default_rng(1234)
-        parts = [draw_jitted_rs(self.cdf, 5, rng),
-                 draw_jitted_rs(self.cdf, 5, rng),
+        parts = [draw_jitted_rng(self.cdf, 5, rng),
+                 draw_jitted_rng(self.cdf, 5, rng),
                  draw(self.cdf, 5, rng)]
         expected = draw(self.cdf, 15, np.random.default_rng(1234))
         assert_array_equal(np.concatenate(parts), expected)
@@ -188,7 +188,7 @@ class TestDraw:
         # Generator for every later consumer. Complements
         # test_generator_state_advances, which covers the in-place
         # mutation half of the same contract.
-        for func in [draw, draw_jitted_rs]:
+        for func in [draw, draw_jitted_rng]:
             for size in [None, 1, 10]:
                 n = 1 if size is None else size
                 rng = np.random.default_rng(1234)
@@ -198,25 +198,11 @@ class TestDraw:
                 assert_array_equal(rng.random(4), ref.random(4))
 
     def test_generator_keyword_form_in_jit(self):
-        out_py = draw(self.cdf, random_state=np.random.default_rng(1234))
-        out_jit = draw_jitted_rs_kw(self.cdf, np.random.default_rng(1234))
+        out_py = draw(self.cdf, rng=np.random.default_rng(1234))
+        out_jit = draw_jitted_rng_kw(self.cdf, np.random.default_rng(1234))
         assert_array_equal(np.asarray(out_py), np.asarray(out_jit))
 
-    # random_state: the Python path keeps check_random_state's breadth #
-
-    def test_int_seed_python_path(self):
-        for size in [None, 10]:
-            out_seed = draw(self.cdf, size, 1234)
-            out_rs = draw(self.cdf, size, np.random.RandomState(1234))
-            assert_array_equal(np.asarray(out_seed), np.asarray(out_rs))
-
-    def test_int_seed_and_generator_differ(self):
-        # Documented in Notes: same integer, different stream.
-        out_seed = draw(self.cdf, 10, 1234)
-        out_gen = draw(self.cdf, 10, np.random.default_rng(1234))
-        assert_(not np.array_equal(out_seed, out_gen))
-
-    # random_state=None: unchanged behaviour on both paths #
+    # rng=None: unchanged behaviour on both paths #
 
     def test_none_matches_legacy_global(self):
         for size in [None, 10]:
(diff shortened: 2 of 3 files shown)


mmcky:
@oyamad just started work on this. This will remain in draft until I get a chance to do a detailed review. I'm not sure we need all these different states supported and we may just upgrade / deprecate.

oyamad:
@mmcky Thanks!

Here's my argument:
1. I would like to take this stance: `draw` is primarily for use in a jitted function; it "happens" to work in pure Python, but how it behaves is implementation detail, and thus unspecified/unsupported by the public API.
2. We are now to encourage the use pattern, `rng = np.random.default_rng(1234); function_with_randomness(..., rng=rng)`. Thus add a new optional argument `rng`, where we don't need to (or shouldn't) try to mimic the existing `random_state`, so it only accepts `None` (default) or Generator; in particular, don't accept integer. This will simplify some of the issues.
3. [This was pointed out by Claude Fable 5] The policy above is stricter than "[SPEC 7](https://scientific-python.org/specs/spec-0007/)", which recommends normalizing `rng` via `np.random.default_rng(rng)` (hence accepting int seeds). That normalization step is not possible in nopython mode. So my proposal is: in Numba functions, take `rng` as `None` | `Generator` only.
4. Regarding merge order: The only restriction is that doctest be activated after this PR is merged.
5. On #918: That is a bug I introduced when I rewrote the code with `@overload`... Just replace `if isinstance(size, int):` with `if isinstance(size, (int, np.integer)):`. That's it.

quantecon/random/utilities.py
```diff
@@ -170,7 +171,7 @@ def _sample_without_replacement(n, r, out):
 
 
 # Pure python implementation that will run if the JIT compiler is disabled
-def draw(cdf, size=None):
+def draw(cdf, size=None, random_state=None):
     """
     Generate a random sample according to the cumulative distribution
     given by `cdf`. Jit-complied by Numba in nopython mode.
```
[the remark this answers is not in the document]
mmcky: Fixed in 59cf867 — the summary line now reads "JIT-compiled". (The typo predated this PR, but the docstring was being rewritten here anyway.)

quantecon/random/tests/test_utilities.py
```diff
@@ -5,13 +5,16 @@
 ---------
 probvec
 sample_without_replacement
+draw
```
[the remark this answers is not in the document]
mmcky: Fixed in 59cf867 — the header now reads "Tests for random/utilities.py". The stale path predated this PR (the module moved years ago) but this was the right place to correct it.

pyproject.toml
```diff
@@ -23,7 +23,7 @@ keywords = [
 dynamic = ["description", "version"]
 requires-python = ">=3.7"
 dependencies = [
-    'numba>=0.49.0',
+    'numba>=0.59.0',
     'numpy>=1.17.0',
```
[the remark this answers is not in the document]
mmcky: Fixed in 59cf867 — `requires-python` is now `>=3.12`, matching the classifiers and the numba floor. The PR originally deferred this line to #864, but the merge order settled in the discussion below puts this PR first, so the metadata fix belongs here; #864 will pick up a trivial one-line conflict when it rebases.

mmcky:
@oyamad Thanks — all five points adopted in 59cf867.

**Points 1–3 (`rng`, `None` | `Generator` only).** The argument is renamed to `rng` and the narrow contract now holds on *both* paths, not just the jitted one: the Python body raises `TypeError` for anything that isn't `None` or a `np.random.Generator`, with the same "build one with `np.random.default_rng(seed)`" guidance the jitted path gives in its compile-time `TypingError`. `check_random_state` is no longer involved. I took your point 3 as implying the Python path should be equally strict — accepting seeds on one path only would let the two paths disagree, which is the thing this PR exists to prevent. The docstring now leads with your point 1 (intended primarily for jitted use), and the int-seed/RandomState material is gone from Notes; the example uses a shared `Generator`.

Tests: the two seed/RandomState *acceptance* tests are gone, and the rejection tests now run on both paths (they no longer need `skipif` under `NUMBA_DISABLE_JIT`, since the Python body rejects too). Full suite: 611 passed; with JIT disabled: 19 passed, 1 skipped; the new doctest passes.

**Point 4 (merge order).** Understood — this PR first, doctest activation after. Since this now lands first, I also bumped `requires-python` to `>=3.12` here (Copilot flagged the inconsistency with the new numba floor); the doctest branch gets a trivial rebase conflict on that line.

**Point 5 (#918).** I'll open a separate one-line PR with `isinstance(size, (int, np.integer))` as you suggest, keeping it out of this diff.

One follow-up worth tracking: `draw` is now the one function in `quantecon.random` taking `rng` while `probvec` and `sample_without_replacement` take `random_state`. If the direction is a SPEC-7-style module-wide move to `rng`, I'm happy to open an issue proposing that migration (with a deprecation cycle for `random_state`) so the naming split is deliberate rather than accidental.

quantecon/random/utilities.py
```diff
@@ -236,38 +234,46 @@ def draw(cdf, size=None, random_state=None):
...
     >>> import quantecon as qe
     >>> cdf = np.cumsum([0.4, 0.6])
-    >>> qe.random.draw(cdf, 10, random_state=1234)
-    array([0, 1, 1, 1, 1, 0, 0, 1, 1, 1])
-    >>> int(qe.random.draw(cdf, random_state=1234))
-    0
+    >>> rng = np.random.default_rng(1234)
+    >>> qe.random.draw(cdf, 10, rng=rng)
+    array([1, 0, 1, 0, 0, 0, 0, 0, 1, 0])
+    >>> int(qe.random.draw(cdf, rng=np.random.default_rng(1234)))
```
mmcky: We need to remove this `int` in favour of just returning the __repr__

pyproject.toml
```diff
@@ -21,9 +21,9 @@ keywords = [
     'economics'
 ]
 dynamic = ["description", "version"]
-requires-python = ">=3.7"
+requires-python = ">=3.12"
```
oyamad: Maybe make this change in a separate PR (cf. #869)?

pyproject.toml
```diff
@@ -21,9 +21,9 @@ keywords = [
...
 dependencies = [
-    'numba>=0.49.0',
+    'numba>=0.59.0',
```
oyamad: This is totally fine, but `0.56` is sufficient for this PR.

quantecon/random/utilities.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 from numba import guvectorize, types
+from numba.core.errors import TypingError
```
oyamad:
```suggestion
from numba import TypingError
```

oyamad (changes requested):
@mmcky Thanks for the revision.

My earlier suggestion was: leave the behavior of the Python version unspecified/unsupported, by which I meant "we promise nothing". So I wouldn't do any input check, and the following would be enough:

```python
def draw(cdf, size=None, rng=None):
    if rng is None:
        rng = np.random
    r = rng.random()
    return np.searchsorted(cdf, r, side='right')
```

In particular, we don't need to reject `RandomState`. This way we can keep the Python version as thin as possible. What do you think?

quantecon/random/utilities.py
mmcky: Done in edcd4c2 — the example now shows `np.int64(1)`.

pyproject.toml
mmcky: Reverted in edcd4c2 — left to #869.

pyproject.toml
mmcky: Lowered to 0.56.0 in edcd4c2; #869 can take the floors further.

quantecon/random/utilities.py
mmcky: Applied in edcd4c2.

mmcky:
Agreed — done in edcd4c2. The Python body is now the thin version (no input check; the `size` branch stays). The `None` | `Generator` contract is enforced only by the overload's `TypingError`, and the docstring now says the Python path is an unchecked fallback, unspecified for other inputs. The two Python-path rejection tests are removed and the three compile-time ones are again skipped under `NUMBA_DISABLE_JIT`. Full suite: 611 passed; with JIT off: 17 passed, 3 skipped.

oyamad (approved):
@mmcky Thanks! This is what I had in mind.

One minor comment: In the Notes, I would not mention the behavior of the Python version at all. But not blocking.

mmcky:
Done in 4d83d40 — the Notes no longer describe the Python path at all. I also dropped the Python-vs-jitted line from the Examples, since it made the same promise; `test_python_jitted_agree` still holds the two paths together internally.

mmcky:
thanks @oyamad for all of your valuable review comments.

PR #919, opened 2026-08-01
FIX: Align random.draw size dispatch for numpy integers

### LINKED ISSUE #918: random.draw: size dispatch diverges between the Python and jitted paths for numpy integers
`quantecon.random.draw` decides whether `size` requests an array of draws or a single scalar draw in two different places, using two different tests that do not agree:

| Path | Test | Location |
| --- | --- | --- |
| Pure Python body | `isinstance(size, int)` | `quantecon/random/utilities.py` |
| `@overload` implementation | `isinstance(size, types.Integer)` | `quantecon/random/utilities.py` |

A NumPy integer is not a Python `int`, but it *is* a `types.Integer` to Numba. A Python `bool` is the reverse: it passes `isinstance(x, int)` but Numba types it as `Boolean`, not `Integer`. So the same call returns different things depending on whether the call site is jit-compiled — with no error either way.

## Reproducing

Verified on numba 0.62.1 / numpy 2.3.5 / python 3.13.9, against current `main`:

```python
import numpy as np
from numba import njit
from quantecon.random import draw

@njit
def draw_jit(cdf, size):
    return draw(cdf, size)

cdf = np.cumsum([0.4, 0.6])

draw(cdf, np.int64(10))       # np.int64(1)                    <- ONE scalar
draw_jit(cdf, np.int64(10))   # array([1, 0, 0, 1, 1, 1, 1, 1, 0, 0])   <- 10 draws

draw(cdf, True)               # TypeError
draw_jit(cdf, True)           # 1        <- a scalar draw
```

Full dispatch table:

| `size` | `isinstance(size, int)` | Numba `typeof` | `isinstance(_, types.Integer)` | Python result | Jitted result |
| --- | --- | --- | --- | --- | --- |
| `10` | True | `int64` | True | 10 draws | 10 draws |
| `np.int64(10)` | **False** | `int64` | **True** | **1 scalar** | **10 draws** |
| `np.int32(10)` | **False** | `int32` | **True** | **1 scalar** | **10 draws** |
| `np.uint8(10)` | **False** | `uint8` | **True** | **1 scalar** | **10 draws** |
| `True` | **True** | `bool` | **False** | **`TypeError`** | **1 scalar** |
| `np.bool_(True)` | False | `bool` | False | 1 scalar | 1 scalar |
| `10.0` | False | `float64` | False | 1 scalar | 1 scalar |

The numpy-integer rows are the ones that matter in practice. Nothing warns, nothing raises — the caller gets a scalar where they asked for an array, and downstream code sees a shape it did not expect.

## Why this is easy to hit

A NumPy integer is what you get from most numpy-valued expressions, so `size` arrives as `np.int64` without the caller doing anything unusual: an element of an integer array (`counts[i]`), a reduction (`arr.sum()`, `arr.argmax()`), a `//` result involving an array, or a value read out of a structured/record array. Plain `len(x)` and `arr.shape[0]` return Python ints and are fine, which is part of why this survives casual testing.

## Suggested fix

Align the Python body with the overload:

```python
if isinstance(size, (int, np.integer)) and not isinstance(size, bool):
```

The `not isinstance(size, bool)` clause is needed because `bool` subclasses `int` in Python but Numba types it as `Boolean`; without it the two paths would still disagree for `size=True`. I verified this predicate agrees with the overload's `isinstance(t, types.Integer)` on every case in the table above, including `None`.

There is a compatibility judgement for the team here rather than a purely mechanical fix. `draw(cdf, np.int64(10))` currently returns a scalar from Python, and the fix changes that to a 10-element array. That is the behaviour the caller asked for and the behaviour the jitted path already gives, but it *is* a silent behaviour change for anyone who relied on the current result. It may deserve a release note, and possibly a decision on whether `size=True` should raise rather than quietly mean "one draw".

## Provenance

Found while implementing #916 (PR #917), which adds a `random_state` argument to `draw`. This divergence predates that work and is untouched by it — deliberately left out so the behaviour change would not ride along in a feature PR and muddy the release note and the bisect story. `probvec` and `sample_without_replacement` are not affected; they build `size` internally and never dispatch on its type.

kocaemre:
Closes #918.

## Summary
- Treat numpy integer sizes (for example `np.int64(10)`) the same as Python `int` in the pure-Python `random.draw` path.
- Keep boolean `size` values on the scalar branch so Python and jitted dispatch agree.
- Add focused regression tests covering Python and `@njit` callers for both cases.

## Test plan
- [x] RED: `pytest quantecon/random/tests/test_utilities.py::TestDraw::test_numpy_integer_size_returns_array quantecon/random/tests/test_utilities.py::TestDraw::test_bool_size_is_treated_as_scalar -q` failed with the Python path returning a scalar for `np.int64(10)` and raising `TypeError` for `True`.
- [x] GREEN: same focused command passed: `2 passed in 6.43s`.
- [x] `pytest quantecon/random/tests/test_utilities.py -q` — `11 passed in 4.45s`.
- [x] `pytest quantecon/random -q` — `11 passed in 4.54s`.
- [x] `pytest quantecon -q` — `602 passed, 2 warnings in 504.77s`.
- [x] `flake8 --select=F401,F405,E231 quantecon`.
- [x] `git diff --check`.

### DIFF
--- a/quantecon/random/tests/test_utilities.py
+++ b/quantecon/random/tests/test_utilities.py
@@ -90,6 +90,17 @@ class TestDraw:
             out = func(self.cdf, size)
             assert_(out.shape == (size,))
 
+    def test_numpy_integer_size_returns_array(self):
+        size = np.int64(10)
+        for func in self.draw_funcs:
+            out = func(self.cdf, size)
+            assert_(out.shape == (size,))
+
+    def test_bool_size_is_treated_as_scalar(self):
+        for func in self.draw_funcs:
+            out = func(self.cdf, True)
+            assert_(isinstance(out, numbers.Integral))
+
     def test_return_values(self):
         for func in self.draw_funcs:
             out = func(self.cdf)
--- a/quantecon/random/tests/test_utilities.py
+++ b/quantecon/random/tests/test_utilities.py
@@ -125,6 +125,17 @@ class TestDraw:
             out = func(self.cdf, size)
             assert_(out.shape == (size,))
 
+    def test_numpy_integer_size_returns_array(self):
+        size = np.int64(10)
+        for func in self.draw_funcs:
+            out = func(self.cdf, size)
+            assert_(out.shape == (size,))
+
+    def test_bool_size_is_treated_as_scalar(self):
+        for func in self.draw_funcs:
+            out = func(self.cdf, True)
+            assert_(isinstance(out, numbers.Integral))
+
     def test_return_values(self):
         for func in self.draw_funcs:
             out = func(self.cdf)
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -246,7 +246,11 @@ def draw(cdf, size=None, rng=None):
     """
     if rng is None:
         rng = np.random
-    if isinstance(size, int):
+    integer_size = (
+        isinstance(size, (int, np.integer))
+        and not isinstance(size, (bool, np.bool_))
+    )
+    if integer_size:
         rs = rng.random(size)
         out = np.searchsorted(cdf, rs, side='right')
         return out
@@ -287,8 +291,9 @@ def _is_no_rng(numba_type):
 # holds the two paths together.
 @overload(draw)
 def ol_draw(cdf, size=None, rng=None):
+    is_integer_size = isinstance(size, types.Integer) and not isinstance(size, types.Boolean)
     if isinstance(rng, types.NumPyRandomGeneratorType):
-        if isinstance(size, types.Integer):
+        if is_integer_size:
             def draw_impl(cdf, size=None, rng=None):
                 rs = rng.random(size)
                 out = np.empty(size, dtype=np.int_)
@@ -300,7 +305,7 @@ def ol_draw(cdf, size=None, rng=None):
                 r = rng.random()
                 return np.searchsorted(cdf, r, side='right')
     elif _is_no_rng(rng):
-        if isinstance(size, types.Integer):
+        if is_integer_size:
             def draw_impl(cdf, size=None, rng=None):
                 rs = np.random.random(size)
                 out = np.empty(size, dtype=np.int_)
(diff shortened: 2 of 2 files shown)


kocaemre:
Rebased this branch onto current `main` and resolved the conflict in `quantecon/random/utilities.py` against the newer `draw(..., rng=...)` implementation.

The preserved behavior is:
- NumPy integer sizes such as `np.int64(10)` take the sized array branch.
- Boolean sizes still take the scalar branch.
- The newer `rng` path is preserved, so sized draws use `rng.random(size)` when a generator is provided.

Local verification after the rebase:

```text
pytest -q quantecon/random/tests/test_utilities.py
22 passed in 6.10s

flake8 --select=F401,F405,E231 quantecon/random/utilities.py quantecon/random/tests/test_utilities.py
passed

git diff --check
passed
```

New head: `3c9f13f`.

mmcky:
Thanks @kocaemre — closing this one, but the fix is landing: I've opened #955 with the reduced form of it.

Two things came out of reviewing this that are worth passing back. The @overload edits turn out to be no-ops: types.Boolean is not a subclass of types.Integer, so isinstance(size, types.Integer) and not isinstance(size, types.Boolean) can never differ from isinstance(size, types.Integer) alone. The jitted path was already dispatching every case correctly — only the pure-Python body was wrong. Similarly, the np.bool_ half of the exclusion in the Python predicate cannot be reached, because np.bool_ fails isinstance(size, (int, np.integer)) first. So the change reduces to the single line #918 proposed, and #955 is that line plus your np.int64 test, which was the genuinely useful part of this PR.

A suggestion for future contributions here: when an issue already specifies the fix — #918 gave the exact predicate, the dispatch table and the reasoning — the most valuable PR is the smallest one that implements it, plus a test. Extra defensive clauses cost review time to prove inert, which is time we would rather spend on the next issue.

Thanks for the work on this, including the rebase onto the rng= path after #917 landed.

PR #924, opened 2026-08-10
FIX: Point README star history chart to a working mirror

Mubelotix:
The star history chart in the README is currently broken, as the upstream project can no longer read the GitHub stargazer API. This updates the badge and link to star-history.dera.page, a mirror that uses a token-free data source, so the chart shows correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,4 +89,4 @@ pages = {5585}
 
 ## Star History
 
-[![Star History Chart](https://api.star-history.com/svg?repos=quantecon/quantecon.py&type=Date)](https://star-history.com/#quantecon/quantecon.py&Date)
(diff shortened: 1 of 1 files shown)


mmcky:
Thanks @Mubelotix — closing this one, because the premise does not hold up.

The chart is not currently broken. Fetching the URL that is in the README today returns HTTP 200 and a valid 65 KB SVG containing the real series — axes running 2014 to 2026 and 0.5K to 2K stars, labelled `quantecon/quantecon.py`.

Given that, we would rather not point the README at a third-party mirror. A badge in the README is fetched by everyone who opens the project page, so the host is worth being conservative about, and `star-history.dera.page` is not a domain we know anything about.

If you are seeing it fail, it would be genuinely useful to know where and when — an outage or a region-specific failure is worth knowing about, and if the official host does become unreliable we would rather drop the chart than mirror it. Thanks for flagging it.

Issue #925, opened 2026-08-13
WASM: browser support for QuantEcon.py (Numba + JupyterLite) — tracking issue


mmcky:
> [!NOTE]
> **Updated 2026-08-31.** Revised against live state after a ten-day gap; the revision log is in the comments. Since 2026-08-21: PR #942 was closed unmerged; PR #943 was rebased onto `main` (int64 commit dropped, no longer touches `comb_jit`); a Numba gufunc-cache segfault (numba/numba#10128, fix numba/numba#10545 still open) constrains how #930 can be implemented and a second implementation exists on the `wasm-930-dynamic-gufuncs` branch; PR #937 remains approved and unmerged; nothing has moved upstream on emscripten-forge/recipes#6309 and the jitted-generator hang (#927) is still unreported. The 2026-08-21 revision note and the original text are in the edit history.
## Background

On 11 August 2026 QuantStack announced [Numba in the Browser](https://notebook.link/blog/numba-in-the-browser/) (Anutosh Bhat): a genuine Numba JIT running entirely in the browser. Python functions are compiled through llvmlite to WebAssembly, linked in-process with LLD, and loaded as Emscripten side modules into a JupyterLite kernel — no server anywhere. Numba is now published on [emscripten-forge](https://github.com/emscripten-forge/recipes) (numba 0.67.0 build 1 + 10 emscripten patches; llvmlite 0.49.0 + 4 patches), and the announcement showcases the economics stack built on top of it (EconForge's interpolation.py and Dolo.py, PyMC/PyTensor). QuantEcon.py is the obvious next citizen of that ecosystem, and its authors explicitly invite the wider Numba ecosystem to validate their packages.

Concretely, "browser support" for QuantEcon.py means: **works in JupyterLite with the xeus-python kernel against the emscripten-forge + conda-forge channels.** This is not the classic Pyodide kernel — Numba remains unavailable there (see numba/numba#3284, pyodide/pyodide-recipes#192).

## Where we stand (verified 2026-08-31 14:53 AEST)

| Fact | Value | Measured |
|---|---|---|
| `main` | 481947d (PR #938); no commit since 2026-08-21 | `git log origin/main`, 2026-08-31 |
| Sub-issues | 0 of 10 closed (#926–#934, #944); milestone 8: 11 open, 1 closed (PR #938) | GitHub, 2026-08-31 |
[...]

PR #935, opened 2026-08-13
WASM: Disable Gini parallelization on Emscripten

### LINKED ISSUE #926: WASM: remove @njit(parallel=True) from gini_coefficient via an O(n log n) rewrite
> [!NOTE]
> **Updated 2026-08-21.** When this was written, `@njit(parallel=True)` failed at first call in the browser; since 2026-08-18 the emscripten-forge Numba build carries patch 0010 (emscripten-forge/recipes#6293), which compiles it on the serial pipeline, so `gini_coefficient` no longer crashes in JupyterLite but runs the serial O(n²) loop. The rewrite remains the intended change, now for performance and for independence from a downstream patch that upstream describes as temporary; PR #937 is approved and awaiting merge, the platform-gate PR #935 is closed, and acceptance criterion 4 now points at #928 without blocking closure. The original text is preserved in the edit history.

Part of #925 (Phase 1). The browser-compliance audit identified this as a hard blocker; it is now a performance problem (see Problem below), and the campaign's open blockers are tracked in #927 (jitted generators) and #944 (`cache=True` chains, upstream emscripten-forge/recipes#6309).

**Status.** Implemented in PR #937 (approved by @kp992 2026-08-14, CI green, awaiting merge), which satisfies criteria 1–3 and will close this issue automatically on merge. It supersedes the platform-gate approach in #935, closed 2026-08-14 after the rewrite was adopted.

## Problem

[`gini_coefficient`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/_inequality.py#L59-L85) is the library's only use of Numba's ParallelAccelerator pass:

```python
@njit(parallel=True)
def gini_coefficient(y):
    ...
    for i in prange(n):
```

The in-browser Numba runtime (emscripten-forge) is single-threaded: OpenMP and TBB are disabled at build time. Until 2026-08-18 its only serial fallback was patch 0007, which covers the parallel *ufunc* target (`guvectorize(..., target='parallel')`) and not `@njit(parallel=True)`, so the call failed at first use. Compilation is lazy, so `import quantecon` was unaffected — the failure surfaced the first time a user called `qe.gini_coefficient(...)` in a JupyterLite notebook.

That is no longer the failure mode. Patch 0010 (emscripten-forge/recipes#6293, merged 2026-08-18, shipped as numba 0.67.0 build 1 on emscripten-forge-4x) rewrites `options['parallel'] = False` inside `numba.core.decorators.jit` on `sys.platform == "emscripten"` and forces `NUMBA_NUM_THREADS=1`, with a recipe test for exactly the `@njit(parallel=True)` + `prange` pattern. On that build the current code compiles and runs in the browser — but on a serial O(n²) path. The benchmarks in the #935 closing comment put that at 9.4 s natively for n = 100,000, roughly 20–30 s in WASM. Deployments built before 2026-08-18 (the notebook.link demo in #927 was on numba 0.66.0) still hit the original crash.

The rewrite is therefore still wanted, on two counts: it removes the serial O(n²) cliff, which matters most in the single-threaded browser, and it removes the library's dependence on a downstream patch whose own comment says it can be removed once the Emscripten stack supports pthreads.

## Upstream

Patch 0010 is the outcome of @anutosh491's 2026-08-17 offer on #925 ("Is quantecon dependent on `@njit(parallel=True)` somehow? I think I might be able to get that working too!"). See emscripten-forge/recipes#6293; the numba recipe now carries ten patches (0001–0010).

## Fix

Replace the O(n²) pairwise double loop with the O(n log n) closed form on sorted data, which removes the need for the parallel loop altogether. For `y` sorted ascending with 1-indexed `i`, the Gini coefficient is `2·Σ(i·y_i) / (n·Σy) − (n+1)/n`:

```python
@njit
def gini_coefficient(y):
    n = len(y)
    y_s = np.sort(y)
    i_weighted = 0.0
    total = 0.0
    for i in range(n):
        i_weighted += (i + 1) * y_s[i]
        total += y_s[i]
    return (2 * i_weighted) / (n * total) - (n + 1) / n
```

This was originally listed here as an optional follow-up, with a `sys.platform != "emscripten"` gate on `parallel=` as the primary fix. Benchmarking in #702 showed the rewrite is the better option on both counts, so the two have been swapped. It removes the Emscripten incompatibility at the source rather than gating around it, and it avoids the serial-O(n²) performance cliff that a platform gate would leave in place for browser users — which matters precisely because the browser is the single-threaded environment. Patch 0010 now applies that same gate upstream for every `@njit(parallel=True)` function, so the gate's performance drawback is exactly what browser users get today.

Measured against the current implementation (numba 0.62.1, numpy 2.3.5, 10 cores, best-of-5 after JIT warm-up, Pareto(3) inputs):

| n | current (10-core parallel) | rewrite (single-threaded) | speedup |
|---|---|---|---|

The speedup grows with n because this is an algorithmic change, not a constant-factor one, and the single-threaded rewrite beats the 10-core parallel implementation at every size tested. Agreement with the current implementation is within 1e-9 on random Pareto data across all five sizes.

Note that `prange` and the `parallel=True` flag are both dropped. `gini_coefficient` is the module's only `prange` user, so the import on line 7 of `_inequality.py` should be narrowed to `from numba import njit`.

## Smoke suite

`ci/wasm/smoke_test.py` (on `main` via #938) currently marks `test_gini_coefficient` as `xfail(IS_EMSCRIPTEN, strict=True)`, and `ci/wasm/test_jupyterlite.py::test_gini_fails_on_emscripten` asserts that the call raises. Both encode the pre-patch-0010 failure mode: against numba 0.67.0 build 1 the first will XPASS (and fail, being strict) and the second will fail its assertion, whether or not #937 has merged. They must flip to a pass expectation before the first browser run — tracked in #933. Once a release containing #937 is installed in the browser, a useful check is that the rewrite is what executes — timing a large input against the serial-fallback figures in the #935 closing comment — since a bare "runs without error" no longer distinguishes the rewrite from the serial fallback.

## Acceptance criteria

- [ ] `gini_coefficient` no longer uses `@njit(parallel=True)` or `prange`
- [ ] Existing `test_gini_coeff` assertions pass unchanged (Pareto and Weibull, `rtol=1e-01`)
- [ ] A regression test pins agreement with the pairwise definition on a small fixed array
- [ ] Browser confirmation is tracked in #928 (its `gini_coefficient` checklist item) and is not a blocker for closing this issue; it can only be run once a release containing the rewrite is installable in the browser

russlan23:
Closes #926

## Summary

- disable Numba's parallel accelerator for `gini_coefficient` on Emscripten;
- keep the existing parallel path on native platforms.

The browser Numba runtime is single-threaded and cannot compile this library's only `@njit(parallel=True)` function. Selecting the decorator option from `sys.platform` lets `prange` use its serial semantics in JupyterLite without changing the public API or native behavior.

## Validation

- `python -m pytest -q quantecon/tests/test_inequality.py` — 4 passed
- `python -m flake8 --select=F401,F405,E231 quantecon` — passed
- Native Python 3.13: compiled target reports `parallel=True`
- Simulated Emscripten import: compiled target reports `parallel=False`; `gini_coefficient([1, 2, 3]) == 2/9`
- `git diff --check` — passed

Automated assistance was used to prepare and validate this change.

### DIFF
--- a/quantecon/_inequality.py
+++ b/quantecon/_inequality.py
@@ -3,8 +3,13 @@ Implements inequality and segregation measures such as Gini, Lorenz Curve
 
 """
 
+import sys
+
 import numpy as np
 from numba import njit, prange
+
+
+PARALLELIZE = sys.platform != "emscripten"
 
 
 @njit
@@ -54,7 +59,7 @@ def lorenz_curve(y):
     return cum_people, cum_income
 
 
-@njit(parallel=True)
+@njit(parallel=PARALLELIZE)
 def gini_coefficient(y):
     r"""
     Implements the Gini inequality index
@@ -152,4 +157,3 @@ def rank_size(data, c=1.0):
     rank_data = np.arange(len(w)) + 1
     size_data = w
     return rank_data, size_data
-


mmcky:
Thanks for this @russlan23, and welcome — this is a clean, well-validated first contribution. To be clear up front: the change does exactly what #926 asked for at the time you wrote it, right down to the diff the issue prescribed, and the validation you did (native vs simulated-Emscripten compiled target, plus the `2/9` check) is more thorough than most. The reason I'm closing it is that we changed the approach in the issue yesterday, after you'd already opened this. That's on our side, not yours.

The background is #702, an old PR to optimise `gini_coefficient` that I benchmarked while deciding whether to close it. That work showed the O(n²) pairwise double loop can be replaced with an O(n log n) closed form on sorted data, which removes `parallel=True` altogether rather than gating it. #926 has been retitled and rewritten to make that the primary fix, so the platform gate here would become dead code as soon as the rewrite lands.

The reason I don't want to merge this as an interim step is that the gate leaves browser users on a serial O(n²) path, and that's the environment where it hurts most. Measured natively (numba 0.62.1, numpy 2.3.5, Pareto(3) inputs), comparing what the gate compiles to on Emscripten against the rewrite:

| n | gated serial O(n²) | rewrite O(n log n) | ratio |
|---|---|---|---|
| 1,000 | 0.8 ms | 0.011 ms | 77x |
| 10,000 | 92.9 ms | 0.521 ms | 178x |
| 50,000 | 2346 ms | 3.13 ms | 750x |
| 100,000 | 9364 ms | 6.91 ms | 1354x |

Those serial figures are native timings; WASM typically runs a further 2–3x slower, so a 100,000-element Gini in JupyterLite would be a 20–30 second call. It wouldn't crash, which is what this PR set out to fix, but it wouldn't really be usable either. The rewrite gets it to single-digit milliseconds and needs no platform branch at all.

If you'd like to take the rewrite, it would be very welcome — it's the same file and the same issue, #926 now carries the formula and acceptance criteria, and you clearly have the setup for validating it. No obligation at all if you'd rather pick up something else; there are other open items under the #925 umbrella. Either way, thanks for jumping on this one so quickly.

Issue #928, opened 2026-08-13
WASM: JupyterLite proof-of-concept deployment and browser smoke suite


mmcky:
> [!NOTE]
> **Updated 2026-08-21.** Phase 0 was overtaken by ad-hoc browser runs: the `_LAPACK` question (#927) was answered on the public numba-ecosystem demo, the `intp` question (#929) was settled in the #942 review, and PR #938 (merged as 481947d) has landed the deployment config and smoke suite under `ci/wasm/`. The checklist below is reworded to match those findings and the upstream Numba change that removed the expected `gini_coefficient` failure (patch 0010, emscripten-forge/recipes#6293); items with recorded browser evidence are ticked and a results table is maintained in the comments. The original text is preserved in the edit history.

Part of #925 (Phase 0). This issue was meant to gate #927 and calibrate #929 and #930; #927 was resolved by ad-hoc browser runs and #929 by the #942 review instead (see the checklist), so its remaining gating value is the cold/warm import timing for #930 and the first full-suite browser run via #933.

## Goal

Prove the stack end-to-end with zero code changes, and convert every "should work" in the audit into a known. Because quantecon is `noarch: python` on conda-forge, there is a real chance it installs and imports in JupyterLite **today** with no packaging work.

## Setup

Build a minimal JupyterLite deployment with [jupyterlite-xeus](https://jupyterlite-xeus.readthedocs.io/en/latest/environment.html):

```yaml
name: qe-lite
channels:
  - https://prefix.dev/emscripten-forge-4x
  - https://prefix.dev/conda-forge
dependencies:
  - xeus-python
  - numba
  - quantecon
```

This file is now committed as [`ci/wasm/environment.yml`](https://github.com/QuantEcon/QuantEcon.py/blob/481947de0ff51a3d60a4be3ee3ad181d92c0a0cd/ci/wasm/environment.yml) (PR #938). `numba` is unpinned, so a fresh build resolves the current emscripten-forge-4x build — numba 0.67.0 build 1 as of 2026-08-18, which carries patch 0010 (emscripten-forge/recipes#6293) forcing `@njit(parallel=True)` to the serial pipeline on Emscripten. Note that the browser results recorded so far did **not** come from this deployment: they came from the public numba-ecosystem demo (Numba 0.66.0) and the #942 review, so the versions per result are recorded in the table.

## Smoke checklist

One representative function per Numba feature class used in the library:

[...]

PR #937, opened 2026-08-14
Rewrite Gini coefficient in O(n log n)

### LINKED ISSUE #926: WASM: remove @njit(parallel=True) from gini_coefficient via an O(n log n) rewrite

russlan23:
Replaces the quadratic pairwise implementation with the sorted closed-form calculation requested in #926. This removes the module's only `prange` use and the `parallel=True` compilation path that blocks the in-browser runtime.

The existing Pareto and Weibull assertions remain unchanged. I added a fixed-array regression against the pairwise definition.

Validation:
- `pytest -q quantecon/tests/test_inequality.py` — 4 passed
- `flake8 --select=F401,F405,E231 quantecon`
- Independent agreement checks against the pairwise definition for n = 4, 10, 100, and 1,000

The separate Phase 0 browser deployment in #928 is still open, so that project-level browser check remains outstanding.

Closes #926.

AI assistance: Used for implementation support and test execution; I reviewed the diff and validation results.

### DIFF
--- a/quantecon/_inequality.py
+++ b/quantecon/_inequality.py
@@ -4,7 +4,7 @@ Implements inequality and segregation measures such as Gini, Lorenz Curve
 """
 
 import numpy as np
-from numba import njit, prange
+from numba import njit
 
 
 @njit
@@ -54,7 +54,7 @@ def lorenz_curve(y):
     return cum_people, cum_income
 
 
-@njit(parallel=True)
+@njit
 def gini_coefficient(y):
     r"""
     Implements the Gini inequality index
@@ -76,11 +76,13 @@ def gini_coefficient(y):
     https://en.wikipedia.org/wiki/Gini_coefficient
     """
     n = len(y)
-    i_sum = np.zeros(n)
-    for i in prange(n):
-        for j in range(n):
-            i_sum[i] += abs(y[i] - y[j])
-    return np.sum(i_sum) / (2 * n * np.sum(y))
+    y_s = np.sort(y)
+    i_weighted = 0.0
+    total = 0.0
+    for i in range(n):
+        i_weighted += (i + 1) * y_s[i]
+        total += y_s[i]
+    return (2 * i_weighted) / (n * total) - (n + 1) / n
 
 
 def shorrocks_index(A):
@@ -152,4 +154,3 @@ def rank_size(data, c=1.0):
     rank_data = np.arange(len(w)) + 1
     size_data = w
     return rank_data, size_data
-
--- a/quantecon/tests/test_inequality.py
+++ b/quantecon/tests/test_inequality.py
@@ -59,6 +59,11 @@ def test_gini_coeff():
     coeff = gini_coefficient(y)
     assert_allclose(expected, coeff, rtol=1e-01)
 
+    # Regression against the pairwise definition
+    y = np.array([1.0, 2.0, 4.0, 8.0])
+    expected = np.abs(y[:, None] - y).sum() / (2 * len(y) * y.sum())
+    assert_allclose(expected, gini_coefficient(y))
+
     # Tests Weibull: G = 1 - 2**(-1/a)
     a = np.random.randint(2, 15)
     expected = 1 - 2 ** (-1 / a)
@@ -130,4 +135,3 @@ def test_rank_size():
     r_sqval_exp = r_value_exp**2
 
     assert_raises(AssertionError, assert_allclose, r_sqval_exp, 1, rtol=1e-3)
-


mmcky:
thanks for your PR @russlan23 

@kp992 can you please review this as part of the WASM project

kp992 (approved):

mmcky:
Thanks @russlan23 for your contribution and AI disclosure (greatly appreciated).

Verified the new closed form against the old pairwise implementation at the current head (worst absolute difference 1.2e-15 across Pareto, lognormal, all-zero, all-equal, mixed-sign and integer inputs), CI is green across the 3.12/3.13/3.14 matrix — marking ready.

Issue #933, opened 2026-08-13
WASM: add a WebAssembly/JupyterLite smoke-test job to CI


mmcky:
> [!NOTE]
> **Updated 2026-08-21.** The seed is now on `main` via #938 (`ci/wasm/environment.yml`, `ci/wasm/smoke_test.py`, the Playwright harness `ci/wasm/test_jupyterlite.py`, and the native gate in `ci_wasm_smoke.yml`), so the Approach below names it instead of "the config from #928", and a task list for the runner PR has been added. What that list corrects: the parked job (2026-08-19 comment) still needs triggers, the gini expectation in both suite files is inverted now that numba 0.67.0 build 1 serialises `parallel=True` (patch 0010, emscripten-forge/recipes#6293), the two jitted-generator tests will hang without a real guard, `test_simplex_grid` fails on a warm cache (#944), and `smoke_test.py` is not yet shipped into the kernel. The original text is preserved in the edit history.

Part of #925 (Phase 3 — keep it working).

## Goal

Browser support must not regress once established: add a `wasm` job to GitHub Actions that runs a smoke suite in the real browser stack.

## Approach

- Seed: PR #938 (merged 2026-08-21) — `ci/wasm/environment.yml` (xeus-python kernel env, emscripten-forge-4x + conda-forge channels), the sentinel-driven Playwright harness `ci/wasm/test_jupyterlite.py`, and the native gate job `native` in `.github/workflows/ci_wasm_smoke.yml`, which runs `ci/wasm/smoke_test.py` against repo source on every PR. The Emscripten runner job is parked in the 2026-08-19 comment below.
- Execute in the xeus-python kernel via headless Chromium + Playwright (the route #938 chose). `pytester`, emscripten-forge's pyjs-based recipe-test harness, remains the alternative only if quantecon is later packaged on emscripten-forge (#932).
- pandas and pytest are both packaged for WASM, so meaningful portions of the real test suite can eventually run in-browser — start with the smoke list from #928 (`smoke_test.py`) and grow from there.
- Track the emscripten-forge numba/llvmlite versions in the job the way conda-forge versions are tracked today; `environment.yml` pins bare `numba`, so have the job log the resolved build (only 0.67.0 build 1 carries patch 0010, see task 2).

[...]

Issue #929, opened 2026-08-13
WASM: audit 32-bit intp behaviour on wasm32 (overflow guards, simplex_index wrapping, dtypes)


mmcky:
> [!NOTE]
> **Updated 2026-08-21 (second edit).** Tracker reconciliation: PR #942 is still open with a live "Fixes #929" and is to be closed unmerged; PR #943 is stacked on it and carries the same `int64` widening, so it must be rebased onto `main` keeping `intp` before merge. The earlier note and the #944 summary below now attribute the browser failure to the cache-hit/cache-miss linking bug in the emscripten-forge Numba build (emscripten-forge/recipes#6309), not to `comb_jit`'s eager signature, and the grep-sweep task names `k_array_rank_jit` as sharing the `intp` boundary. Labels changed from `tests` to `documentation`.

> [!NOTE]
> **Updated 2026-08-21.** The original body described three problems; review of #942 showed two of them were wrong and the third is a documentation task, not a code change. The `comb_jit` / `simplex_grid` use of `np.intp` is **intentional** and should be kept — see below. The genuine WASM blocker surfaced by #942 (`cache=True` callers failing to link cache-restored callees on the emscripten-forge Numba build) is tracked separately in #944. The original text is preserved in the edit history.

Part of #925 (Phase 1). Verification runs on the Phase 0 deployment (#928).

## Background

Emscripten/wasm32 is a 32-bit platform: `np.intp` and NumPy's default integer are int32, so the library behaves as it would on 32-bit Linux. This issue audits what that means for QuantEcon.py.

## Findings

[...]

Issue #927, opened 2026-08-13
WASM: support_enumeration and vertex_enumeration hang in the browser — root cause is jitted generators (numba.np.linalg._LAPACK verified working)


mmcky:
> [!NOTE]
> **Updated 2026-08-21.** Second revision today, after the cross-issue consistency pass: `vertex_enumeration` joins the scope — [`_vertex_enumeration_gen`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/vertex_enumeration.py#L117-L118) is the package's only other `@jit(nopython=True)` generator, so it is expected to hang for the same reason (by mechanism; not yet observed in the browser) — and the status note, the Problem section, the acceptance criteria and a new Related section are updated accordingly, now that PR #938 has merged and `ci/wasm/smoke_test.py` exists on main. The former "Side benefit" section is removed because its premise was wrong: `_indiff_mixed_action`, which makes the `_LAPACK` call, is already [`cache=True`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/support_enumeration.py#L145), and the [`# cache=True raises _pickle.PicklingError`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/support_enumeration.py#L95) comment on the generator is a 2016 leftover unrelated to `_LAPACK` (`cache=True` on the generator works natively on Numba 0.62.1; re-enabling it is a separate maintenance item). Two permalinks and the patch count are corrected in place, and a draft of the upstream report for criterion 3 is in the comment below.

[...]

PR #938, opened 2026-08-15
WASM: JupyterLite environment config and browser smoke suite

kp992:
Part of #928 and #925 Phase 0.

Adds under `ci/wasm/`:

- `environment.yml` — xeus-python kernel config for a minimal JupyterLite deployment (emscripten-forge + conda-forge channels). Seed for the CI job in #933.
- `smoke_test.py` — pytest suite covering every Numba feature class in the library: lazy `@njit`, NRT-allocated arrays, parallel `guvectorize` (cpu fallback), eager `guvectorize`, `nelder_mead`/`brent_max`/`newton`, `lemke_howson`, `vertex_enumeration` (`numba.typed.Dict`), bare `np.linalg.solve` in `@njit` (isolates the `_LAPACK` mechanism from #927), `support_enumeration` (end-to-end #927 test), `gini_coefficient` (marked `xfail` on Emscripten per #926), `simplex_grid` (feeds #929), and the `objmode` shim in `searchsorted`.
- `test_jupyterlite.py` — Playwright harness that drives the suite's key items inside a locally-built JupyterLite site. Manual for now; the CI job that builds the site and runs it lands with #933.

Plus a native CI gate (`ci_wasm_smoke.yml`) that runs the suite against this repo's source on every PR, and `testpaths = quantecon` in `pytest.ini` so bare `pytest` keeps collecting only the package tests.

All 16 tests pass natively (Python 3.11, Numba 0.61).

---

**Maintainer edit (2026-08-19):** scoped per review — the Emscripten/JupyterLite runner job moved out to a follow-up PR under #933 (it exercises the released conda-forge package rather than this PR's source, and needs its own red/green iteration; the salvaged job with review fixes applied is parked in #933). This PR now references #928 as Part-of, since the results table there can only be filled after the first successful WASM run.

### DIFF
--- a/.github/workflows/ci_wasm_smoke.yml
+++ b/.github/workflows/ci_wasm_smoke.yml
@@ -10,12 +10,14 @@ on:
 
 jobs:
   # ------------------------------------------------------------------
-  # Job 1 — native Python.  Fast gate: ensures the suite itself is
-  # correct and all 16 tests pass before the slow WASM build starts.
+  # Native Python gate: ensures the smoke suite itself is correct and
+  # all tests pass against this repo's source.  The Emscripten /
+  # JupyterLite job that consumes the same suite lands with issue #933.
   # ------------------------------------------------------------------
   native:
     name: Smoke suite (native)
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
     - uses: actions/checkout@v7
 
@@ -44,59 +46,3 @@ jobs:
     - name: Run smoke suite (native)
       shell: bash -l {0}
       run: pytest ci/wasm/smoke_test.py -v
-
-  # ------------------------------------------------------------------
-  # Job 2 — Emscripten / JupyterLite.
-  # Builds a real JupyterLite site from emscripten-forge packages
-  # (xeus-python WASM kernel + numba + quantecon), then drives it
-  # headlessly with Playwright to verify the stack end-to-end.
-  # ------------------------------------------------------------------
-  wasm:
-    name: Smoke suite (Emscripten / JupyterLite)
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v7
-
-    - uses: actions/setup-python@v5
-      with:
-        python-version: "3.11"
-
-    - name: Cache JupyterLite build output
-      uses: actions/cache@v6
-      with:
-        path: _site
-        key: jupyterlite-site-${{ hashFiles('ci/wasm/environment.yml') }}
-
-    - name: Install micromamba
-      uses: mamba-org/setup-micromamba@v2
-      with:
-        micromamba-binary-path: /usr/local/bin/micromamba
-        init-shell: none
-        generate-run-shell: false
-
-    - name: Install JupyterLite build tools
-      run: |
-        pip install \
-          "jupyterlite-core>=0.4" \
-          "jupyterlite-xeus>=4.5" \
-          pytest \
-          pytest-playwright
-
-    - name: Install Playwright browser
-      run: python -m playwright install --with-deps chromium
-
-    - name: Build JupyterLite site
-      run: |
-        jupyter lite build \
-          --XeusAddon.environment_file=ci/wasm/environment.yml \
-          --output-dir=_site
-
-    - name: Start local server
-      run: python -m http.server 8000 --directory _site &
-
-    - name: Run Playwright smoke tests (Emscripten kernel)
-      run: |
-        pytest ci/wasm/test_jupyterlite.py \
-          --browser chromium \
-          --timeout=660 \
-          -v
--- a/.github/workflows/ci_wasm_smoke.yml
+++ b/.github/workflows/ci_wasm_smoke.yml
@@ -1,0 +1,38 @@
+name: WASM smoke suite (native)
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  smoke:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v7
+      with:
+        fetch-depth: 0
+    - name: Cache conda
+      uses: actions/cache@v6
+      env:
+        CACHE_NUMBER: 0
+      with:
+        path: ~/conda_pkgs_dir
+        key:
+          ${{ runner.os }}-3.13-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
+    - uses: conda-incubator/setup-miniconda@v4
+      with:
+        auto-update-conda: true
+        miniforge-version: latest
+        environment-file: environment.yml
+        python-version: "3.13"
+        auto-activate-base: false
+        use-only-tar-bz2: true
+        activate-environment: qe
+    - name: Run smoke suite
+      shell: bash -l {0}
+      run: |
+        pytest ci/wasm/smoke_test.py -v
--- a/.github/workflows/ci_wasm_smoke.yml
+++ b/.github/workflows/ci_wasm_smoke.yml
@@ -1,4 +1,4 @@
-name: WASM smoke suite (native)
+name: WASM smoke suite
 
 on:
   push:
@@ -9,20 +9,24 @@ on:
       - main
 
 jobs:
-  smoke:
+  # ------------------------------------------------------------------
+  # Job 1 — native Python.  Fast gate: ensures the suite itself is
+  # correct and all 16 tests pass before the slow WASM build starts.
+  # ------------------------------------------------------------------
+  native:
+    name: Smoke suite (native)
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v7
-      with:
-        fetch-depth: 0
+
     - name: Cache conda
       uses: actions/cache@v6
       env:
         CACHE_NUMBER: 0
       with:
         path: ~/conda_pkgs_dir
-        key:
-          ${{ runner.os }}-3.13-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
+        key: ${{ runner.os }}-3.13-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
+
     - uses: conda-incubator/setup-miniconda@v4
       with:
         auto-update-conda: true
@@ -32,7 +36,56 @@ jobs:
         auto-activate-base: false
         use-only-tar-bz2: true
         activate-environment: qe
-    - name: Run smoke suite
+
+    - name: Run smoke suite (native)
       shell: bash -l {0}
+      run: pytest ci/wasm/smoke_test.py -v
+
+  # ------------------------------------------------------------------
+  # Job 2 — Emscripten / JupyterLite.
+  # Builds a real JupyterLite site from emscripten-forge packages
+  # (xeus-python WASM kernel + numba + quantecon), then drives it
+  # headlessly with Playwright to verify the stack end-to-end.
+  # ------------------------------------------------------------------
+  wasm:
+    name: Smoke suite (Emscripten / JupyterLite)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v7
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.11"
+
+    - name: Cache JupyterLite build output
+      uses: actions/cache@v6
+      with:
+        path: _site
+        key: jupyterlite-site-${{ hashFiles('ci/wasm/environment.yml') }}
+
+    - name: Install JupyterLite build tools
       run: |
-        pytest ci/wasm/smoke_test.py -v
+        pip install \
+          "jupyterlite-core>=0.4" \
+          "jupyterlite-xeus>=4.5" \
+          pytest \
+          pytest-playwright
+
+    - name: Install Playwright browser
+      run: python -m playwright install --with-deps chromium
+
+    - name: Build JupyterLite site
+      run: |
+        jupyter lite build \
+          --XeusAddon.environment_file=ci/wasm/environment.yml \
+          --output-dir=_site
+
+    - name: Start local server
+      run: python -m http.server 8000 --directory _site &
+
+    - name: Run Playwright smoke tests (Emscripten kernel)
+      run: |
+        pytest ci/wasm/test_jupyterlite.py \
+          --browser chromium \
+          --timeout=660 \
+          -v
--- a/.github/workflows/ci_wasm_smoke.yml
+++ b/.github/workflows/ci_wasm_smoke.yml
@@ -37,6 +37,10 @@ jobs:
         use-only-tar-bz2: true
         activate-environment: qe
 
+    - name: Install quantecon from source
+      shell: bash -l {0}
+      run: pip install -e . --no-deps
+
     - name: Run smoke suite (native)
       shell: bash -l {0}
       run: pytest ci/wasm/smoke_test.py -v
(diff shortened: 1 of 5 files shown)


.github/workflows/ci_wasm_smoke.yml
```diff
@@ -0,0 +1,102 @@
...
+      shell: bash -l {0}
+      run: pytest ci/wasm/smoke_test.py -v
+
+  # ------------------------------------------------------------------
+  # Job 2 — Emscripten / JupyterLite.
+  # Builds a real JupyterLite site from emscripten-forge packages
+  # (xeus-python WASM kernel + numba + quantecon), then drives it
+  # headlessly with Playwright to verify the stack end-to-end.
+  # ------------------------------------------------------------------
+  wasm:
```
mmcky: The Job 1 comment says the native gate runs "before the slow WASM build starts", but without `needs: native` the two jobs run in parallel. Adding `needs: native` here makes the gating real.

.github/workflows/ci_wasm_smoke.yml
```diff
@@ -0,0 +1,102 @@
...
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.11"
+
+    - name: Cache JupyterLite build output
+      uses: actions/cache@v6
+      with:
+        path: _site
+        key: jupyterlite-site-${{ hashFiles('ci/wasm/environment.yml') }}
```
mmcky: Since `quantecon` is unpinned in the env file, a new release changes the built site without changing this key — and `jupyter lite build` runs unconditionally anyway, so a cache hit just restores a stale `_site` that then gets rebuilt over. I'd drop the cache for now; the build isn't the slow part of this job.

.github/workflows/ci_wasm_smoke.yml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Install JupyterLite build tools
+      run: |
+        pip install \
+          "jupyterlite-core>=0.4" \
+          "jupyterlite-xeus>=4.5" \
+          pytest \
+          pytest-playwright
+
+    - name: Install Playwright browser
+      run: python -m playwright install --with-deps chromium
```
mmcky: This step hung for the full 6-hour job limit on the last run — `--with-deps` shells out to `sudo apt-get`, a known hang point, and ubuntu-latest already ships Chromium's system deps, so plain `python -m playwright install chromium` should do. Regardless, worth adding `timeout-minutes: 45` to both jobs so a future hang can't burn 6 hours of runner time.

.github/workflows/ci_wasm_smoke.yml
```diff
@@ -0,0 +1,102 @@
...
+          --output-dir=_site
+
+    - name: Start local server
+      run: python -m http.server 8000 --directory _site &
+
+    - name: Run Playwright smoke tests (Emscripten kernel)
+      run: |
+        pytest ci/wasm/test_jupyterlite.py \
+          --browser chromium \
+          --timeout=660 \
```
mmcky: `--timeout` belongs to `pytest-timeout`, which isn't in the pip install list above — once the browser install is fixed, pytest will exit immediately with "unrecognized arguments". Add `pytest-timeout` to the install step; the flag itself is worth keeping given the hang history.

ci/wasm/environment.yml
```diff
@@ -0,0 +1,9 @@
+name: qe-lite
+channels:
+  - https://prefix.dev/emscripten-forge-4x
+  - https://prefix.dev/conda-forge
+dependencies:
+  - xeus-python
+  - numba
+  - quantecon
```
mmcky, line 8: quantecon isn't packaged on emscripten-forge-4x (checked the wasm32 repodata), so this resolves to the released noarch package from conda-forge — the wasm job tests the latest release, not this PR's source. Fine for Phase 0, but as a per-PR gate it can't catch regressions in PR code and can go red for unrelated upstream reasons — consider `workflow_dispatch` + `schedule` (or main-only) triggers until it's proven stable.

ci/wasm/environment.yml
```diff
@@ -0,0 +1,9 @@
...
+  - pytest
```
mmcky: Nothing ships `smoke_test.py` into the JupyterLite site or runs pytest inside the kernel, so this dependency — and the `IS_EMSCRIPTEN`/xfail branches in smoke_test.py — are currently unused. Either wire that up (e.g. `jupyter lite build --contents`) or drop this and soften the smoke_test.py docstring; the Playwright file covers the critical items in the meantime.

ci/wasm/test_jupyterlite.py
```diff
@@ -0,0 +1,162 @@
...
+
+Build the site first:
+    jupyter lite build --XeusAddon.environment_file=ci/wasm/environment.yml \\
+                       --output-dir=_site
+
+Then run:
+    pytest ci/wasm/test_jupyterlite.py --browser chromium -v
+"""
+import pytest
+from playwright.sync_api import Browser, Page
```
mmcky, line 21: This import breaks `ci_np2.yml`, which runs bare `pytest` from the repo root and now dies collecting this file (`ModuleNotFoundError: No module named 'playwright'`) — that's the red `tests` check on this PR. Cleanest fix: add `testpaths = quantecon` to pytest.ini, which also keeps bare `pytest` working locally.

ci/wasm/test_jupyterlite.py
```diff
@@ -0,0 +1,162 @@
...
+    # Wait for the kernel to reach idle (downloads + first JIT compile)
+    _wait_idle(page, BOOT_MS)
+
+
+def _wait_idle(page: Page, timeout: int) -> None:
+    """Block until the JupyterLab kernel status indicator reads 'Idle'."""
+    page.wait_for_function(
+        """() => {
+            const el = document.querySelector(
+                ".jp-Toolbar-kernelStatus, [data-status]");
```
mmcky: The bare `[data-status]` fallback can match an unrelated element that never reads "idle", turning every `_run` into a 10-minute timeout. Pin this to the console toolbar's kernel-status node.

ci/wasm/test_jupyterlite.py
```diff
@@ -0,0 +1,162 @@
...
+    Paste *code* into the active console prompt, execute it with Shift+Enter,
+    wait for the kernel to return to idle, and return the last output text.
+    """
+    prompt = page.locator(".jp-CodeConsole-input .cm-content, "
+                          ".jp-Console-promptCell .CodeMirror")
+    prompt.last.click()
+    page.keyboard.press("Control+a")
+    page.keyboard.type(code)
+    page.keyboard.press("Shift+Enter")
+    _wait_idle(page, timeout)
```
mmcky: Race: the status indicator may not have flipped to "busy" yet when this runs, so it can see the stale "idle" and return before the cell executes, capturing the previous cell's output. Wait for busy → idle, or wait for the `.jp-OutputArea-output` count to increase.

ci/wasm/test_jupyterlite.py
```diff
@@ -0,0 +1,162 @@
...
+def test_kernel_boots(console: Page):
+    """xeus-python WASM kernel loads and reaches idle without error."""
+    # If the fixture succeeds the kernel already booted; just assert no crash.
+    assert console.url.startswith(SITE)
+
+
+def test_import_quantecon(console: Page):
+    """import quantecon succeeds in the Emscripten environment."""
+    out = _run(console, "import quantecon as qe; print('v', qe.__version__)")
+    assert "v" in out, f"unexpected output: {out!r}"
```
mmcky: These substring assertions match tracebacks too — any `ValueError` contains "v", and `"3" in out` in test_support_enumeration matches traceback line numbers. Print unique sentinels instead (e.g. `print("QE_IMPORT_OK", qe.__version__)`) and assert on those.

mmcky (changes requested):
Nice scoping on the smoke suite — one test per Numba feature class, matching #928's checklist exactly, and the native job mirrors ci.yml's conventions cleanly. All 16 tests passing natively is confirmed by the green native check.

Requesting changes on the CI wiring, which is currently red in three independent ways (details inline):

1. `ci_np2.yml` runs bare `pytest` from the repo root and now fails collecting `ci/wasm/test_jupyterlite.py` — needs `testpaths = quantecon` in pytest.ini.
2. The Emscripten job hung at the Playwright browser install for the full 6-hour job limit — drop `--with-deps` and add `timeout-minutes:` to both jobs.
3. `--timeout=660` needs `pytest-timeout`, which isn't installed.
4. The wasm job lacks `needs: native`, so the fast gate doesn't actually gate.

Also worth noting: since quantecon isn't packaged on emscripten-forge-4x, the wasm job installs the released conda-forge noarch package — it tests the stack, not this repo's source. That's exactly what Phase 0 needs, but it argues for `schedule`/`workflow_dispatch` triggers rather than every PR, and it means the Playwright helpers in test_jupyterlite.py have never actually executed — I've flagged a few likely flakes inline.

On scope: the workflow itself is really #933's deliverable, and #928's remaining deliverable is the results table in the issue — which can't be filled in until the wasm job completes once. Two options: split the wasm job into a follow-up PR under #933 and merge the environment + suite + native gate now, or keep it together and switch "Closes #928" to "Part of #928" until the results are recorded. Happy either way.

ci/wasm/test_jupyterlite.py
DerThorsten: not sure if this is the issue here, but when installing playwright from the conda-forge channel instead of the [microsoft channel](https://anaconda.org/channels/Microsoft/packages/playwright/overview), one sees the error "ModuleNotFoundError: No module named 'playwright'"

.github/workflows/ci_wasm_smoke.yml
mmcky: The Emscripten job moved out to the #933 follow-up per the split (3baf546) — `needs: native` rides along in the parked version there.

.github/workflows/ci_wasm_smoke.yml
mmcky: Dropped along with the wasm job in 3baf546, and it stays out of the parked #933 version — as you say, the build isn't the slow part.

.github/workflows/ci_wasm_smoke.yml
mmcky: Native job got `timeout-minutes: 30` in 3baf546; the wasm job moves to #933 with `--with-deps` dropped and `timeout-minutes: 45`.

.github/workflows/ci_wasm_smoke.yml
mmcky: Moot here now — the wasm job moves to the #933 follow-up with `pytest-timeout` installed and `--timeout=900` (a healthy cold boot can legitimately exceed 660s while still making progress).

ci/wasm/environment.yml
mmcky: Agreed — this drove the split: the runner job moves to a #933 PR that iterates under `pull_request` and flips to `schedule` + `workflow_dispatch` just before merge (dispatch only works once the workflow exists on the default branch, so it can't be exercised from this PR anyway).

ci/wasm/environment.yml
mmcky: Dropped `pytest` and softened the docstring in 1e0c598. Kept the `IS_EMSCRIPTEN`/xfail branches — inert natively, and they document the expected behaviour for when #933 ships the file into the site.

ci/wasm/test_jupyterlite.py
mmcky: Done in a18c110 — `testpaths = quantecon` in pytest.ini. Verified bare `pytest` collects exactly what it did before this PR (605 tests, no ci/wasm), and explicit paths like `pytest ci/wasm/smoke_test.py` still work.

ci/wasm/test_jupyterlite.py
mmcky: Gone entirely in 564dc8d — `_run` now waits for a per-cell sentinel in the output area (and boot waits for a probe cell's sentinel), so nothing depends on the status indicator at all.

ci/wasm/test_jupyterlite.py
mmcky: Fixed in 564dc8d by the same sentinel rework — each cell is wrapped in `try/finally: print(<unique token>)` and `_run` waits for that token, so stale-idle can't return early, and a failing cell still emits the sentinel (traceback in the assertion message instead of a timeout).

ci/wasm/test_jupyterlite.py
mmcky: Done in 564dc8d — unique uppercase sentinels throughout (`QE_IMPORT_OK`, `QE_NE_COUNT 3`, `QE_GINI_EXPECTED_ERROR`, …), which can't collide with traceback text.

ci/wasm/test_jupyterlite.py
mmcky: Thanks @DerThorsten — really appreciate you taking a look! 🙏 Good tip on the conda-forge playwright packaging; here it turned out simpler (the numba-rc job's env doesn't install playwright at all — fixed by scoping pytest collection to `quantecon/`), but definitely one to remember if playwright ever lands in one of our conda env files.

mmcky:
Thanks @kp992 — really nice scoping on the suite itself: every API call checks out and the native run is green, so all the changes below are CI wiring, not test content. Per the review I've pushed the fixes directly to your branch (hope that's OK!):

- a18c110 — `testpaths = quantecon` in pytest.ini, which un-breaks the numba-rc workflow's bare `pytest` (it was dying collecting `test_jupyterlite.py` without playwright). Bare `pytest` collects exactly what it did before this PR; `pytest ci/wasm/smoke_test.py` still works.
- 3baf546 — the Emscripten/JupyterLite job is split out to a follow-up PR under #933: it tests the released conda-forge package rather than this PR's source, and its first run hung at the Playwright browser install for the full 6-hour limit, so it needs its own red/green iteration without holding this PR hostage. The job — with the review fixes folded in — is parked in #933 ready to pick up. The native gate stays on every PR and now has `timeout-minutes: 30`.
- 564dc8d — `test_jupyterlite.py` stays here, reworked to be sentinel-driven: each cell is wrapped in `try/finally` printing a unique token and `_run` waits for that token in the output area. This removes the kernel-status selector, the stale-idle race, and the traceback-prone substring assertions in one stroke, and switches to `insert_text` so multi-line cells aren't mangled by the console's run-on-Enter binding.
- 1e0c598 — dropped the unused `pytest` from the kernel env and softened the smoke_test docstring (in-kernel pytest execution is #933 wiring).

PR body updated to "Part of #928" since the results table there can only be filled after the first successful WASM run.

kp992:
Thanks @mmcky for pushing the fixes. This looks a good start to me. Please feel free to update or merge this. I will move to the next item in the TODO list.

mmcky:
Thanks @kp992 for the quick turnaround! Nothing further to apply on top of a18c110…1e0c598 — CI is fully green (the numba-rc `tests` job is back to passing, and the native smoke gate runs in about a minute), so this is ready to merge once I refresh my review.

Follow-ups queued after merge: the #933 PR seeds from the runner job parked there, and the first green WASM run supplies the results table for #928. Enjoy the next TODO item!

mmcky:
@oyamad are you happy for us to work through the WASM work package and make staged updates. I have documented the plan in #941 re: interaction with releases. I have gated checks to make sure any merges don't hold up any release milestones.

oyamad:
@mmcky Yes; now the blocker is not LAPACK, but the jitted generator issue (https://github.com/QuantEcon/QuantEcon.py/issues/927#issuecomment-5338104706).

mmcky (approved):
Approving — every item from the 2026-08-16 review is addressed (a18c110…1e0c598), all review threads are resolved, and CI is fully green: the full 3.12–3.14 × three-OS matrix, the numba-rc `tests` job, Coverage, RTD, and the new native smoke gate (~1 min).

Re @oyamad's note above: the jitted-generator hang is a #927 finding about the browser runtime, not something this PR exercises — nothing here runs on Emscripten yet (the runner job lives in #933). `test_support_enumeration` is exactly the test that will surface it once the runner lands, which is the suite doing its job. I've updated #927 with the new root cause and added the hang-guard/isolation-test notes to #933.

Merging via the UI. Thanks @kp992!

PR #939, opened 2026-08-15
CI: Update Numba to 0.67 release version

oyamad:
https://numba.discourse.group/t/ann-numba-0-67-and-llvmlite-0-49-final-numpy-2-5-and-windows-arm64-support


mmcky:
thanks @oyamad

PR #942, opened 2026-08-19
WASM: use int64 in comb_jit for consistent overflow guards on wasm32

### LINKED ISSUE #929: WASM: audit 32-bit intp behaviour on wasm32 (overflow guards, simplex_index wrapping, dtypes)

kp992:
Fixes #929. Part of #925 (Phase 1).

`comb_jit` was typed `intp(intp, intp)`, making the overflow guard trip at `2**31-1` on wasm32 instead of `2**63-1`. This caused `simplex_grid` to reject grids that fit comfortably in memory, and could silently wrap in `simplex_index` for near-boundary inputs.

- Change `comb_jit` signature to `int64(int64, int64)` and update the `INT64_MAX` guard throughout
- Update `num_compositions_jit` docstring
- Update `TestCombJit` to test against `np.int64` limits (platform-agnostic)
- Update overflow comments in `test_gridtools.py`
- Add `test_simplex_grid_comb_int64` to the WASM smoke suite

Assumes #938 as baseline.

### DIFF
--- a/quantecon/util/numba.py
+++ b/quantecon/util/numba.py
@@ -77,11 +77,11 @@ def _numba_linalg_solve_ol(a, b):
     return _numba_linalg_solve_impl
 
 
-@jit(types.intp(types.intp, types.intp), nopython=True, cache=True)
+@jit(types.int64(types.int64, types.int64), nopython=True, cache=True)
 def comb_jit(N, k):
     """
     Numba jitted function that computes N choose k. Return `0` if the
-    outcome exceeds the maximum value of `np.intp` or if N < 0, k < 0,
+    outcome exceeds the maximum value of `np.int64` or if N < 0, k < 0,
     or k > N.
 
     Parameters
@@ -97,14 +97,14 @@ def comb_jit(N, k):
     """
     # From scipy.special._comb_int_long
     # github.com/scipy/scipy/blob/v1.0.0/scipy/special/_comb.pyx
-    INTP_MAX = np.iinfo(np.intp).max
+    INT64_MAX = np.iinfo(np.int64).max
     if N < 0 or k < 0 or k > N:
         return 0
     if k == 0:
         return 1
     if k == 1:
         return N
-    if N == INTP_MAX:
+    if N == INT64_MAX:
         return 0
 
     M = N + 1
@@ -114,7 +114,7 @@ def comb_jit(N, k):
 
     for j in range(1, nterms + 1):
         # Overflow check
-        if val > INTP_MAX // (M - j):
+        if val > INT64_MAX // (M - j):
             return 0
 
         val *= M - j
--- a/.github/workflows/ci_wasm_smoke.yml
+++ b/.github/workflows/ci_wasm_smoke.yml
@@ -10,12 +10,14 @@ on:
 
 jobs:
   # ------------------------------------------------------------------
-  # Job 1 — native Python.  Fast gate: ensures the suite itself is
-  # correct and all 16 tests pass before the slow WASM build starts.
+  # Native Python gate: ensures the smoke suite itself is correct and
+  # all tests pass against this repo's source.  The Emscripten /
+  # JupyterLite job that consumes the same suite lands with issue #933.
   # ------------------------------------------------------------------
   native:
     name: Smoke suite (native)
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
     - uses: actions/checkout@v7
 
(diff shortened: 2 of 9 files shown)


oyamad:
@kp992 Thanks, but I remember the `np.intp` typing in `comb_jit` was intentional. Its output is used as an index or size. So I would keep `intp`.

(`simplex_index` is a pure Python function by the way.)

kp992:
Running this example:
```py
import quantecon as qe
grid = qe.simplex_grid(3, 4)
# shape: (L, m) where L = C(4+3-1, 3-1) = 15
assert grid.shape == (15, 3)
assert np.all(grid.sum(axis=1) == 4)
assert np.all(grid >= 0)
```

It throws runtime error:
```
File /lib/python3.13/site-packages/numba/core/codegen.py:632, in CodeLibrary._get_compiled_object(self)
    630     raise ValueError("object caching not enabled in %s" % (self,))
    631 if self._compiled_object is None:
--> 632     raise RuntimeError("no compiled object yet for %s" % (self,))
    633 return self._compiled_object

RuntimeError: no compiled object yet for <Library 'comb_jit' at 0x9e12360>
```

quantecon/util/numba.py
```diff
@@ -97,14 +97,14 @@ def comb_jit(N, k):
     """
     # From scipy.special._comb_int_long
     # github.com/scipy/scipy/blob/v1.0.0/scipy/special/_comb.pyx
-    INTP_MAX = np.iinfo(np.intp).max
+    INT64_MAX = np.iinfo(np.int64).max
     if N < 0 or k < 0 or k > N:
```
kp992, line 101: We do have checks here for negative indices cases @oyamad if we are fine to migrate to int64?

oyamad:
The `no compiled object yet` error confirmed.

Opened an Issue at upstream: emscripten-forge/recipes#6309

mmcky:
Thanks @kp992! I've updated #929 — the `intp` typing in `comb_jit` is intentional (as @oyamad noted, it's used as an array size), so that issue is now documentation-only. The `no compiled object yet` error you found is the real blocker and now has its own issue, #944.

kp992:
Thanks for the review @oyamad and @mmcky. Since the related issue is updated, I am closing this as the experimental PR.

PR #943, opened 2026-08-21
WASM: defer eager Numba gufunc/jit compilation to first use (#930)

### LINKED ISSUE #930: WASM: reduce import-time eager Numba compilation
> [!NOTE]
> **Updated 2026-08-21.** Added a Status section: PR #943 implements Option 1 but is stacked on #938 (now merged) and on the superseded #942, and the "measure first" gate has not been performed. Corrected the claim that `comb_jit` benefits from the persistent cache — per #944 and emscripten-forge/recipes#6309 its `cache=True` is currently part of a failing cached→cached chain on warm sessions. The original text is preserved in the edit history.

Part of #925 (Phase 1). **Measure first**: the Phase 0 deployment (#928) was meant to produce cold/warm import-time benchmarks — if warm import lands under a couple of seconds, deprioritise this issue. As of 2026-08-21 that measurement has not been made; see Status below.

## Problem

`import quantecon` triggers eager Numba compilation of five callables:

| Callable | Where | Why eager |
|---|---|---|
| `_probvec_parallel` | [random/utilities.py#L94](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/random/utilities.py#L94) | module-level `guvectorize` with explicit signature |
| `_probvec_cpu` | [random/utilities.py#L98](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/random/utilities.py#L98) | same |
| `sample_without_replacement` gufunc | [random/utilities.py#L162](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/random/utilities.py#L162) | same |
| `_ints_arr_to_bits` | [vertex_enumeration.py#L305](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/vertex_enumeration.py#L305) | same |
| `comb_jit` | [util/numba.py#L80](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/util/numba.py#L80) | eager `@jit` signature `intp(intp, intp)` |

Natively this costs milliseconds and nobody notices. In the browser, each eager compile is a full LLVM optimisation + WASM object emission + in-process LLD link + side-module load, and the emscripten-forge Numba patch 0007 force-disables `cache=True` for `@guvectorize`/`@vectorize` — so the four gufunc compiles are **not** amortised by the persistent cache and are paid every session, on the critical path of every notebook.

`comb_jit` is an ordinary dispatcher with `cache=True`, but on the emscripten-forge build that is currently a liability rather than an asset. Per #944 and emscripten-forge/recipes#6309, a `cache=True` caller that is a cache miss fails with `RuntimeError: no compiled object yet` when it links a `cache=True` callee restored from the persistent cache in the same session, and `comb_jit` is exactly such a callee for `num_compositions_jit` → `simplex_grid` and for `k_array_rank_jit`. The eager signature only makes `comb_jit` a guaranteed cache hit at import in every warm session, which is why the failure is deterministic today.

## Options, in increasing order of ambition

1. Convert the eagerly-signed gufuncs to lazy compilation — drop the explicit signature lists (dynamic gufuncs), or construct the parallel/cpu variants on first use inside `probvec`.
2. Replace the small gufuncs with `@njit(cache=True)` loop implementations exposed through `@overload` — converts them from never-cached to persistently cached on Emscripten.
3. Defer heavy submodule imports in `quantecon/__init__.py` via module `__getattr__` (PEP 562) — also improves native import time.

## Status (2026-08-21)

**Implementation.** PR #943 (@kp992, opened 2026-08-21, not yet reviewed) implements Option 1 in its "construct on first use" form: the explicit gufunc signatures are retained, the four gufuncs are built lazily behind module-level sentinels on first call, and `comb_jit` becomes a lazy `@jit(nopython=True, cache=True)`. Options 2 and 3 are untouched.

**Rebase required before review.** #943 is stacked on #938 (merged 2026-08-21, now on `main`) and on #942's commit 97518e0, so its diff still carries #942's `int64` widening of `comb_jit` — the `np.int64()` casts and `INT64_MAX`, the docstrings in `util/numba.py` and `_gridtools.py`, the `MAX_INT64`/`test_max_int64` renames in `test_numba.py`, and the smoke test `test_simplex_grid_comb_int64` — and its body cites "the int64 arithmetic guarantees from #929". The rewritten #929 concluded that `intp` is intentional (the result is an array size) and #942 is being closed unmerged. #943 therefore has to be rebased onto `main`, dropping 97518e0 and all `int64` residue and keeping a lazy `@jit(nopython=True, cache=True)` over the unchanged `intp` body; its "Depends on #942" line should be retracted.

**The "measure first" gate is unmet.** No JupyterLite `import quantecon` timing exists anywhere in #928, #938 or #943: #928's timing item is unchecked, #938's native `test_import_time` only asserts `elapsed < 30` s, the browser harness `test_import_quantecon` is untimed, and the Emscripten runner is parked in #933. `import quantecon` does work in the xeus-python kernel (#944), so a manual measurement is feasible today. **The before/after warm-cache import timing (acceptance criterion 1) is a pre-merge requirement for #943** unless a maintainer explicitly waives it in a comment here; merging #943 as written would auto-close this issue with that criterion unmet. The number can come from a manual JupyterLite run now, or from the #933 runner once the browser harness records cold and warm import times.

**What Option 1 does and does not buy.** Patch 0007 forces `cache=False` for every `@vectorize`/`@guvectorize` on Emscripten, so after #943 the four gufuncs are still never cached in the browser: the compile cost moves from import to first call but is paid once per session. Only Option 2 (`@njit(cache=True)` loops behind `@overload`) would make them persistently cached. Whether that is worth doing is what the measurement above should tell us.

**Relation to #944.** #943 changes the `comb_jit` decorator that #944 is about, but it does not fix #944. The upstream reproducer (emscripten-forge/recipes#6309) fails with two plain lazy `@njit(cache=True)` functions whenever a cache-miss caller links a cache-hit callee, so a lazy `comb_jit` restored from the persistent cache still breaks the first compile of `num_compositions_jit`/`simplex_grid` or `k_array_rank_jit` in that session. #943 narrows the exposure — the literal `qe.simplex_grid(3, 4)` reproducer passes in cold sessions and in warm sessions where the caller is also cached — but the failure remains whenever `comb_jit`'s cache entry is a hit and the caller's is a miss. #944's decision (wait for upstream) stands, and #943 should not be cited as resolving it.

## Acceptance criteria

- [ ] Warm-cache `import quantecon` time in JupyterLite measured before and after; result recorded here (pre-merge requirement for #943 unless explicitly waived)
- [ ] No public API change (`qe.random.probvec(..., parallel=...)` keeps its signature)

## Related

- #943 — implements Option 1; needs a rebase onto `main` dropping 97518e0, and the import timing above, before merge
- #942 — `int64` widening of `comb_jit`; superseded by #929 and being closed unmerged; #943 currently carries its commit
- #929 — `intp` in `comb_jit` is intentional; documentation-only
- #944 — `cache=True` callers of cache-restored functions fail with "no compiled object yet" (reported upstream as emscripten-forge/recipes#6309); not addressed by #943
- #928 — Phase 0 results table; its cold/warm import-timing item feeds this issue and is still unchecked
- #938 (merged) and #933 — browser smoke suite and its parked Emscripten runner; the natural home for a recorded import timing

kp992:
Closes #930, part of #925.

### What

Four guvectorize callables compiled at `import quantecon` time are now compiled lazily on first call:

| Callable | File | Change |
|---|---|---|
| `_probvec_parallel` | `random/utilities.py` | module-level gufunc → lazy Python wrapper |
| `_probvec_cpu` | `random/utilities.py` | same |
| `_sample_without_replacement` | `random/utilities.py` | `@guvectorize` decorator → lazy wrapper |
| `_ints_arr_to_bits` | `game_theory/vertex_enumeration.py` | same |

### How

Gufuncs are replaced by thin Python wrapper functions that create the underlying ufunc on first call and cache it in a module-level sentinel. `_probvec_cpu` keeps the same name so the `game_theory/random.py` import is unchanged.

### Why this matters for WASM

On native CPython, eager compilation costs milliseconds. In a JupyterLite/WASM kernel, each compile is a full LLVM optimise + WASM object emit + in-process LLD link. The emscripten-forge Numba patches also force-disable `cache=True` for `@guvectorize`, so the four gufunc compiles were paid on every session on the critical import path of every notebook.

### Public API

No change. `qe.random.probvec(..., parallel=...)`, `qe.random.sample_without_replacement(...)`, and `qe.game_theory.vertex_enumeration(...)` are unaffected.

### DIFF
--- a/quantecon/game_theory/vertex_enumeration.py
+++ b/quantecon/game_theory/vertex_enumeration.py
@@ -296,25 +296,33 @@ class _BestResponsePolytope:
         self.trans_recip = trans_recip
 
 
-@guvectorize(['(i4[:], u8[:])'], '(m)->()', nopython=True, cache=True)
-def _ints_arr_to_bits(ints_arr, out):
+def _ints_arr_to_bits_kernel(ints_arr, out):  # pragma: no cover
+    m = ints_arr.shape[0]
+    out[0] = 0
+    for i in range(m):
+        out[0] |= np.uint64(1) << np.uint64(ints_arr[i])
+
+
+_ints_arr_to_bits_gufunc = None
+
+
+def _ints_arr_to_bits(ints_arr):
     """
     Convert an array of integers representing the set bits into the
     corresponding integer.
 
-    Compiled as a ufunc by Numba's `@guvectorize`: if the input is a
-    2-dim array with shape[0]=K, the function returns a 1-dim array of
-    K converted integers.
+    If the input is a 2-dim array with shape[0]=K, returns a 1-dim array
+    of K converted integers.
 
     Parameters
     ----------
-    ints_arr : ndarray(int32, ndim=1)
+    ints_arr : ndarray(int32, ndim=1 or 2)
         Array of distinct integers from 0, ..., 63.
 
     Returns
     -------
-    np.uint64
-        Integer with set bits represented by the input integers.
+    np.uint64 or ndarray(uint64)
+        Integer(s) with set bits represented by the input integers.
 
     Examples
     --------
@@ -326,10 +334,12 @@ def _ints_arr_to_bits(ints_arr, out):
     array([ 7, 11], dtype=uint64)
 
     """
-    m = ints_arr.shape[0]
-    out[0] = 0
-    for i in range(m):
-        out[0] |= np.uint64(1) << np.uint64(ints_arr[i])
+    global _ints_arr_to_bits_gufunc
+    if _ints_arr_to_bits_gufunc is None:
+        _ints_arr_to_bits_gufunc = guvectorize(
+            ['(i4[:], u8[:])'], '(m)->()', nopython=True, cache=True
+        )(_ints_arr_to_bits_kernel)
+    return _ints_arr_to_bits_gufunc(ints_arr)
 
 
 @jit(nopython=True, cache=True)
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -89,14 +89,26 @@ def _probvec(r, out):  # pragma: no cover
         out[i] = r[i] - r[i-1]
     out[n] = 1 - r[n-1]
 
-_probvec_parallel = guvectorize(
-    ['(f8[:], f8[:])'], '(n), (k)', nopython=True, target='parallel',
-    cache=True
-    )(_probvec)
-_probvec_cpu = guvectorize(
-    ['(f8[:], f8[:])'], '(n), (k)', nopython=True, target='cpu',
-    cache=True
-    )(_probvec)
+_probvec_parallel_gufunc = None
+_probvec_cpu_gufunc = None
+
+
+def _probvec_parallel(r, out):
+    global _probvec_parallel_gufunc
+    if _probvec_parallel_gufunc is None:
+        _probvec_parallel_gufunc = guvectorize(
+            ['(f8[:], f8[:])'], '(n),(k)', nopython=True,
+            target='parallel', cache=True)(_probvec)
+    _probvec_parallel_gufunc(r, out)
+
+
+def _probvec_cpu(r, out):
+    global _probvec_cpu_gufunc
+    if _probvec_cpu_gufunc is None:
+        _probvec_cpu_gufunc = guvectorize(
+            ['(f8[:], f8[:])'], '(n),(k)', nopython=True,
+            target='cpu', cache=True)(_probvec)
+    _probvec_cpu_gufunc(r, out)
 
 
 def sample_without_replacement(n, k, num_trials=None, random_state=None):
@@ -153,10 +165,9 @@ def sample_without_replacement(n, k, num_trials=None, random_state=None):
     return result
 
 
-@guvectorize(['(i8, f8[:], i8[:])'], '(),(k)->(k)', nopython=True, cache=True)
-def _sample_without_replacement(n, r, out):
+def _sample_without_replacement_kernel(n, r, out):  # pragma: no cover
     """
-    Main body of `sample_without_replacement`. To be complied as a ufunc
+    Main body of `sample_without_replacement`. To be compiled as a ufunc
     by guvectorize of Numba.
 
     """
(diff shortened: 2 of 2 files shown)


mmcky:
Thanks @kp992. Now that #938 is merged, could you rebase this onto `main` and drop 97518e0 (the `int64` widening from #942)? #929 concluded the `intp` typing in `comb_jit` is intentional since its result is an array size, so please keep `intp` and drop the `np.int64` casts, `INT64_MAX`, the `MAX_INT64`/`test_max_int64` renames and `test_simplex_grid_comb_int64` — the lazy `@jit(nopython=True, cache=True)` over the unchanged body is all #930 needs. The "Depends on #942" line can go too.

Two more things. #930 asks for a before/after warm-cache `import quantecon` timing in JupyterLite before this merges — could you add it to the PR description, or say so if you can't and we'll decide whether to waive it. And this doesn't address #944 (the upstream reproducer in emscripten-forge/recipes#6309 uses lazy `@njit(cache=True)` functions only), so please don't reference it as such.

mmcky:
Thanks @kp992 — closing this in favour of #957, which opens @oyamad's implementation of the same idea from his `wasm-930-dynamic-gufuncs` branch. That is not a reflection on the work here: your 21 Aug rebase did exactly what was asked, the `int64` residue is gone, the change is confined to two files, and CI is green.

Two things decided it. His version builds the gufuncs through memoised factory functions rather than module-level sentinels mutated through `global`, which is less state to reason about; and it converts `_sample_without_replacement` and `_ints_arr_to_bits` to dynamic (signature-less) gufuncs rather than keeping explicit signatures behind wrappers. More importantly, it documents *why* the explicit signature is retained for the two `_probvec` targets — an explicit-signature gufunc and a dynamic one of the same kernel share an incompatible cache entry under `cache=True` (numba#10128). That constraint was written down nowhere in the codebase, and it is the kind of thing that gets silently re-broken later.

One correction that applied to both versions: this PR says "Closes #930", but #930's acceptance criterion 1 — warm-cache `import quantecon` timed in JupyterLite, before and after — is still unmet, so merging either implementation would have auto-closed the issue with that criterion outstanding. #957 says "Part of #930" instead. It matters more than bookkeeping: emscripten-forge patch 0007 forces `cache=False` for every `@guvectorize`, so in the browser these four compiles are still paid once per session either way. The change moves the cost from import to first call rather than removing it, and whether that is a win depends on how many notebooks reach `probvec`, `sample_without_replacement` or `vertex_enumeration` — which is what the measurement would tell us.

Thanks for the work on this. The analysis in your description was accurate and it is what got the approach agreed.

kp992:
Thanks @mmcky for the review and feedback. Sounds good to me.

Issue #877, opened 2026-07-25
DISCUSS: Migration path from legacy RandomState to NumPy Generator


mmcky:
### The decision needed

Our RNG plumbing is the 2015-era scikit-learn idiom: `check_random_state` (`util/random.py`) normalizes to `np.random.RandomState`, reaches into `np.random.mtrand._rand` for the global instance, and `rng_integers` exists solely to bridge `RandomState`/`Generator` signature differences. NumPy has recommended `Generator`/`default_rng` since 1.17 (2019): the actively developed API, faster bit generators, better streams. We accept `Generator` today, but every internal default and documented example takes the legacy path.

This is a decision, not a patch, because reproducibility is the product: for integer seeds, `default_rng(seed)` and `RandomState(seed)` produce different streams, so a silent default switch changes every seeded simulation result in downstream research and the QuantEcon lectures.

### Questions for this thread

1. Do we migrate defaults at all, or only document `Generator` as the preferred input?
2. If we migrate: flag-day at a major-ish version with a loud changelog, or a transition period where integer seeds emit a `FutureWarning` naming the change date?
3. What do the lecture repositories need from us (coordination, pinned examples)?
4. Does `check_random_state` grow a `default="generator"` parameter, or do we adopt `np.random.default_rng` passthrough plus `spawn` for parallel paths?

### Suggested outcome

A short policy (target version, warning strategy, docs plan) recorded here, then this thread closes and spawns the implementation issue. Related: the vendoring of `rng_integers` (this milestone) is worth doing first regardless; a completed migration eventually deletes the vendored copy too.

---

_From the July 2026 technical-debt audit (AI-assisted; claims verified against `28d4b3b` on 2026-07-25)._

mmcky:
Some facts to frame this, then a concrete proposal on each of the four questions so the thread has something to agree with or reject rather than a blank page.

**The surface.** `random_state=` appears 182 times across 13 non-test modules — `game_theory/` (brd, fictplay, localint, logitdyn, normal_form_game, random, game_generators/bimatrix_generators), `markov/` (approximation, core, random), `random/utilities.py`, and `util/` (random, `__init__`). Against that, the `rng=` spelling introduced by #917 exists on exactly one public function, `random.draw`.

[...]

PR #946, opened 2026-08-30
MAINT: vendor rng_integers helper

### LINKED ISSUE #868: RFC: Vendor rng_integers instead of importing from SciPy's private _lib._util
### Problem

`quantecon/util/random.py:9` imports a helper from a private SciPy module:

```python
from scipy._lib._util import rng_integers  # noqa: F401
```

`scipy._lib` carries no stability guarantee — SciPy may rename, move, or delete `rng_integers` in any release, including a patch release. Because our requirement is an unbounded `scipy>=1.5.0`, the day that happens every fresh install of quantecon breaks at import of the affected modules, with no action on our side. The `noqa` on the import line is the linter having warned us already.

The symbol is re-exported via `quantecon/util/__init__.py` and used at 13 call sites in 8 modules: `markov/core.py:510`; `game_theory/normal_form_game.py:416`; `game_theory/logitdyn.py:112,153`; `game_theory/brd.py:115`; `game_theory/localint.py:115,180`; `game_theory/random.py:178`; `game_theory/game_generators/bimatrix_generators.py:258,262,596,612,614`.

### Proposed change

Vendor the function instead of importing it — the same pattern this repo already uses for `copy_if_needed` in `quantecon/util/compat.py` (copied from SciPy with attribution rather than imported):

1. Copy the (small) `rng_integers` implementation into `quantecon/util/random.py` with a provenance comment and SciPy license attribution.
2. Keep the name and signature identical — `quantecon.util.rng_integers` — so zero call sites change. Behavior-preserving by construction.
3. Add a smoke test pinning the contract (dtype, inclusive/exclusive endpoints, `RandomState` vs `Generator` dispatch) so future drift is caught here, not by users.

### Acceptance criteria

- [ ] `grep -rn "scipy._lib" quantecon/` returns nothing
- [ ] Full suite passes; no public API or seeded-stream change
- [ ] New unit test covers both RNG types and endpoint semantics

---

mameikagou:
Fixes #868.

Replaces the private SciPy import with the BSD-licensed `rng_integers` implementation and adds compatibility coverage for `RandomState`, `Generator`, and the global NumPy random state.

Tests:
- `pytest quantecon/util/tests/test_random.py` (5 passed)
- `flake8 --select=F401,F405,E231 quantecon`
- Full suite: 653 passed; the notebook dependency test failed while connecting to `raw.githubusercontent.com`.

### DIFF
--- a/quantecon/util/random.py
+++ b/quantecon/util/random.py
@@ -5,8 +5,101 @@ Utilities to Support Random State Infrastructure
 import numpy as np
 import numbers
 
-# To be called as util.rng_integers
-from scipy._lib._util import rng_integers  # noqa: F401
+# Adapted from SciPy's ``rng_integers`` implementation:
+# https://github.com/scipy/scipy/blob/v1.17.1/scipy/_lib/_util.py
+# Copyright (c) 2001-2002 Enthought, Inc. 2003, SciPy Developers.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+#    contributors may be used to endorse or promote products derived from
+#    this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


mmcky:
Thanks @mameikagou — this is verified and in good shape. The vendored function matches SciPy's implementation exactly across 18 comparisons (both `RandomState` and `Generator`, `endpoint` either way, `size` as `None`, an int and a tuple, and the `high=None` form), the private `scipy._lib._util` import is gone, and the `quantecon/util` tests pass.

We are holding it rather than merging, because #877 — settling our migration path from legacy `RandomState` to NumPy `Generator` — turns out to govern the shape of this helper rather than merely sit alongside it. SciPy's implementation resolves `gen=None` to `np.random.mtrand._rand`, the legacy global singleton, and that is exactly the choice #877 exists to make; adopting the SciPy version wholesale would settle it by default. There is also a licensing question worth raising: QuantEcon.py is MIT and SciPy is BSD-3, so vendoring brings a licence block with it, where `quantecon/util/compat.py` carries a one-line attribution for code from the same SciPy file.

So the approach may change — writing a small adapter of our own is one option, since the logic is dictated by the two NumPy APIs — and I would rather settle that on #877 than ask you to guess. I have put a concrete proposal there to move it along. Apologies for the delay on a first contribution; the underlying problem in #868 is real and worth fixing, and none of this is a criticism of the work.

PR #948, opened 2026-09-05
WASM: raise OSError for network calls on Emscripten (#931)

### LINKED ISSUE #931: WASM: document browser limitations and raise helpful platform-specific errors
> [!NOTE]
> **Updated 2026-08-21.** Corrected the first item: the URL loaders in `game_converters.py` read the GameTracer `.gam` format (`GAMReader.from_url` / `from_gam_url`), not Gambit. The parallel-fallback items now cite both numba emscripten patches (0007 for the ufunc target, 0010 for `@njit(parallel=True)`, shipped 2026-08-18), and a Related section links the issues that supply the "what works / what doesn't yet" content for the docs page. Scope is unchanged: documentation and error messages only.

Part of #925 (Phase 1). Documentation and error-message work — no behaviour changes on native platforms.

## Items

- [ ] **Network utilities fail in the browser.** [`fetch_nb_dependencies`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/util/notebooks.py#L86) (via `requests`) and the GameTracer `.gam` URL loaders in `game_theory/game_converters.py` — [`GAMReader.from_url`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/game_converters.py#L203-L211) and its wrapper [`from_gam_url`](https://github.com/QuantEcon/QuantEcon.py/blob/13b436b8a43aa53313f6523509bd3f7a35d9dfd2/quantecon/game_theory/game_converters.py#L378-L393) (via `urllib.request`) — perform socket-level HTTP, which does not exist in WASM; JS-fetch-backed shims are kernel-specific and not guaranteed in xeus-python. Both imports are already lazy (`urllib.request` is imported inside the classmethod and is not in `sys.modules` after `import quantecon`) so nothing breaks at import — but the calls themselves will fail obscurely in-browser. Raise a helpful platform-specific error when `sys.platform == "emscripten"` and document the limitation in the docstrings. Pair the new message with an Emscripten-gated assertion in `ci/wasm/smoke_test.py` (see Related).
- [ ] **`probvec` docstring.** `parallel=True` (the default) silently executes serially on Emscripten. Two numba emscripten patches produce this: patch 0007 falls back from the `target='parallel'` ufunc target to `'cpu'` (the path `probvec` takes), and patch 0010 (emscripten-forge/recipes#6293, merged 2026-08-18, shipped as numba 0.67.0 build 1) compiles `@njit(parallel=True)` serially and forces `NUMBA_NUM_THREADS=1`. A one-line note keeps the documented contract honest.
- [ ] **Consider demoting `requests` from a hard dependency to an optional extra.** Its only runtime user is `fetch_nb_dependencies`, which already imports it lazily. It exists as conda-forge `noarch` so it resolves fine in the browser either way — this is dependency hygiene, not a blocker.
- [ ] **A short "QuantEcon.py in the browser" docs page.** The supported stack (JupyterLite + xeus-python + emscripten-forge); what works (sourced from the per-function results table #928 is meant to record); what does not work yet (the jitted-generator hang, #927, and the warm-cache `cache=True` failure, #944); single-threaded execution — all Numba parallelism is silently serialised in the browser (`guvectorize` `target='parallel'` via patch 0007; `@njit(parallel=True)` / `prange` and `NUMBA_NUM_THREADS=1` via patch 0010, emscripten-forge numba 0.67.0 build 1 or later); and the wasm32 limits (coordinate with #929).

## Related

- #928 — Phase 0 smoke checklist; its per-function results table (not yet posted) is the source for the "what works" section of the docs page
- #927 — jitted generators hang on first call in the browser, so `support_enumeration` and `vertex_enumeration` never return (LAPACK itself is verified working); no upstream report yet. The docs page should list both as not working
- #944 — on a warm session, a `cache=True` caller whose callee was restored from the persistent cache fails with `RuntimeError: no compiled object yet` (`simplex_grid` via `num_compositions_jit` → `comb_jit`; also `k_array_rank_jit`); reported upstream as emscripten-forge/recipes#6309 (open); decision is to wait for upstream
- #926 / PR #937 — `gini_coefficient` runs serially in the browser under patch 0010, at O(n²) until the O(n log n) rewrite in #937 (approved, awaiting merge) lands
- #930 / PR #943 — import time: patch 0007 also disables `cache=True` for `@vectorize` / `@guvectorize`, so the four eagerly compiled gufuncs recompile every session; #943 (open) defers all five import-time compiles to first call and does not fix #944
- #929 — wasm32 limits (int32 default integer, `comb_jit` / `simplex_grid` boundary at 2³¹−1, 2–4 GB memory cap shared with the whole kernel); documentation-only, and its second task hands the user-facing wording to this issue
- `ci/wasm/smoke_test.py` — on main since PR #938 merged (2026-08-21); already defines `IS_EMSCRIPTEN`, so it is the home for an Emscripten-gated assertion on the new platform-specific error message. No browser job runs it yet; that runner is #933

kp992:
Closes #931.

- `fetch_nb_dependencies` and `GAMReader.from_url` now raise `OSError` with a descriptive message on `sys.platform == "emscripten"` instead of failing with a cryptic socket error.
- `probvec` docstring notes that `parallel=True` executes serially on Emscripten via Numba patch 0007.
- Two `smoke_test.py` tests verify the new errors fire in the browser kernel (skipped on native).

### DIFF
--- a/ci/wasm/smoke_test.py
+++ b/ci/wasm/smoke_test.py
@@ -229,7 +229,25 @@ def test_simplex_grid():
 
 
 # ---------------------------------------------------------------------------
-# 13. searchsorted — objmode() shim (deprecated helper)
+# 13. Network utilities — raise OSError on Emscripten (#931)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.skipif(not IS_EMSCRIPTEN, reason="Emscripten-only guard check")
+def test_fetch_nb_dependencies_raises_on_emscripten():
+    from quantecon.util import fetch_nb_dependencies
+    with pytest.raises(OSError, match="unavailable in the browser"):
+        fetch_nb_dependencies(["dummy.csv"])
+
+
+@pytest.mark.skipif(not IS_EMSCRIPTEN, reason="Emscripten-only guard check")
+def test_gam_reader_from_url_raises_on_emscripten():
+    from quantecon.game_theory.game_converters import GAMReader
+    with pytest.raises(OSError, match="unavailable in the browser"):
+        GAMReader.from_url("http://example.com/dummy.gam")
+
+
+# ---------------------------------------------------------------------------
+# 15. searchsorted — objmode() shim (deprecated helper)
 # ---------------------------------------------------------------------------
 
 def test_searchsorted():
--- a/quantecon/game_theory/game_converters.py
+++ b/quantecon/game_theory/game_converters.py
@@ -204,7 +204,20 @@ class GAMReader:
         """
         Read from a URL.
 
+        Raises
+        ------
+        OSError
+            On Emscripten (JupyterLite/xeus-python), where HTTP socket access
+            is unavailable.
+
         """
+        if sys.platform == "emscripten":
+            raise OSError(
+                "GAMReader.from_url requires HTTP socket access, which is "
+                "unavailable in the browser (Emscripten/xeus-python). "
+                "Use GAMReader.from_string or GAMReader.from_file with "
+                "pre-downloaded data instead."
+            )
         import urllib.request
         with urllib.request.urlopen(url) as response:
             string = response.read().decode()
--- a/quantecon/random/utilities.py
+++ b/quantecon/random/utilities.py
@@ -33,7 +33,8 @@ def probvec(m, k, random_state=None, parallel=True):
     parallel : bool(default=True)
         Whether to use multi-core CPU (parallel=True) or single-threaded
         CPU (parallel=False). (Internally the code is executed through
-        Numba.guvectorize.)
+        Numba.guvectorize.) On Emscripten (JupyterLite), parallel=True
+        executes serially via the Numba emscripten-forge patch 0007.
 
     Returns
     -------
--- a/quantecon/util/notebooks.py
+++ b/quantecon/util/notebooks.py
@@ -20,6 +20,7 @@ TODO
 """
 
 import os
+import sys
 
 #-Remote Structure-#
 REPO = "https://github.com/QuantEcon/QuantEcon.notebooks"
@@ -71,6 +72,13 @@ def fetch_nb_dependencies(files, repo=REPO, raw=RAW, branch=BRANCH, folder=FOLDE
     by setting ``overwrite=True``.
 
     """
+    if sys.platform == "emscripten":
+        raise OSError(
+            "fetch_nb_dependencies requires HTTP socket access, which is "
+            "unavailable in the browser (Emscripten/xeus-python). "
+            "Pre-download the required files before running in JupyterLite."
+        )
+
     import requests
 
     #-Generate Common Data Structure-#


mmcky:
Thanks @kp992 — this PR implements exactly what #931 asked for but the issue was wrong. See comments from @oyamad. Better to flag it now than after you've built on it. When building these out, please take an adversarial view of the issue and validate it before proceeding on a PR. Let's design before we implement. Does that make sense?

The evidence. Daisuke tested both call paths in a live xeus-python deployment on 6 Sep ([#931 comment](https://github.com/QuantEcon/QuantEcon.py/issues/931#issuecomment-5559369094)): GAMReader.from_url read a .gam file from raw.githubusercontent.com and returned a full 3-player game, and requests.get(...) returned 200 with 4117 bytes. Both functions this PR guards currently work in the browser.

Why they work. urllib3 has supported Emscripten natively since 2.2.0, and per [its documentation](https://urllib3.readthedocs.io/en/stable/reference/contrib/emscripten.html) that support "is automatically enabled if sys.platform is emscripten" — it substitutes EmscriptenHTTPConnection and routes through XMLHttpRequest, or fetch on a worker thread for streaming requests. requests sits on urllib3, so fetch_nb_dependencies inherits it. emscripten-forge has already acted on this: requests-wasm-polyfill was dropped as redundant ([recipes#782](https://github.com/emscripten-forge/recipes/issues/782)) and the companion request to patch urllib.urlopen was closed as "not needed anymore" ([requests-wasm-polyfill#4](https://github.com/emscripten-forge/requests-wasm-polyfill/issues/4)).

The awkward part is that sys.platform == "emscripten" is the very condition that switches urllib3 onto its browser transport — so the guard fires precisely when the network path is being enabled.

Keep: the probvec docstring note. It is correct and independent of all the above — patch 0007 does fall back from target='parallel' to 'cpu', so parallel=True genuinely executes serially. That is item 2 of #931 and it should land.

Drop: both sys.platform == "emscripten" guards, in notebooks.py and game_converters.py, along with the two Emscripten-gated smoke tests asserting them.

If we still want a friendlier error, the shape to aim for is wrapping a failure that actually occurs rather than refusing before trying — let the call proceed, and if it raises, re-raise with guidance about pre-downloading data. That keeps the working case working and still improves the message where the runtime genuinely cannot do it. Better as a follow-up once #933's browser runner exists and we can see what a real failure looks like, rather than guessing now.

Housekeeping: the description says "Closes #931". With the guards removed this covers one of the issue's four items, so "Part of #931" is more accurate. Separately, we will correct item 1 of #931 itself — its claim that these calls "perform socket-level HTTP, which does not exist in WASM" is what sent you down this path, and leaving it would send the next person the same way.

One thing to wait for before reworking. There is an open question about deprecating quantecon/util/notebooks.py in 0.12.0 and dropping requests as a dependency. If that goes ahead, fetch_nb_dependencies is on its way out and any work on it now is wasted. Give us a few days to settle it, then the rework is just the probvec note plus whatever we decide about GAMReader.from_url.

PR #949, opened 2026-09-06
FIX: Pivoting routines: Report numerical breakdown correctly

oyamad:
Fixes to the pivoting routines and their callers, found while benchmarking `linprog_simplex` on the Netlib problems.

- `_lex_min_ratio_test`: `found=False` was returned both when the pivot column has no positive entry and when the lexicographic tie breaking failed to single out one row; the latter was reported as "unbounded" by `linprog_simplex`. Return `found=True` with the first candidate in that case, together with a third value `resolved`.
- `lemke_howson` and `polym_lcp_solver` discarded `found`; they now stop with `converged=False` on breakdown instead of returning a wrong "equilibrium".
- `lcp_lemke`: require `d > 0`; measure ties in the initial ratio test against the minimum ratio.
- `linprog_simplex`: `fun = inf` when unbounded; Phase 1 infeasibility test relative to the scale of `b`; pivot artificial variables out on the largest entry; docstring fixes.
- Guard docstring formatting against `python -OO`, which made `quantecon` unimportable.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Fable 5.1)

### DIFF
--- a/quantecon/optimize/lcp_lemke.py
+++ b/quantecon/optimize/lcp_lemke.py
@@ -196,9 +196,10 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
     return LCPResult(z, success, status, num_iter)
 
 
-lcp_lemke.__doc__ = lcp_lemke.__doc__.format(
-    FEA_TOL=FEA_TOL, TOL_PIV=TOL_PIV, TOL_RATIO_DIFF=TOL_RATIO_DIFF
-)
+if lcp_lemke.__doc__ is not None:  # None under python -OO
+    lcp_lemke.__doc__ = lcp_lemke.__doc__.format(
+        FEA_TOL=FEA_TOL, TOL_PIV=TOL_PIV, TOL_RATIO_DIFF=TOL_RATIO_DIFF
+    )
 
 
 @jit(nopython=True, cache=True)
--- a/quantecon/optimize/lcp_lemke.py
+++ b/quantecon/optimize/lcp_lemke.py
@@ -133,6 +133,8 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
 
     if d is None:
         d = np.ones(n)
+    elif not np.all(d > 0):
+        raise ValueError('d must be strictly positive')
     if tableau is None:
         tableau = np.empty((n, 2*n+2))
     if basis is None:
@@ -143,14 +145,20 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
     art_var = 2*n  # Artificial variable
     pivcol = art_var
 
-    # Equivalent to lex_min_ratio_test
+    # Equivalent to lex_min_ratio_test: the lexicographic tie breaking
+    # reduces to taking the largest row index, as the slack columns form
+    # the identity matrix. Ties are measured against the smallest ratio
+    # found so far, which is not updated on a tie, so that the accepted
+    # set cannot drift away from the minimum by chaining tolerances.
     pivrow = 0
     ratio_min = q[0] / d[0]
     for i in range(1, n):
         ratio = q[i] / d[i]
-        if ratio <= ratio_min + piv_options.tol_ratio_diff:
+        if ratio < ratio_min - piv_options.tol_ratio_diff:  # Smaller
             pivrow = i
             ratio_min = ratio
+        elif ratio <= ratio_min + piv_options.tol_ratio_diff:  # Tie
+            pivrow = i
 
     _pivoting(tableau, pivcol, pivrow)
     basis[pivrow], pivcol = pivcol, pivrow + n
@@ -281,7 +289,7 @@ def _get_solution(tableau, basis, z):
         View to `z`.
 
     """
-    n = z.size
+    n = basis.shape[0]
 
     z[:] = 0
     for i in range(n):
--- a/quantecon/optimize/lcp_lemke.py
+++ b/quantecon/optimize/lcp_lemke.py
@@ -91,6 +91,7 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
                     0 : Solution found successfully
                     1 : Iteration limit reached
                     2 : Secondary ray termination
+                    3 : Numerical difficulties encountered
 
             num_iter : int
                 The number of iterations performed.
@@ -168,7 +169,7 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
     argmins = np.empty(n, dtype=np.int_)
 
     while num_iter < max_iter:
-        pivrow_found, pivrow, _ = _lex_min_ratio_test(
+        pivrow_found, pivrow, resolved = _lex_min_ratio_test(
             tableau, pivcol, 0, argmins,
             piv_options.tol_piv, piv_options.tol_ratio_diff
         )
@@ -176,6 +177,10 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
         if not pivrow_found:  # Ray termination
             success = False
             status = 2
+            break
+        if not resolved:  # Numerical breakdown: lexicographic tie not
+            success = False  # broken, impossible in exact arithmetic
+            status = 3
             break
 
         _pivoting(tableau, pivcol, pivrow)
--- a/quantecon/optimize/lcp_lemke.py
+++ b/quantecon/optimize/lcp_lemke.py
@@ -117,6 +117,8 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
 
     """
     n = M.shape[0]
+    if d is not None and not np.all(d > 0):
+        raise ValueError('d must be strictly positive')
 
     success = False
     status = 1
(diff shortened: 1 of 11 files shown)


quantecon/optimize/lcp_lemke.py
```diff
@@ -133,6 +133,8 @@ def lcp_lemke(M, q, d=None, max_iter=10**6, piv_options=PivOptions(),
 
     if d is None:
         d = np.ones(n)
+    elif not np.all(d > 0):
+        raise ValueError('d must be strictly positive')
```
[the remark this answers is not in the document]
oyamad: Done.

quantecon/optimize/linprog_simplex.py
```diff
@@ -285,6 +288,8 @@ def linprog_simplex(c, A_ub=np.empty((0, 0)), b_ub=np.empty((0,)),
         solve_phase_1(tableau, basis, max_iter, piv_options=piv_options)
     num_iter += num_iter_1
     if not success:
+        if status == 3:  # Unbounded
+            fun = np.inf
```
quantecon-services: Tiny one: I think this branch is unreachable now, since `solve_phase_1` maps status 3 to 4 before returning (which is exactly what `test_linprog_simplex_phase_1_breakdown_is_not_unbounded` checks). It could be dropped, or kept with a short comment if you like the belt-and-braces. Totally harmless either way!

---

quantecon/optimize/linprog_simplex.py
```diff
@@ -574,13 +586,19 @@ def solve_phase_1(tableau, basis, max_iter=10**6, piv_options=PivOptions()):
...
                       piv_options=piv_options)
 
-    if not success:  # max_iter exceeded
+    if not success:
+        if status == 3:
+            # The auxiliary problem is bounded: a numerical breakdown
+            status = 4
         return success, status, num_iter_1
-    if tableau[-1, -1] > piv_options.fea_tol:  # Infeasible
+    if tableau[-1, -1] > piv_options.fea_tol * b_scale:  # Infeasible
```
quantecon-services, line 601: This is the line the main review comment is about. Scaling by the sum of `|b|` is a great idea for FFFFF800, but when the right-hand sides have mixed scales, a real contradiction in a small row can hide under the total. The review body has a two-line reproducer and a sketch of a per-constraint check that seems to work well alongside this total test. Happy to help try it out if useful!

---

quantecon/optimize/tests/test_lcp_lemke.py
```diff
@@ -117,3 +119,60 @@ def test_bug_768(self):
...
+                    basis=np.empty(n, dtype=int), z=np.empty(n))
+    assert_(res.success)
+    assert_allclose(M @ res.z + q, np.zeros(n) + (M @ res.z + q).clip(0))
+    assert_allclose(res.z @ (M @ res.z + q), 0, atol=1e-12)
+
+
+def test_lcp_lemke_numerical_breakdown():
+    # Entries of order 1e14: the lexicographic tie breaking fails within
+    # `tol_ratio_diff`; an arbitrary pivot led to a wrong "solution"
+    # reported as success
```
quantecon-services: Small wording nit: on `main` this instance actually ends with status 2 (ray termination), not a wrong success. The wrong success only shows up at the intermediate commit where unresolved ties were pivoted through. Maybe something like: "`main` reported it as ray termination, and pivoting through the tie led to a wrong 'solution' reported as success". The test itself is spot on!

---

quantecon-services (commented):
Thanks so much for this, @oyamad! This is a really careful piece of work. Separating "no positive entry" from "tie not broken" in `_lex_min_ratio_test` fixes a whole family of silent wrong answers in one go, and the commit-by-commit history made it a pleasure to review.

## What I checked

- Pulled the branch and ran the affected suites locally (pivoting, linprog_simplex, lcp_lemke, lemke_howson, howson_lcp, polymatrix_game, compute_fp): 74 passed, and the CI flake8 selects are clean.
- Reproduced every item in the description on `main` and confirmed it is fixed here:
  - the duplicated-row LP (`A_ub = [[1e14], [1e14]]`) that `main` reports as "unbounded" now returns status 4;
  - an unbounded Phase 2 exit now reports `fun = inf` instead of a finite non-optimal value;
  - the 1e15-scale game that `main` "solves" with a non-equilibrium and `converged=True` now reports `converged=False`;
  - the 2e14 LCP goes from status 2 (ray termination) to the more accurate status 3;
  - the imitation game stops with the specific breakdown warning after 6 evaluations, where `main` runs to `max_iter` after 20 and warns about that instead;
  - `python -OO -c "import quantecon"` fails on `main` and works here.
- Ran a small sweep of 30 random 4x4 games per payoff scale, validating each answer against the equivalent unit-scale game (so that the absolute tolerance in `is_nash` doesn't flatter tiny payoffs). The branch mostly turns silent wrong answers into explicit failures; the only place it gives up answers that `main` had right is the 1e12 scale (20 valid before, 18 after):

| payoff scale | `main`: converged / valid NE | this PR: converged / valid NE |
|---|---|---|
| 1e-12 | 30 / 3 | 0 / 0 |
| 1e-10 | 30 / 13 | 30 / 13 |
| 1 | 30 / 30 | 30 / 30 |
| 1e12 | 30 / 20 | 20 / 18 |
| 1e15 | 30 / 3 | 0 / 0 |

## One thing I'd love your eyes on

The relative Phase 1 infeasibility test (`fea_tol * sum|b|`) is a good idea for FFFFF800, but as written it can let a genuinely infeasible problem through when one constraint has a large right-hand side and the contradiction sits in a small one:

```python
A_eq = np.array([[1., 0.], [0., 1.], [0., 1.]])
b_eq = np.array([1e6, 0., 0.9])          # x2 = 0 and x2 = 0.9
res = linprog_simplex(np.array([-1., -1.]), A_eq=A_eq, b_eq=b_eq)
# this PR: success=True,  status=0, x=[1e6, 0]
# main:    success=False, status=2
```

The residual 0.9 sits below `fea_tol * (1e6 + 0.9) ≈ 1.0`, the artificial variable of the third row stays basic at level 0.9 with an all-zero row in the `x` columns (so the pivot-out loop never touches it), and Phase 2 proceeds as if the problem were feasible.

One fix that seems to work well is to keep the total test and additionally check each artificial variable that is still basic against the scale of its *own* constraint, using the `|b|` values from the initial tableau:

```python
    b_scale = max(1., tableau[-1, -1])
    b_abs = tableau[:L, -1].copy()  # |b_k| per constraint, before Phase 1
    success, status, num_iter_1 = \
        solve_tableau(tableau, basis, max_iter, skip_aux=False,
                      piv_options=piv_options)
    ...
    if tableau[-1, -1] > piv_options.fea_tol * b_scale:  # Infeasible
        ...
    for i in range(L):
        if basis[i] >= nm:  # Artificial variable k still basic
            k = basis[i] - nm
            if tableau[i, -1] > piv_options.fea_tol * max(1., b_abs[k]):
                success = False
                status = 2
                return success, status, num_iter_1
```

I tried this in a scratch copy: it rejects the example above, still solves the 1e5-scale case in `TestLinprogSimplexPhase1Scale`, and all 20 tests in `test_linprog_simplex.py` pass. The one caveat is that a roundoff residual landing in a row with a small `|b_k|` would now be flagged, so it would be worth re-running the Netlib benchmark with it. Very happy to help with that if useful!

## Small nits

I left three inline notes: an unreachable branch in `linprog_simplex`, a pointer to the tolerance line above, and a test comment that describes the intermediate state of the branch rather than `main`. One more that I couldn't attach inline because the line isn't in the diff: the `polym_lcp_solver` docstring (line 71) still says the result is an equilibrium "if not stopped early by reaching `max_iter`"; breakdown is now a second early exit worth a mention there.

## Ideas for follow-ups (definitely not for this PR!)

- **Scale normalization in `_initialize_tableaux`.** Equilibria are invariant to positive affine transformations of each player's payoffs, and the setup already shifts payoffs to be positive. Dividing by the maximum payoff as well would make the absolute tolerances effectively relative. The table above suggests the unit scale is the only one that is reliably right today (see the 1e-10 row, where both branches return wrong answers with `converged=True`). I think this might be a very cheap way to get the "scale-aware tolerance" mentioned in the first commit message.
- **A `RuntimeWarning` on non-convergence in `lemke_howson` / `polym_lcp_solver`.** With the default `full_output=False`, a caller still receives the vector read from the partially pivoted tableaux (all zeros in the new tests) with no signal at all. A warning like the one in `compute_fixed_point` would make the new breakdown reporting visible to everyone, not only `full_output=True` users.
- **`x` and `lambd` on Phase 1 failure.** They come back as allocated by `np.empty`, so callers can see values like `6e151` next to `fun = -inf`. Zeroing them (or a note in the docstring) would round out the new failure-output conventions nicely.

Thanks again! This is a big improvement in how honestly the pivoting routines report what happened, and I learned a lot reading it.

---

oyamad:
- The mixed-scale infeasibility: correct, and the diagnosis is exact. The artificial variable of a constraint enters that constraint only, so its level after Phase 1 is the residual of the constraint; ecc31df tests each artificial variable still basic against `fea_tol` times the scale of its own right hand side, in addition to the total, with the example above as a test. On the Netlib problems of the benchmark (51 problems, default tolerances) nothing changes: same 40 solved, FFFFF800 included, and no problem declared infeasible.
- The unreachable branch is removed in the same commit.
- The test comment is reworded (also in the Julia port, QuantEcon/QuantEcon.jl#406), and the `polym_lcp_solver` docstring now mentions breakdown as a second early exit (bb28dbe).

On the follow-ups: agreed that they are for separate PRs. Payoff normalization in `_initialize_tableaux` was considered here and deferred, as it changes what the tolerances mean; the 1e-10 row of the sweep is a good argument for revisiting it. The warning on non-convergence in the game solvers is worth a discussion of its own.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Fable 5.1)

mmcky:
thanks @oyamad. Merging.

quantecon/optimize/linprog_simplex.py
mmcky: #951

PR #962, opened 2026-09-10
TST: Cover the multi-distribution branch in test_left_eigen_vec

### LINKED ISSUE #961: TST: test_left_eigen_vec's multi-distribution branch is never executed
### Problem

`Test_markovchain_stationary_distributions_KMRMarkovMatrix2.test_left_eigen_vec` branches on `self.n_stat_dists`, but the class's only fixture is the KMR sequential-move matrix with `N = 27`. That chain is irreducible, so it has exactly one stationary distribution: `self.stationary` is always `(1, 28)` and `n_stat_dists` is always `1`. The `else` branch at `quantecon/markov/tests/test_core.py:230-233` therefore never executes.

The same applies to `setup_method`. `stat_shape` always has length 2, so the `if len(stat_shape) == 1` arm at line 201 is dead too, and `n_stat_dists` is only ever assigned from `stat_shape[0]`.

Nothing is broken here — the assertion that does run is the correct one for an irreducible chain. The gap is that the reducible, multiple-stationary-distribution case the `else` branch was written for is not covered anywhere in this class, so it is untested code masquerading as a test.

Noticed while reviewing #798, which rewrites both `np.dot` calls in this method to `@`. That PR is a faithful mechanical rewrite and the dead branch predates it, so this is a separate pre-existing gap rather than anything introduced there.

### Verification

Against `e7b86e5`:

| Fact | Value |
|---|---|
| `P.shape` | `(28, 28)` |
| `mc.is_irreducible` | `True` |
| `stationary_distributions.shape` | `(1, 28)` |
| `len(stat_shape)` | `2` — so the `setup_method` `if` arm at line 201 is never taken |
| `n_stat_dists` | `1` — so the `test_left_eigen_vec` `else` arm at lines 230-233 is never taken |

### Suggested fix

Add a reducible fixture so the branch actually runs. A four-state chain with two absorbing states yields `n_stat_dists == 2`:

```python
P = np.array([[1.0, 0.0, 0.0, 0.0],
              [0.0, 1.0, 0.0, 0.0],
              [0.3, 0.3, 0.4, 0.0],
              [0.0, 0.2, 0.0, 0.8]])
```

Both stationary rows satisfy `v @ P == v`, so the existing assertion passes unchanged once the branch is reachable — the branch is correct, it is simply never entered.

The alternative, if the multi-distribution case is judged to be covered well enough by `test_markovchain_pmatrices`, is to delete the `else` branch and the dead `setup_method` arm instead. Either resolution is fine; leaving an unreachable assertion in place is the outcome worth avoiding.

### Acceptance criteria

- [ ] The `else` branch of `test_left_eigen_vec` is either exercised by a reducible fixture or removed
- [ ] The dead `if len(stat_shape) == 1` arm in `setup_method` is resolved the same way
- [ ] Full suite still passes

---

_Noticed during review of #798 (AI-assisted; claims verified against `e7b86e5` on 2026-09-10)._

techreign:
## What was wrong

`Test_markovchain_stationary_distributions_KMRMarkovMatrix2` in
`quantecon/markov/tests/test_core.py` only ever used the irreducible KMR
sequential-move matrix (`N = 27`), which has exactly one stationary
distribution. Because of that:

- The `else` branch of `test_left_eigen_vec` (the multiple-stationary-
  distributions case) was never executed by the suite.
- The `if len(stat_shape) == 1:` arm in `setup_method` was similarly dead,
  since `MarkovChain.stationary_distributions` is documented as always
  `ndim=2`, so `stat_shape` always has length 2.

Both branches were correct code, just untested code masquerading as tested.

## What changed

- Factored the four existing test methods (`test_markov_matrix`,
  `test_sum_one`, `test_nonnegative`, `test_left_eigen_vec`) out into a
  shared, non-collected base class `_Test_markovchain_stationary_distributions_Base`.
- Kept `Test_markovchain_stationary_distributions_KMRMarkovMatrix2` as a
  subclass with the exact same `N=27` KMR matrix, `setup_method` logic, and
  assertions as before (only the dead `if len(stat_shape) == 1` arm was
  removed, since it can never be reached).
- Added a new subclass, `Test_markovchain_stationary_distributions_ReducibleMarkovMatrix`,
  built on a reducible 4-state matrix with two absorbing states (states 0
  and 1) reached from two transient states (2 and 3). This gives
  `stationary_distributions.shape == (2, 4)`, so `n_stat_dists == 2` and the
  `else` branch of `test_left_eigen_vec` is now actually exercised and
  verified.

No production code in `quantecon/markov/core.py` changed; this is a
test-only fix, matching the "TST:" convention used for the issue.

## How it was verified

```
python -m pytest quantecon/markov/tests/test_core.py -v
# 36 passed
```

To prove the previously-dead `else` branch is now hit, a temporary print
was added inside the `else` branch of `test_left_eigen_vec` and run with
`pytest -k test_left_eigen_vec -s`:

```
Test_markovchain_stationary_distributions_KMRMarkovMatrix2::test_left_eigen_vec PASSED
Test_markovchain_stationary_distributions_ReducibleMarkovMatrix::test_left_eigen_vec TEMP-COVERAGE-PROBE: else branch hit, n_stat_dists= 2
PASSED
```

confirming the KMR case takes the `if` branch (`n_stat_dists == 1`, no
print) and the new reducible-matrix case takes the `else` branch with
`n_stat_dists == 2`. The temporary print was removed before the final
commit (`pytest-cov`/`coverage.py` could not be used for this in the local
dev environment — both hit an unrelated `ImportError: cannot load module
more than once per process` from numpy 2.x's C-extension guard when
combined with coverage's import hooks on this machine).

```
flake8 --select=F401,F405,E231 quantecon/markov/tests/test_core.py
# clean, no output
```

Fixes #961

### DIFF
--- a/quantecon/markov/tests/test_core.py
+++ b/quantecon/markov/tests/test_core.py
@@ -178,30 +178,17 @@ def test_markovchain_pmatrices():
 # Basic Class Structure with Setup #
 ####################################
 
-class Test_markovchain_stationary_distributions_KMRMarkovMatrix2():
+class _Test_markovchain_stationary_distributions_Base():
     """
-    Test Suite for MarkovChain.stationary_distributions using KMR Markov
-    Matrix [suitable for nose]
+    Shared test suite for MarkovChain.stationary_distributions. Concrete
+    subclasses provide `setup_method`, which must set `self.mc`,
+    `self.stationary` (`self.mc.stationary_distributions`) and
+    `self.n_stat_dists` (the number of stationary distributions), so that
+    both the single- and multiple-stationary-distribution cases are
+    exercised by `test_left_eigen_vec`.
     """
 
-    # Starting Values #
-
-    N = 27
-    epsilon = 1e-2
-    p = 1/3
     TOL = 1e-2
-
-    def setup_method(self):
-        """ Setup a KMRMarkovMatrix and Compute Stationary Values """
-        self.P = KMR_Markov_matrix_sequential(self.N, self.p, self.epsilon)
-        self.mc = MarkovChain(self.P)
-        self.stationary = self.mc.stationary_distributions
-        stat_shape = self.stationary.shape
-
-        if len(stat_shape) == 1:
-            self.n_stat_dists = 1
-        else:
-            self.n_stat_dists = stat_shape[0]
 
     def test_markov_matrix(self):
         "Check that each row of matrix sums to 1"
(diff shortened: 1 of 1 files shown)


quantecon-services:
Hi @techreign — a quick process note while this is in the review queue.

QuantEcon's **Code of AI Use** ([QEP-5, QuantEcon/qeps#13](https://github.com/QuantEcon/qeps/pull/13)) doesn't restrict AI-assisted contributions, but it does ask that meaningful AI involvement be disclosed. We'd like it in two places: a trailer in the commit message, and a one-line note in the PR description.

The commit here already carries a `Co-Authored-By: Claude Fable 5.1` trailer, so it's a small change — please add the canonical form alongside it:

```
Assisted-by: Claude Fable 5.1
```

(`git commit --amend`, then force-push the branch.) And add a line to the PR description saying which tools were used and for what — e.g. "Assisted by Claude Fable 5.1 for the test refactor".

The intent is that reviewers know what to expect, and that a human has chosen the task, checked the result, and will own the follow-up discussion.

---

mmcky:
Hi @techreign, thanks for this, and apologies for the slow response. The PR does exactly what #961 suggested, and both the fixture and the refactor are correct. On a closer look, though, the multiple-distribution case is already tested more strictly by `test_markovchain_pmatrices`, which checks `stationary_distributions` against exact values for `np.eye(2)` and a two-class periodic chain, so this adds a path through the test rather than coverage of `MarkovChain`. We're going with the simpler route #961 also offered: since `stationary_distributions` is always 2-D, a single `s @ P ≈ s` assertion checks every row and the branch can go. Closing this one — thanks again!

PR #970, opened 2026-09-21
FIX: game_converters: Number formats in the .gam reader and writer

oyamad:
Four fixes to how numbers are read and written in `game_converters.py`, found while working on the Julia counterpart (QuantEcon/GameTheory.jl#242).

## Reader

- **Numbers with an exponent and no decimal point.** `_str2num` chose between `int` and `float` by whether the token contains a decimal point, so that a token such as `1e3` was passed to `int` and rejected with a `ValueError`. It now tries `int` first and falls back to `float`. As a side effect, `inf` and `nan` become readable.
- **Rational tokens `n/d`.** `_str2num` converts a token of the form `n/d` to the nearest float (via `fractions.Fraction`, with a zero denominator reported as a `ValueError`). This is for the upcoming reader for the Gambit .nfg format, where rational payoffs are part of the format, and which will share this function. For the .gam format, which GameTracer defines for floating-point payoffs, it is left unspecified: it is tested on `_str2num` directly, not through `from_gam_string`.

## Writer

- **Loss of precision.** `to_gam` formatted the payoffs with `np.array2string` at its default precision of 8 digits, so that a payoff such as `1/3` or `0.1 + 0.2` was not recovered by reading the output back. Each float is now formatted with `np.format_float_positional(x, trim='.')`, which gives the shortest representation that round-trips, in the same style as before (`1.`), so that the existing expected strings are unchanged. It never uses the exponent notation; previously the whole array switched to it once a value reached `1e8`. This is also much faster: 0.27 s instead of 18.5 s for a game with 648,000 payoffs.
- **`bool` payoffs.** A game with payoffs of dtype `bool` was written with `True` and `False`, which `from_gam_string` rejects. They are now written as `0` and `1`.

With payoffs of other dtypes formatted with `str`, `Fraction` payoffs are now written as `1/3` instead of `Fraction(1, 3)`; like reading them, this is left unspecified for .gam.

## Tests

`test_from_gam_string_number_formats`, `test_str2num`, `test_gam_writer_float_precision` (`float64` and `float32`; checks that no token has an exponent and that reading back is exact), and `test_gam_writer_bool`.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Fable 5.1)


mmcky:
thanks @oyamad

